Issue #71, opened 2014-12-30
Enhancement: Add Token type


ibuildthecloud:
Add a new type for token.  You follow io.cattle.platform.extension.api.model.ResourceDefinition as an example.

- [ ] Create simple Token POJO
- [ ] Register in an appropriate spring XML using code/implementation/extension-api/src/main/resources/META-INF/cattle/extension-api/spring-extension-api-context.xml as an example (You need to create a TypeSet object)

Sub task of https://github.com/rancherio/cattle/issues/62

deniseschannon:
Closed in favor of rancher/rancher#3829

PR #85, opened 2015-01-05
[Issue #71] Add Token Type

sidharthamani:
@ibuildthecloud

### DIFF
--- a/tests/integration/cattletest/core/test_schema.py
+++ b/tests/integration/cattletest/core/test_schema.py
@@ -17,3 +17,10 @@ def test_account_default_fields(admin_client):
     assert network.nullable
     assert network.create
     assert network.update
+
+
+def token_account_default_fields(token_client):
+    schema = token_client.schema
+    assert schema.types['token'] is not None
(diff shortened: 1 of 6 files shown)


ibuildthecloud:
Can you add one more test to ensure that the token user has access to the token type and no other types.  You can check the schema.types map.  You may find that besides token it has access to types: schema, collection, error, resource.  Those types are fine

sidharthamani:
@ibuildthecloud ok. Doing it now

tests/integration/cattletest/core/test_schema.py
```diff
@@ -17,3 +17,10 @@ def test_account_default_fields(admin_client):
     assert network.nullable
     assert network.create
     assert network.update
+
+
+def token_account_default_fields(token_client):
+    schema = token_client.schema
+    assert schema.types['token'] is not None
```
ibuildthecloud: You don't need to change this but the idiomatic way to say this is `'token' in schema.types` if token was not in schema.types it would raise a KeyError and not return None.

Issue #95, opened 2015-01-08
test_account.test_account_create fails when cattle.account table has >~ 50 records


alena1108:
When test_account runs with the DB having data from the previous test runs, the failure happens:

________________________________________________________________________________ test_account_create ________________________________________________________________________________
Traceback (most recent call last):
  File "/Users/alena/repos/cattle/tests/integration/cattletest/core/test_account.py", line 22, in test_account_create
    assert (cred_count+2) == new_count
AssertionError: assert (100 + 2) == 100

Never observed on a clean db

alena1108:
Pagination issue. Should be fixed by putting "limit=<some big numer>" to the list call to request the full set

PR #99, opened 2015-01-08
[Issue #95] fixed test_account.py: request for a larger data set when qu...

alena1108:
@cjellick

### DIFF
--- a/tests/integration/cattletest/core/test_account.py
+++ b/tests/integration/cattletest/core/test_account.py
@@ -3,7 +3,10 @@ import re
 
 
 def test_account_create(admin_client):
-    cred_count = len(admin_client.list_credential())
+    cred_count = len(admin_client.list_credential(limit=1000))
+
+    assert cred_count < 1000, "Too many credentials; cleanup the database"
+
     count = len(admin_client.list_account())
     account = admin_client.create_account()
 
@@ -18,7 +21,7 @@ def test_account_create(admin_client):
     new_count = len(admin_client.list_account())
     assert (count+1) == new_count
 
-    new_count = len(admin_client.list_credential())
+    new_count = len(admin_client.list_credential(limit=1000))
     assert (cred_count+2) == new_count
 
     creds = account.credentials()


tests/integration/cattletest/core/test_account.py
```diff
@@ -3,7 +3,7 @@
 
 
 def test_account_create(admin_client):
-    cred_count = len(admin_client.list_credential())
+    cred_count = len(admin_client.list_credential(limit=1000))
```
cjellick: The bug is that when this test is ran locally, it will start failing as soon as you hit over 100 total credentials, which happens after just a few runs of the suite of tests. This PR would push back the occurrence of that situation, though it will still _eventually_ occur.
This bug never impacts CI because that is ran with a fresh DB.

I'm fundamentally good with this fix, my only additional thought is do we want to just have an explicit check to effect of:
if cred_count == 1000:
  # Too many credentials. Clear you database
  assert True == False

Or is that overkill? CC @ibuildthecloud

alena1108:
Craig, I see your concern as valid. I'll add the assert.

tests/integration/cattletest/core/test_account.py
ibuildthecloud: I think the assertion is good, it will help some poor developer troubleshoot this.

cjellick:
Code looks good. @alena1108 can you squash this down to a single commit?

cjellick:
I feel so OCD asking, but would you mind shortening the subject of the commit down to 50 characters? 

Just a git commit -ammend
git push -f should do it i think.

cjellick:
Comment updated. Used to say "-ammend" instead of "--amend". Dont do "-ammend" Thanks @vincent99

ibuildthecloud:
@alena1108 sorry for the troubles. Can you rebase? The tests should pass then because the tests on master now are a bit more reliable (not completely yet).

alena1108:
@ibuildthecloud Darren, ran into rebase conflict due to your checkin for tests: 

https://github.com/rancherio/cattle/commit/1a48c16401db74543bc1cd99dc50daec77a3005a

Looks like you removed the lines testing the credentials altogether? If we are not planning to test credentials, then my PR would be obsolete.

ibuildthecloud:
Well shoot, yeah, I removed that code.  That was awfully anticlimactic.  I was all excited the build was working better.

PR #116, opened 2015-01-13
Adding CheckStyle to CI

hibooboo2:
These changes make the build fail when the checkstyle poilcy detects that the rules are not followed. The elipse formatter is to accomany and complement the checkstyle rules.


hibooboo2:
@ibuildthecloud @cjellick This seems to be inline with what we discussed. When the files are formatted on my system there are about 200 errors

ibuildthecloud:
And 200 out of 6300 ain't so bad.

hibooboo2:
Yeah.

PR #143, opened 2015-01-20
Machine integration

cjellick:
Uses a single new type: ```machineHost``` with config properties for each driver.
Currently implements VirtualBox and DigitalOcean


code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/api/machinhost/filter/MachineHostValidationFilter.java
```diff
@@ -0,0 +1,50 @@
...
+    private static final String DRIVER_CONFIG_EXACTLY_ONE_REQUIRED = "DriverConfigExactlyOneRequired";
+
+    @Override
+    public String[] getTypes() {
+        return new String[] { MACHINE_KIND };
+    }
+
+    @Override
+    public Class<?>[] getTypeClasses() {
+        return new Class<?>[] { PhysicalHost.class };
```
ibuildthecloud: If you don't want to match the parent type "physicalHost" then don't return a class.  It's sufficient to return either something from getTypes or getTypeClasses, but both aren't required.

code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/process/machinehost/MachineHostPreCreate.java
```diff
@@ -0,0 +1,65 @@
...
+        Map<Object, Object> newFields = new HashMap<Object, Object>();
+
+        if ( StringUtils.isEmpty(physHost.getExternalId()) ) {
+            String externalId = UUID.randomUUID().toString();
+            newFields.put(PHYSICAL_HOST.EXTERNAL_ID, externalId);
+        }
+
+        // TODO This sets a read-only "driver" field that will be useful to the
+        // UI.
+        // Should we set it here or in a request filter?
```
ibuildthecloud: Up to you, either is fine.  Being that the request filter checks these fields, I'd probably put it there.

code/implementation/docker/machine/src/main/resources/META-INF/cattle/docker-machine-api/spring-docker-machine-api-context.xml
```diff
@@ -0,0 +1,34 @@
...
+    xmlns:util="http://www.springframework.org/schema/util"
+    xsi:schemaLocation="http://www.springframework.org/schema/aop http://www.springframework.org/schema/aop/spring-aop-3.0.xsd
+		http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
+		http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util-3.2.xsd
+		http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-3.0.xsd">
+
+    <bean id="MachineHostAddonTypeSet" class="io.cattle.platform.object.meta.TypeSet" >
+        <property name="typeClasses" >
+            <set>
+                <value>io.cattle.platform.docker.machine.api.machinhost.addon.BaseMachineConfig</value>
```
ibuildthecloud: Typo in package "machinhost"

code/implementation/docker/machine/src/main/resources/schema/base/digitaloceanConfig.json.d/digitalocean.json
```diff
@@ -0,0 +1,20 @@
+{
+    "resourceFields": {
```
ibuildthecloud: Do these types show up at `http://localhost:8080/v1`?

ibuildthecloud:
Is this really all the there is to the cattle side of this?

ibuildthecloud:
If you rebase the flake8 errors will go away.  Apparently flake8 decided to change it's rules.  @sidharthamani already fixed it on master.

code/implementation/docker/machine/src/main/resources/schema/base/digitaloceanConfig.json.d/digitalocean.json
cjellick: No, they don't. Not 100% sure why. Possibly because their parent is tagged with list = false?
https://github.com/rancherio/cattle/pull/143/files#diff-408ea3c153e76fa0049e58a7953d5ba6R6

code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/process/machinehost/MachineHostPreCreate.java
cjellick: If it's truly up to me, I'm going to leave it as-is. My reasoning/justification is that the filter is purely a validation filter right now. It doesn't modify the entity at all. And this class's sole purpose is to modify the entity as part of the create process.

I like that clean separation.

Issue #139, opened 2015-01-16
Setting account.kind=token throws error


ibuildthecloud:

deniseschannon:
Closed in favor of rancher/rancher#3840

PR #144, opened 2015-01-20
[bugfix #139] check that child schema type is actually a child of wrappi...

sidharthamani:
@ibuildthecloud


ibuildthecloud:
And a test....

ibuildthecloud:
Can you make the test a little bit more explicit.
```python
len(admin_client.list_account(kind='type')) > 0
```

PR #150, opened 2015-01-22
add tests for jwt authentication for host_api

pranavs18:
Need to run the tests locally once more. Sometimes the websocket connection fails, sometimes it's successful.

### DIFF
--- a/tests/validation/cattleintegrationtest/core/test_host_api.py
+++ b/tests/validation/cattleintegrationtest/core/test_host_api.py
@@ -1,0 +1,32 @@
+import websocket as ws
+
+from cattle import ApiError
+from common_fixtures import *
+
+
+def test_host_api_token(admin_client):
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
Please put these tests in the validation test project

pranavs18:
The build on Drone is going to fail as the python packages which get installed by default doesn't have the python websocket , websocket-client module. I had to install them separately on my local machine and activate my virtual environment again. @cloudnautique  You would need to probably add these two python packages - websocket, websocket-client.

ibuildthecloud:
Why is this passing now?

pranavs18:
Probably, because I moved it to the validation tests I assume ? Not sure ,if  drone runs the validation tests also ?

ibuildthecloud:
There is a different project for validation twsts

pranavs18:
Oh, just realized that. Moved this file to that project.

PR #152, opened 2015-01-23
Fix the creation logic of child resource type 

sidharthamani:
@ibuildthecloud

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/utils/ApiUtils.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/utils/ApiUtils.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+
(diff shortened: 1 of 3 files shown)


code/framework/api/src/main/java/io/cattle/platform/api/utils/ApiUtils.java
```diff
@@ -186,7 +187,7 @@ public static String getSchemaIdForDisplay(SchemaFactory factory, Object obj) {
         if ( schema.getChildren().size() > 0 ) {
             String kind = ObjectUtils.getKind(obj);
             Schema kindSchema = factory.getSchema(kind);
-            if ( kindSchema != null ) {
+            if ( kindSchema != null && schema.getChildren().contains(kindSchema.getType()) ) {
```
ibuildthecloud: This issue here is that the code should get `getId()` and not `getType()` so

    schema.getChildren().contains(kindSchema.getId())

Issue #161, opened 2015-01-24
Correct the inconsistent request/reply event data structure in AgentBasedProcessHandler


yasker:
When AgentBasedProcessHander is sending a event out, it would contain the top level key, like:

InstanceHostMap: { Instance: { Volumes: { .. } .. }.. }

But currently when agent response with the reply event, the presumed data structure is:

Instance: { Volumes: { .. } .. } 

The reason is this data would be processed as key/value pairs in DefaultProcessInstanceImpl directly, as parameter to "state.applyData()", which would only work without the top level key, since it would apply the data to a known resource.

ibuildthecloud:
As mentioned before this is low priority so please do not dedicate too much time to this.  Your example of modifying the request and replying is actually a bad one.  Because we have no mvcc style checks in place such a pattern can cause data inconsistencies in the DB.  The response should include only the fields that should change, not all of them.  Events can take minutes to be processed so that leave a huge windiw

yasker:
@ibuildthecloud In fact I am already preparing pull request for it. Seems it's not that complex and integration test is passed.

yasker:
@ibuildthecloud I've updated the comment to remove the transfer back the object part.

ibuildthecloud:
Sorry, I accidentally closed this.

PR #162, opened 2015-01-24
Add resource type as top level key of reply event

yasker:
Fixes the issue #161

Currently the request event and reply event's format are inconsistent. The
request event would contained a map as data, using the resource type as the top
level key, but in the reply event's data, the top level is the fields of
resources, rather than resource itself. The inconsistent is very confusing for
agent communication.

This patch would enable the agent to send reply event with consistent format as
request event, say the top level key would be resource type itself, rather than
fields of resource.

### DIFF
--- a/code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/handler/AgentBasedProcessHandler.java
+++ b/code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/handler/AgentBasedProcessHandler.java
@@ -132,10 +132,19 @@ public class AgentBasedProcessHandler extends AbstractObjectProcessHandler imple
 
         postProcessEvent(event, reply, state, process, eventResource, dataResource, agentResource);
 
-        return new HandlerResult(shouldContinue, CollectionUtils.castMap(reply.getData()));
+        return new HandlerResult(shouldContinue, getResourceDataMap(getObjectManager().getType(eventResource), reply.getData()));
     }
 
-    protected void postProcessEvent(EventVO<?> event, Event reply, ProcessState state, ProcessInstance process,
+    protected Map<Object, Object> getResourceDataMap(String type, Object data) {
+    	Object map = CollectionUtils.castMap(data).get(type);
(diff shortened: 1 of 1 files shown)


yasker:
Need python-agent change: https://github.com/rancherio/python-agent/pull/56

yasker:
Integration test passed. Need validation test.

I didn't find cases in simulator using the reply event, so only modify the python-agent to make it compatible.

code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/handler/AgentBasedProcessHandler.java
```diff
@@ -132,10 +132,19 @@ public void progress(Event event) {
 
         postProcessEvent(event, reply, state, process, eventResource, dataResource, agentResource);
 
-        return new HandlerResult(shouldContinue, CollectionUtils.castMap(reply.getData()));
+        return new HandlerResult(shouldContinue, getResourceDataMap(getObjectManager().getType(eventResource), reply.getData()));
     }
 
-    protected void postProcessEvent(EventVO<?> event, Event reply, ProcessState state, ProcessInstance process,
+    protected Map<Object, Object> getResourceDataMap(String type, Object data) {
+    	Object map = CollectionUtils.castMap(data).get(type);
```
ibuildthecloud: Please use asMap.  castMap will result in a runtimeexception if it is not a map.

ibuildthecloud:
Please update the scripts at https://github.com/rancherio/cattle/tree/master/docs/examples/handler-bash too

code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/handler/AgentBasedProcessHandler.java
yasker: @ibuildthecloud I think you meant, "toMap()"? Would update.

yasker:
@ibuildthecloud I would update scripts as well.

PR #167, opened 2015-01-25
Add resource type as top level key of reply event

yasker:
Fixes the issue #161

Currently the request event and reply event's format are inconsistent. The
request event would contained a map as data, using the resource type as the top
level key, but in the reply event's data, the top level is the fields of
resources, rather than resource itself. The inconsistent is very confusing for
agent communication.

This patch would enable the agent to send reply event with consistent format as
request event, say the top level key would be resource type itself, rather than
fields of resource.

The patch updated example bash as well.


yasker:
Test fail due to python-agent change needed: https://github.com/rancherio/python-agent/pull/59

ibuildthecloud:
That's expected.   Once Python agent is released and included in cattle just rebase.

yasker:
@ibuildthecloud I've updated cattle server with temporarily backward compatability.

PR #181, opened 2015-01-28
snapshot: Add schema and snapshot lifecycle handlers

yasker:
This patch added schema and lifecycle handlers for snapshot. The functions
include create snapshot(which chained both create and backup), delete, and test
cases for API and states, as well as input validation for snapshot.

The backend is not included. As well as mechanism to restore snapshot to
container, since there are two possible way of doing it: if we can attach a
volume to container, that would be preferred and standard way, but requires
docker change; or we can restore the snapshot directly when docker is booting
up. Since it would depends on how backend works, we can deal with it later.

Also the storage_pool adding for snapshot backup is missing. Currently the
storage_pool is not expose to user. It need to be expanded to add
secondary storage for backing up snapshot as well.


yasker:
Close pull request for now, and switch to backend work.

PR #190, opened 2015-01-30
Enable passing the reportedUuid from cattle

pranavs18:
PR for enabling JWT-auth by sending the reportedUuid

### DIFF
--- a/code/implementation/host-api/src/main/java/io/cattle/platform/host/service/impl/HostApiServiceImpl.java
+++ b/code/implementation/host-api/src/main/java/io/cattle/platform/host/service/impl/HostApiServiceImpl.java
@@ -70,7 +70,7 @@ public class HostApiServiceImpl implements HostApiService {
         Map<String,Object> data = new HashMap<String,Object>(inputData);
         String uuid = DataAccessor.fields(host).withKey(HostConstants.FIELD_REPORTED_UUID).as(String.class);
         if(uuid != null){
-            data.put(HostConstants.FIELD_REPORTED_UUID,uuid);
+            data.put(HOST_UUID,uuid);
         }
         else{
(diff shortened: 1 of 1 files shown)


code/implementation/host-api/src/main/java/io/cattle/platform/host/service/impl/HostApiServiceImpl.java
```diff
@@ -66,7 +68,13 @@ public HostApiAccess getAccess(Long hostId, Map<String,Object> data) {
...
     protected String getToken(Host host, Map<String,Object> inputData) {
         Map<String,Object> data = new HashMap<String,Object>(inputData);
-        data.put(HOST_UUID, host.getUuid());
+        String uuid = DataAccessor.fields(host).withKey(HostConstants.FIELD_REPORTED_UUID).as(String.class);
+        if(uuid != null){
+            data.put(HostConstants.FIELD_REPORTED_UUID,uuid);
+        }
+        else{
+            data.put(HOST_UUID, host.getUuid());
+        }
```
ibuildthecloud: The key should always be HOST_UUID, the only that should be conditional is from where the uuid is obtained.

pranavs18: Oh, i see. This is the reason, I had to put that check on ReportedUuid and HostUuid in GO code.I'll change this to HOST_UUID and modify the GO code accordingly.

pranavs18:
Updated the PR as per the comments. Drone is failing on test_docker_volume, need to check that.

pranavs18:
Rebased Cattle and drone passes. @ibuildthecloud !

PR #209, opened 2015-02-02
Add cluster data model with some process logic and placeholder cert data model

sonchang:
@ibuildthecloud Somehow I must've accidently closed off the previous PR.  I'm going to see whether this passes CI or not and let you know.


code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/cluster/ClusterAddHostActionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        if (existingMapping != null) {
+            throw new ValidationErrorException(ValidationErrorCodes.INVALID_REFERENCE, ClusterConstants.ADD_REMOVE_HOST_PARAM);
+        }
+
+        ClusterHostMap newMapping = objectManager.newRecord(ClusterHostMap.class);
+        newMapping.setClusterId(cluster.getId());
+        newMapping.setHostId(hostToAdd.getId());
+        newMapping.setState("created");
+
+        newMapping = objectManager.create(newMapping);
```
ibuildthecloud: In general we would like to still orchestrate the creation of mapping entities if possible.  Is there any downside of doing `io.cattle.platform.core.dao.GenericResourceDao.createAndSchedule(Class<T>, Map<String, Object>)`

ibuildthecloud: Call that API you wouldn't pass a state.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/cluster/ClusterRemoveHostActionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public Object perform(String name, Object obj, ApiRequest request) {
+        if (!objectManager.isKind(obj, ClusterConstants.KIND)) {
+            return null;
+        }
+        Host cluster = (Host)obj;
+
+        Long hostToRemoveId = (Long)((Map)request.getRequestObject()).get(ClusterConstants.ADD_REMOVE_HOST_PARAM);
```
ibuildthecloud: A slightly safer and type safe way to do this would be

```java
DataAccessor.fromMap(request.getRequestObject())
   .withKey(ClusterConstants.ADD_REMOVE_HOST_PARAM)
   .as(Long.class);
```

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/cluster/ClusterRemoveHostActionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        Host hostToRemove = objectManager.loadResource(Host.class, hostToRemoveId);
+
+        ClusterHostMap mapping = clusterHostMapDao.getClusterHostMap(cluster, hostToRemove);
+
+        if (mapping == null) {
+            throw new ValidationErrorException(ValidationErrorCodes.INVALID_REFERENCE, ClusterConstants.ADD_REMOVE_HOST_PARAM);
+        }
+        processManager.scheduleProcessInstance(ClusterConstants.CLUSTER_HOST_MAP_REMOVE_PROCESS, mapping, null);
```
ibuildthecloud: You can call `io.cattle.platform.object.process.ObjectProcessManager.scheduleStandardProcess(StandardProcess, Object, Map<String, Object>)` with StandardProcess.REMOVE and avoid having to have your own constant.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/cluster/ClusterRemoveHostActionHandler.java
```diff
@@ -0,0 +1,62 @@
...
+        return cluster;
+    }
+
+    @Inject
```
ibuildthecloud: @cjellick and I had this discussion and were fine with going forward just adding "@Inject" to class member variables, thus removing the need for tons of getters/settings like this.

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostPurge.java
```diff
@@ -20,11 +26,23 @@
...
 
+        // TODO: Implement some bulk operation
+        List<ClusterHostMapRecord> clusterHostMaps = Collections.EMPTY_LIST;
+        if (host.getKind().equals(ClusterConstants.KIND)) {
+            clusterHostMaps = clusterHostMapDao.findClusterHostMapsForCluster(host);
+        } else {
+            clusterHostMaps = clusterHostMapDao.findClusterHostMapsHavingHost(host);
+        }
+        for (ClusterHostMapRecord mapping : clusterHostMaps) {
+            objectProcessManager.scheduleProcessInstance(ClusterConstants.CLUSTER_HOST_MAP_REMOVE_PROCESS, mapping, null);
```
ibuildthecloud: Can use `StandardProcess` API here.

code/iaas/model/src/main/java/io/cattle/platform/core/constants/ClusterConstants.java
```diff
@@ -0,0 +1,10 @@
+package io.cattle.platform.core.constants;
+
+public class ClusterConstants {
+
+    public static final String KIND = "cluster";
+
+    public static final String ADD_REMOVE_HOST_PARAM = "host";
```
ibuildthecloud: To be consistent with the current API conventions, the value should be "hostId".  Changing this of course has a trickling effect on the json files and all the tests.

code/iaas/model/src/main/java/io/cattle/platform/core/constants/ClusterConstants.java
```diff
@@ -0,0 +1,10 @@
...
+
+    public static final String CLUSTER_HOST_MAP_REMOVE_PROCESS = "clusterhostmap.remove";
```
ibuildthecloud: Should be able to remove this constant

resources/content/db/core-022.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.2.xsd http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+    <property name="mediumtext" value="TEXT" dbms="postgresql" />
+    <property name="mediumtext" value="MEDIUMTEXT" />
+    <changeSet author="sonchang (generated)" id="dump1">
+        <createTable tableName="certificate_data">
```
ibuildthecloud: I think we want to just go do "certificate" for the table name.

resources/content/schema/base/cluster.json
```diff
@@ -0,0 +1,35 @@
...
+            "description": "Equivalent to the value used for the --discovery parameter for the cluster server"
+        },
+        "port": {
+            "type": "int",
+            "create": true,
+            "update": true,
+            "nullable": false,
+            "description":"Port the cluster server is running on"
+        },
+        "certificateDataRef": {
```
ibuildthecloud: For consistency this should be "certificateId"

ibuildthecloud: I wonder if we should add a certificate_id column to host, or a host_id column to certificate_id.  Most relationships between object to represent with strong foreign key references.

tests/integration/cattletest/core/common_fixtures.py
```diff
@@ -231,6 +231,24 @@ def wait_transitioning(client, obj, timeout=DEFAULT_TIMEOUT):
     return client.wait_transitioning(obj, timeout=timeout)
 
 
+@pytest.fixture
+def wait_until_expected_state(client, resource, expected_state,
```
ibuildthecloud: In the current tests we have the pattern

```python
obj = client.wait_success(obj)
assert obj.state == expected_state
```

Is there a particular advantage to this function over that pattern?

resources/content/schema/base/clusterHostMap.json
```diff
@@ -0,0 +1,3 @@
+{
+    "collectionMethods" : []
```
ibuildthecloud: If you don't want this object to show up in `/v1` just de-authorize the type, as opposed to limiting the collection methods.  By nothing having things in `/v1` people get the false impression that the object is not accessible, but all it really means is that you can't list them.  One could still fetch the object by ID if they knew it.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/cluster/ClusterAddHostActionHandler.java
sonchang: Thanks!  Didn't know about that way until now.  I also just discovered a bug with the way I was doing it previously too so this is perfect timing.  I'll make this change.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/cluster/ClusterRemoveHostActionHandler.java
sonchang: Wasn't sure whether this would work or not since I'm not using the default process lifecycle for the clusterHostMap resource.  I defined a simplified lifecycle.

resources/content/schema/base/cluster.json
sonchang: I noticed that as well.  Can't do a SQL join to a value in a JSON value

sonchang: Although, not sure whether I'll actually need to do a join.  I think for now, I'm able to get away with just making 2 queries.

sonchang: Also, since the resource name is certificateData, I'll call this 'certificateDataId'

tests/integration/cattletest/core/common_fixtures.py
sonchang: I guess I ran into this when there were multiple state changes.  So the assert would occasionally fail depending on how quickly the resource went through the transitions.

resources/content/db/core-022.xml
sonchang: Ok.  I can make this change.  Are you ok with the column names?

tests/integration/cattletest/core/test_cluster.py
```diff
@@ -0,0 +1,139 @@
+__author__ = 'sonchang'
```
ibuildthecloud: In general we don't add author text to source files.  `git blame` is better and more accurate.

ibuildthecloud:
@sonchang LTGM,  can you just squash and then I'll merge it

sonchang:
squashed

sonchang:
So far I don't see anything wrong and the tests passes for me locally including subsequent runs.

sonchang:
@ibuildthecloud : Passing CI now.  It was previously failing because it was trying to execute a 'remove' action while it was in the 'creating' state instead of the 'created' state.  Added 'creating' as a starting state for 'remove'.

PR #213, opened 2015-02-04
LoadBalancerInstance

alena1108:
@ibuildthecloud PR for LB Instance Deployment logic. 

1) LB agent/instance get created as a part of PostListener ("LoadBalancerAddHostPostListener") for load_balancer_host_map.create process. The mapping creation  is invoked in the ProcessHander for LoadBalancer.addHost API ("LoadBalancerAddRemoveHost")
 
2) LB agent/instance get scheduled for removal as a part of PostListener ("LoadBalancerRemoveHostPostListener") for load_balancer_host_map.remove process. The mapping removal is invoked in:

a) ProcessHandler for LoadBalancer.removeHost API ("LoadBalancerAddRemoveHost") 
b) ProcessHandler for LoadBalancer.remove("LoadBalancerRemove")
c) PreHandler for Host.remove("HostRemovePreAction")

3) LB agent/instance removal consists of:

a) LB instance gets scheduled for Stop with Removed as a chain process (in "LoadBalancerRemoveHostPostListener")
b) LB agent gets scheduled for Deactivate with Remove as a chain action (in "LoadBalancerInstanceRemovePostListener") on LB instance removal process

Removal is more complicated than the creation as it can be triggered by more actions + it has to be scheduled instead of being executed directly. And as the removal consists of multiple steps, scheduling those required putting more logic in.

Creation is called for execution directly, and we don't finish updating LoadBalancer object till LBAgent/Instance creation is completed. In the future, when Set API comes in (and we will be able to set multiple hosts with a single API), I will put implementation for executing for multiple hosts of the same LB, in parallel.


code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/factory/AgentInstanceBuilder.java
```diff
@@ -31,6 +33,10 @@
 
     AgentInstanceBuilder withName(String name);
 
+    AgentInstanceBuilder withUri(String uri);
+
+    AgentInstanceBuilder forParameters(Map<String, Object> params);
```
ibuildthecloud: I think `withParameters()` would be more consistent.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/factory/impl/AgentInstanceFactoryImpl.java
```diff
@@ -263,4 +272,13 @@ public void setResourceMonitor(ResourceMonitor resourceMonitor) {
         this.resourceMonitor = resourceMonitor;
     }
 
+    @SuppressWarnings("unchecked")
+    protected Object getNetworkIds(Agent agent, AgentInstanceBuilderImpl builder) {
```
ibuildthecloud: Can the signature be more strongly typed?  Return `List<Long>`?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/lb/LoadBalancerHostValidationFilter.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class LoadBalancerHostValidationFilter extends AbstractDefaultResourceManagerFilter {
+    
+    GenericMapDao mapDao;
+
+    private static final Map<String, Boolean> actions;
+    static
+    {
+        actions = new HashMap<>();
+        actions.put(LoadBalancerConstants.ACTIONT_ADD_HOST.toLowerCase(), true);
```
ibuildthecloud: You constant has a typo should be ACTION_ADD_HOST, not ACTION**T**_ADD_HOST.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/lb/LoadBalancerHostValidationFilter.java
```diff
@@ -0,0 +1,71 @@
...
+        actions.put(LoadBalancerConstants.ACTIONT_REMOVE_HOST.toLowerCase(), false);
+    }
+    
+    @Override
+    public String[] getTypes() {
+        return new String[] { LoadBalancerConstants.OBJECT_LB };
```
ibuildthecloud: You don't actually have to do both getTypes() and getTypeClasses().  You can just return new String[0] here because the getTypeClasses() below will match the LB type.  The string variation is only really helpful for subtypes.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/lb/LoadBalancerHostValidationFilter.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+        } else {
+            if (mapDao.findToRemove(LoadBalancerHostMap.class, Host.class, hostId, LoadBalancer.class, id) == null) {
+                ValidationErrorCodes.throwValidationError(ValidationErrorCodes.INVALID_OPTION,
+                        LoadBalancerConstants.FIELD_LB_HOST_ID);
+            }
+        }
+    }
+
+    @Inject
```
ibuildthecloud: Side note.  Although the current cattle code base doesn't do this we've decided going forward that we can just add @Inject to class member fields and save all those useful bytes wasted on accessor methods for bean properties.  Just imagine if we refactored the whole code base like that, we'd probably have half the lines :)

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/dao/impl/LoadBalancerInstanceDaoImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Inject
+    public void setObjectManager(ObjectManager objectManager) {
+        this.objectManager = objectManager;
+    }
+
+    @Override
+    public Network getManagedNetwork() {
+        return objectManager.findOne(Network.class, NETWORK.NAME, MANAGED_NETWORK_NAME.get());
```
ibuildthecloud: We can't hard code this.  We need to get more clever on determining what is the network for the load balancer.  Currently we have one network for everyone.  But going forward in a multi-tenant situation we need to create one "managed" network per account.

Let me think about this one.  Currently what I'm thinking is that we add networkId to the Loadbalancer object.  The UI can select the managed one and pass it in (well, the "UI team" won't like that).  But it seems wrong to tie a LB to a network.  So let's think a bit more on this one.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/HostRemovePreAction.java
```diff
@@ -0,0 +1,52 @@
...
+import io.cattle.platform.process.common.handler.AbstractObjectProcessLogic;
+import io.cattle.platform.util.type.Priority;
+
+import java.util.List;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+public class HostRemovePreAction extends AbstractObjectProcessLogic implements ProcessPreListener, Priority {
```
ibuildthecloud: So we don't confuse ourselves, lets go with a name like "HostRemovePreListener"  so that the class name corresponds to the fact it's a ProcessPreListener.

alena1108: Got it, will apply the same change for Post*

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/HostRemovePreAction.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    public int getPriority() {
+        return Priority.DEFAULT;
+    }
+
+    @Inject
```
ibuildthecloud: See look, you could safe upwards of 4 lines of codes by moving the @Inject :)

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerAddHostPostAction.java
```diff
@@ -0,0 +1,47 @@
...
+        return new String[] { LoadBalancerConstants.PROCESS_LB_HOST_MAP_CREATE };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        LoadBalancerHostMap map = (LoadBalancerHostMap) state.getResource();
+        LoadBalancer lb = loadResource(LoadBalancer.class, map.getLoadBalancerId());
+        long hostId = map.getHostId();
+
+        lbInstanceManager.getLoadBalancerInstances(lb, hostId);
```
ibuildthecloud: What is this doing?

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerInstanceRemovePostAction.java
```diff
@@ -0,0 +1,62 @@
...
+import io.cattle.platform.object.process.StandardProcess;
+import io.cattle.platform.process.common.handler.AbstractObjectProcessLogic;
+import io.cattle.platform.util.type.CollectionUtils;
+import io.cattle.platform.util.type.Priority;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+public class LoadBalancerInstanceRemovePostAction extends AbstractObjectProcessLogic implements ProcessPostListener,
```
ibuildthecloud: Same comment on the name, something like "*PostListener"

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/lb/LoadBalancerHostValidationFilter.java
alena1108: I am going to change it for the files modified/created by this commit. Will create a separate PR for code cleanup for other files.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerInstanceRemovePostAction.java
```diff
@@ -0,0 +1,62 @@
...
+        Agent lbAgent = objectManager.loadResource(Agent.class, instance.getAgentId());
+
+        // try to remove first
+        try {
+            getObjectProcessManager().scheduleProcessInstance(AgentConstants.PROCESS_REMOVE, lbAgent,
+                    null);
+        } catch (ProcessCancelException e) {
+            // ignore
+        }
+        objectProcessManager.scheduleStandardProcess(StandardProcess.DEACTIVATE,
```
ibuildthecloud: I think you want to move this to the catch block.  So either you remove, but if an exception is thrown you deactivate and remove.

Random note.  Seems annoying that you have to do getObjectProcessManager() everywhere.  I think we should just make objectProcessManager protected.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerRemoveHostPostAction.java
```diff
@@ -0,0 +1,66 @@
...
+import io.cattle.platform.util.type.CollectionUtils;
+import io.cattle.platform.util.type.Priority;
+
+import java.util.List;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+public class LoadBalancerRemoveHostPostAction extends AbstractObjectProcessLogic implements ProcessPostListener, Priority {
```
ibuildthecloud: more name comments.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerAddHostPostAction.java
alena1108: Makes a call to lbInstanceManager to create LB instance (if it doesn't exist already). Method name is misleading I guess, I will rename it to "createLoadBalancerInstance"

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/HostRemovePreAction.java
alena1108: Changing it already! :)

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerInstanceRemovePostAction.java
alena1108: Its already exposed. I will change it from getObjectProcessManager() to objectProcessManager everywhere

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/dao/impl/LoadBalancerInstanceDaoImpl.java
alena1108: @ibuildthecloud Darren, if you decide to go with putting a networkId on the LB, do you think we should allow only Targets (containers) from the same network to be added to this LB?

code/iaas/logic/src/main/java/io/cattle/platform/process/lb/LoadBalancerAddRemoveHost.java
```diff
@@ -19,22 +18,22 @@
...
     public String[] getProcessNames() {
         return new String[] { LoadBalancerConstants.PROCESS_LB_ADD_HOST, LoadBalancerConstants.PROCESS_LB_REMOVE_HOST };
-
     }
 
     @Override
     public HandlerResult handle(ProcessState state, ProcessInstance process) {
         LoadBalancer lb = (LoadBalancer) state.getResource();
-        Long hostId = DataAccessor.fromMap(state.getData()).
+        long hostId = DataAccessor.fromMap(state.getData()).
```
ibuildthecloud: Is it optimistic to assume a NPE would never be possible?

alena1108: hostId is set with required=true in the API (addRemoveLoadBalancerHostInput.json)

"resourceFields" : {
        "hostId" : {
            "type": "reference[host]",
            "nullable": false,
            "required": true
        }

Unless there is any other entry point for the Process rather than API, it should be fine I think.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/dao/impl/LoadBalancerInstanceDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        List<Long> hostIds = new ArrayList<Long>();
+        for (LoadBalancerHostMap map : mapDao.findToRemove(LoadBalancerHostMap.class, LoadBalancer.class, lbId)) {
+            hostIds.add(map.getHostId());
+        }
+        return hostIds;
+    }
+
+    @Override
+    public Network getLoadBalancerInstanceNetwork(LoadBalancer loadBalancer) {
+        // today the only network in cattle is owned by the system account, so passing its id
```
ibuildthecloud: Can you just check for the account owned one and if none is found go to the system account?

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/dao/impl/LoadBalancerInstanceDaoImpl.java
```diff
@@ -0,0 +1,55 @@
...
+        // in the future, will pass the lb owner id
+        List<? extends Network> networks = ntwkDao.getNetworksForAccount(accountDao.getSystemAccount().getId(),
+                NetworkConstants.KIND_HOSTONLY);
+        if (networks.isEmpty()) {
+            throw new RuntimeException(
+                    "Load Balancer account doesn't own a network where Load Balancer Instance can be started");
+        }
+        return objectManager.findOne(Network.class, NETWORK.ID, networks.get(0).getId());
```
ibuildthecloud: You can just return networks.get(0).  There's no difference in an object loaded from a jOOQ query or from the ObjectManager.

Also, if you do want to load an object from the ObjectManager, you can use ObjectManager.loadResource(Network.class, 42L)

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/dao/impl/LoadBalancerInstanceDaoImpl.java
alena1108: Sure, makes sense.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/dao/impl/LoadBalancerInstanceDaoImpl.java
alena1108: One of my random blonde moments; will fix.

Yea, I do use loadResource pretty often in the process logic code.

PR #235, opened 2015-02-05
Add warning message if there are different handlers with the same name

sonchang:
@ibuildthecloud 
Add warning message.  I encountered this scenario when I added a new AgentBasedProcessHandler to also handle the host.activate event.  Unfortunately, I also gave it the same name and noticed it wasn't getting executed.  We preserve the list of handlers elsewhere but it disappears when we try to prioritize it in the TreeSet.  Can easily resolve the problem by using different handler names.  But adding this warning in case others accidentally encounter it.

### DIFF
--- a/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
+++ b/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
@@ -30,8 +30,6 @@ public class ExtensionManagerImpl implements ExtensionManager, InitializationTas
 
     private static final String WILDCARD = "regexp:";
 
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
Is there any reason we wouldn't want to just throw a runtime exception and fail the start of the server?

sonchang:
I'm good with that too

ibuildthecloud:
Let's go with that.  I like that fail fast approach.  Rarely do people ever seem to notice the really helpful warning messages.

sonchang:
done

ibuildthecloud:
darn, tests failed with a transient failure.  I'll put in a PR that will hopefully fix that.

PR #244, opened 2015-02-08
Closes #221: return userType on /v1/token

### LINKED ISSUE #221: Return user type on POST /v1/token

sidharthamani:
@ibuildthecloud


ibuildthecloud:
Why don't you just return account.getKind() and then you don't need to hardcode "admin" checks

ibuildthecloud:
I'm guessing userType should be "r" not "cr."   Sorry I didn't notice that before.

PR #252, opened 2015-02-09
LoadBalancer: added more filters

alena1108:
@ibuildthecloud Darren, this PR adds more filters to LB API actions. 

* filter on add/remove Target action
* filter on add/remove Listener action

Let me know if you think the helper method I've added - "validateGenericMapAction" - should be moved from AbstractDefaultResourceManagerFilter to some *Utils class, and I'll fix it.


ibuildthecloud:
Yes a util method would be better.  You will have to accept the DAO as a parameter though to the util since you don't want/can't inject it.  Or if anything, move it to AbstractResourceManagerFilter

alena1108:
Updated the PR with the fix moving the helper method to the *Util class.

PR #259, opened 2015-02-12
LB haproxy configuration

alena1108:
@ibuildthecloud Darren, its WIP PR; have a couple of questions:

1) I've realized we don't expose "algorithm" for the local load balancer. Should we just default it to "roundrobin" in haproxy.cfg, or let user specify it.

2) Logging. HA proxy sends its messages to rsyslog, and its not running by default on agent-instance. Do you see any potential downsides if I enable it?

3) There are couple of LB config files that I need to modify before HA proxy starts for the very first time. What would be the right place to invoke the modification: a) DOCKERFILE of agent-instance b) startup.sh script

### DIFF
--- a/code/iaas/config-item/server/pom.xml
+++ b/code/iaas/config-item/server/pom.xml
@@ -49,5 +49,10 @@
             <groupId>org.freemarker</groupId>
             <artifactId>freemarker</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-iaas-lb-instance</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 </project>
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/LoadBalancerTargetInfo.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/LoadBalancerTargetInfo.java
@@ -1,0 +1,29 @@
+package io.cattle.platform.configitem.context.data;
+
+public class LoadBalancerTargetInfo {
+    private String name;
+    private String ipAddress;
+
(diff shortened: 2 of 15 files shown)


ibuildthecloud:
>1) I've realized we don't expose "algorithm" for the local load balancer. Should we just default it to "roundrobin" in haproxy.cfg, or let user specify it.

I would expose it as a parameter on the LB or listener.

> 2) Logging. HA proxy sends its messages to rsyslog, and its not running by default on agent-instance. Do you see any potential downsides if I enable it?

I would prefer to not have rsyslog running.  Can haproxy not just log to a file?

> 3) There are couple of LB config files that I need to modify before HA proxy starts for the very first time. What would be the right place to invoke the modification: a) DOCKERFILE of agent-instance b) startup.sh script

I'd prefer that the only change to the Dockerfile would be the addition of `apt-get install haproxy`.  All other configuration changes should just be done on demand in the `apply.sh`.  Do you have examples of other config files that need to change?

alena1108:
> I would expose it as a parameter on the LB or listener.

Would put it on a listener as a separate PR

> I would prefer to not have rsyslog running. Can haproxy not just log to a file?

Unfortunately, not. For performance & maintenance reasons HAProxy doesn't log directly to files; it always logs against a syslog server. 

> Do you have examples of other config files that need to change?

By default, harpoxy is disabled. To enable it, I have to run this command "sed -i -e 's/ENABLED=0/ENABLED=1/g' /etc/default/haproxy". Also if we decide to enable syslog, I will have to modify rsyslog.conf to enable listening on UDP port. Sure, I can put it to apply.sh

alena1108:
Based on offline discussion - this PR will go in w/o haproxy logs enabled.

alena1108:
@ibuildthecloud Darren, removed "WIP" from the subject - the PR is ready for a review.

PR #263, opened 2015-02-13
LB: expose algorithm for LBListener

alena1108:
@ibuildthecloud Darren, this PR exposes algorithm on LB listener. Following algorithms are exposed:

roundrobin
leastconn
source
uri
first

There are couple of extras that haproxy supports (static-roundrobin, url-param,rdp-cookie) but I think for v1 it would make sense to expose a limited set of most popular ones. Let me know what you think

### DIFF
--- a/resources/content/schema/base/loadBalancerListener.json
+++ b/resources/content/schema/base/loadBalancerListener.json
@@ -21,7 +21,12 @@
             "type": "int",
             "min": 1,
             "max": 65535
+        },
+        "algorithm": {
+            "type": "enum",
+            "options": [ "roundrobin", "leastconn", "source"],
+            "nullable": true,
+            "default": "roundrobin"
         }
-
     }
 }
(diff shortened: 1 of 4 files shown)


ibuildthecloud:
Can you explain what uri and first are?  I'm guess first is to pick the first available host.  I'm conformable with roundrobin, leastconn, source as that is what we've always done in IaaS.  I don't know what uri is and then first might be problematic because it seems like we should then allow the user to deterministically order the targets.

I think for the first pass lets just add the first three.

ibuildthecloud:
I'm on the fence here.  Do you really think this should be a DB field?  I think I'm leaning towards just adding it to the data field.

alena1108:
Darren, sorry for the delayed reply. Its fine if we expose just 3 algorithms in V1 (in CloudStack we also did just leastconn, roundrobin and source). I will submit an another PR for this. Let me know if you want me to move it to the data field.

ibuildthecloud:
i feel better with having it as a data field right now.

resources/content/schema/base/loadBalancerListener.json
```diff
@@ -21,7 +21,12 @@
             "type": "int",
             "min": 1,
             "max": 65535
+        },
+        "algorithm": {
+            "type": "enum",
+            "options": [ "roundrobin", "leastconn", "source"],
```
ibuildthecloud: I think we could default to round robin

alena1108: Ok, will fix.

PR #267, opened 2015-02-20
LB config item update and public port binding

alena1108:
@ibuildthecloud WIP PR; will update you once its fully ready and tested.


alena1108:
@ibuildthecloud Darren, PR is ready to be merged. I've tested LB from end to end by deploying 2 nginx containers and creating an LB for them. HA proxy was configured correctly, ports were open on the host for the public listener's ports, and traffic was balanced between the servers according to RoundRobin algorithm.
If further testing reveals some bugs in the process, I'll be fixing them along the way, and create a separate PR for each of them.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/service/impl/AgentInstanceManagerImpl.java
```diff
@@ -64,9 +72,8 @@
                         .forVnetId(nic.getVnetId())
                         .withSystemContainerType(SystemContainer.NetworkAgent)
                         .build();
-            } else {
-                start(agentInstance);
             }
+            start(agentInstance);
```
ibuildthecloud: Can you explain why this change was necessary?

code/iaas/agent-instance/src/main/resources/META-INF/cattle/system-services/agent-instance-defaults.properties
```diff
@@ -9,7 +9,7 @@ agent.instance.start.items.increment.extra=
 agent.instance.services.base.items=${agent.instance.start.items.apply},${agent.instance.start.items.increment},${agent.instance.services.base.items.extra}
 agent.instance.services.base.items.extra=
 
-agent.instance.services.processes=nic.activate,instancelink.update,instancelink.activate,hostipaddressmap.activate,${agent.instance.services.processes.extra}
+agent.instance.services.processes=nic.activate,instancelink.update,instancelink.activate,hostipaddressmap.activate${agent.instance.services.processes.extra}
```
ibuildthecloud: Was there a problem with have the value end with a ","?  This changes means that if somebody used the extra field they would need to start with a ",".  In the end if the extra field is causing problem we can just removed the property.  It was a useless hook added for extension that is no longer needed.

alena1108: Sorry, Darren, removed the "," accidentally when I removed "port.activate"/"port.deactivate" (added them previously, and then realized that it shouldn't be there). I'll roll back the changes for this file.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/service/impl/AgentInstanceManagerImpl.java
alena1108: One more leftover from the thing I've tried to implement, and that didn't work with idempotent checks on so I rolled changes back. I've tried to implement creating system Instance w/o starting it in AgentInstanceFactory, and delegating the start to AgentInstanceManager/LoadBalancerManager instead. Will rollback this as well.

PR #270, opened 2015-02-24
LB healthcheck and policy configuration

alena1108:


alena1108:
@ibuildthecloud Darren, the PR is ready. Let me know if you are OK with the APIs

PR #274, opened 2015-02-26
Added "systemContainer" field for instance object

alena1108:
@ibuildthecloud Darren, this PR introduces "Managed" field (values: Unmanaged, NetworkAgent, LoadBalancerAgent) so the UI/API can do sort based on that.

### DIFF
--- a/resources/content/db/core-024.xml
+++ b/resources/content/db/core-024.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+    <property name="mediumtext" value="TEXT" dbms="postgresql" />
+    <property name="mediumtext" value="MEDIUMTEXT" />
+    <changeSet author="alena (generated)" id="dump1">
+        <addColumn tableName="instance">
+            <column name="system_container" type="VARCHAR(128)"/>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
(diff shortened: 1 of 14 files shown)


vincent99:
Can the field be an enum in the schema?

resources/content/db/core-024.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+    <property name="mediumtext" value="TEXT" dbms="postgresql" />
+    <property name="mediumtext" value="MEDIUMTEXT" />
+    <changeSet author="alena (generated)" id="dump1">
+        <addColumn tableName="instance">
+            <column defaultValue="Unmanaged" name="managed" type="VARCHAR(128)"/>
```
ibuildthecloud: Default to null

alena1108:
Sure, Vince.

cjellick:
Is `managed` the right name for the field? It seems odd that a container would have `"managed": "unmanged"`. The name `managed` to me would indicate that the container is managed by rancher, but what we really want to indicate is whether the container is a rancher management container, right? 

Also, the value `unmanaged` confuses me because I assume it means that the container is somehow not managed by rancher. I could also see it being confusing with our network options of "managed" and "unmanaged" because I'd potentially have a container that is marked as unmanaged but has managed networking.

Straw man suggestion that I don't love:
* field name: `function`
* values: `userContainer | NetworkAgent | LoadBalancerAgent`

I'm potentially being pendantic, so if no one else agrees with this, I won't fight it.

alena1108:
Yes, by "managed" I've meant that the container is used to manage certain rancher service(s).

How about "systemContainer" as the name of the field? The value of this field will be NULL for user containers.

sonchang:
I agree with Craig the name is confusing.  I do like the name
"systemContainer" although this might get a little confusing with rancherOS
as well.

On Thu, Feb 26, 2015 at 1:58 PM, Alena Prokharchyk <notifications@github.com

cjellick:
I like systemContainer with a null value for user containers.

vincent99:
'systemContainer' sounds fine.  :-1:  for 'function' (and anything else that's a language keyword).

ibuildthecloud:
@alena1108 Can you put in a PR for ipsec_tunnel.py on python-agent that checks for agent_id != null && ( systemContainer == null || systemContainer == NetworkAgent ).  Then i'll merge both at about the same time.

alena1108:
@ibuildthecloud https://github.com/rancherio/python-agent/pull/80

PR #280, opened 2015-03-03
loadBalancer.restart API

alena1108:
@ibuildthecloud @vincent99 one more API for a load balancer - in case something really goes south, user can call it and the haproxy instance(s) will get re-deployed + haproxy config will get reloaded on them.


ibuildthecloud:
I think we should hold off and do a graceful restart.  Lets get this into the next phase.

alena1108:
Temporarily closing based on the last comment.

PR #286, opened 2015-03-06
ServiceDiscovery initial checkin

alena1108:
@ibuildthecloud Darren, very WIP PR. Once you review the flow, I am yet to:

* add more config items to launchConfig.json
* add more config items to service.json
* add more validation to CRUD processes logic


ibuildthecloud:
Can your rebase and rename the core-028.xml so that I run this locally

alena1108:
@ibuildthecloud Darren, all the checks you've suggested on Friday - embedded launchConfig/UI progress checkpoints/improved style for exported docker-compose.yml - are in

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -144,4 +144,25 @@
...
+    <!-- Service Discovery Environment -->
+    <process:process name="environment.create" resourceType="environment" start="requested" transitioning="activating" done="active" />
+    <process:process name="environment.activateservices" resourceType="environment" start="active" transitioning="updating-active" done="active" />
+    <process:process name="environment.deactivateservices" resourceType="environment" start="active" transitioning="updating-active" done="active" />
+    <process:process name="environment.remove" resourceType="environment" start="active, activating, updating-active" transitioning="removing" done="removed" />
+    
+    <!-- Service Discovery Service -->
+    <process:process name="service.create" resourceType="service" start="requested" transitioning="registering" done="inactive" />
+    <process:process name="service.activate" resourceType="service" start="inactive" transitioning="activating" done="active" />
+    <process:process name="service.deactivate" resourceType="service" start="active,activating,updating-active" transitioning="deactivating" done="inactive" />
```
sonchang: I don't see an "updating-active" state for this process.  Same with the "service.remove" start state below.

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -144,4 +144,25 @@
     <!-- Cluster host map -->
     <process:process name="clusterhostmap.create" resourceType="clusterHostMap" start="requested" transitioning="creating" done="created" />
     <process:process name="clusterhostmap.remove" resourceType="clusterHostMap" start="created,creating" transitioning="removing" done="removed" />
+
+    <!-- Service Discovery Environment -->
+    <process:process name="environment.create" resourceType="environment" start="requested" transitioning="activating" done="active" />
+    <process:process name="environment.activateservices" resourceType="environment" start="active" transitioning="updating-active" done="active" />
+    <process:process name="environment.deactivateservices" resourceType="environment" start="active" transitioning="updating-active" done="active" />
```
sonchang: done state for this is confusing since it's still "active".  I had a similar question above too with the transition from "active" to "active".  I'm wondering whether activateservices and deactivateservices should be regular actions rather than actual transition states.  Sort of like the resourceActions "addhost" and "removehost" in cluster.json and ClusterAddHostActionHandler.java

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
alena1108: @sonchang Good catch; i used to have the process setting the state to updating-active, then I remove it, but forgot to remove it from the start state for the object.

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
alena1108: @sonchang The reason why I've put the state, is to prevent activate/deactivate services to be called in parallel. Keeping the state should be harmless I guess.

sonchang: Hmm.  I'm thinking a lock would be a better way to prevent activate/deactivate to be called in parallel.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceAddRemoveLinkServiceValidationFilter.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class ServiceAddRemoveLinkServiceValidationFilter extends AbstractDefaultResourceManagerFilter {
+
+    @Inject
+    ServiceConsumeMapDao consumeMapDao;
+
+    private static final Map<String, Boolean> ACTIONS;
+
+    static {
+        ACTIONS = new HashMap<>();
```
sonchang: Quick question: Is the boolean in this map just to drive the logic in validateAction()?
If so, I'd suggest maybe using an enum?

Sort of like the following:
public class foo {

public enum Actions {
 foo, bar
}

public static void main(String[] args) {
  String input = "foo";
  Actions a = Actions.valueOf(input);
  switch(a) {
        case foo:
          System.out.println("execute foo");
          break;
        case bar:
          System.out.println("execute bar");
          break;
  }
}
}

There are also similar examples you can look up online if we have spaces in the action name.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceConsumeMapDaoImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.inject.Inject;
+
+public class ServiceConsumeMapDaoImpl extends AbstractJooqDao implements ServiceConsumeMapDao {
+
+    @Inject
+    ObjectManager objectManager;
+
+    @Override
+    public ServiceConsumeMap findMapToRemove(long serviceId, long consumedServiceId) {
+        ServiceConsumeMap map = objectManager.findOne(ServiceConsumeMap.class,
```
sonchang: You'll have to add the removed filter to both of these calls: findMapToRemove() and findNonRemovedMap()

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/api/action/AddServiceLinkActionHandler.java
```diff
@@ -0,0 +1,58 @@
...
+
+    protected void createMap(long serviceId, long consumedServiceId) {
+        ServiceConsumeMap map = consumeMapDao.findNonRemovedMap(serviceId, consumedServiceId);
+
+        if (map == null) {
+            map = objectManager.create(ServiceConsumeMap.class,
+                    SERVICE_CONSUME_MAP.SERVICE_ID,
+                    serviceId, SERVICE_CONSUME_MAP.CONSUMED_SERVICE_ID, consumedServiceId);
+        }
+        objectProcessManager.executeProcess(ServiceDiscoveryConstants.PROCESS_SERVICE_CONSUME_MAP_CREATE,
```
sonchang: Super minor, but I'm wondering whether this should be scheduled as well rather than executed immediately.  I forget whether I had to change mine back then or not.

alena1108: @sonchang Its just a DB operation, so I'd rather wait for it to finish.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceConsumeMapDaoImpl.java
alena1108: @sonchang yup, will do, thank you for the catch.

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
alena1108: @sonchang I'm changing the logic - activate/deactivate services will become actionHandlers and can be executed on the environment being in any state.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/api/action/AddServiceLinkActionHandler.java
sonchang: no problem.  yeah.  i doubt it matters much. :-)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServerImpl.java
```diff
@@ -0,0 +1,296 @@
...
+
+    @Override
+    public int[] getWeights(int size, int total) {
+        int[] weights = new int[size];
+        Integer sum = 0;
+        int i = 0;
+        while (true) {
+            int value = total / size;
+            if (sum + value > total) {
+                weights[i - 1] = total - sum + value;
```
sonchang: I think this should be:
weights[i-1] = total - sum;

For example, if total = 100, sum = 70, and value = 35,
I'm guessing we want that last value to be 30?

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceAddRemoveLinkServiceValidationFilter.java
alena1108: @sonchang as the datastructure is used only within the same class, and the field you refer to can have just 2 values, I think enum would be an overkill

sonchang: I agree that it is a bit of an overkill in terms of the amount of code.  But I do feel it'd be a lot more clear.  The original use of the map felt non-intuitive.  I think I'd even prefer just having a Set with 2 constants and doing the comparisons explicitly.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/EnvironmentActivateServices.java
```diff
@@ -0,0 +1,83 @@
...
+                try {
+                    objectProcessManager.scheduleProcessInstance(ServiceDiscoveryConstants.PROCESS_SERVICE_ACTIVATE,
+                            service, data);
+                } catch (ProcessCancelException e) {
+                    // do nothing
+                }
+            }
+        }
+        for (Service service : services) {
+            progress.checkPoint("activate service " + service.getName());
```
sonchang: I'm guessing we probably want the service.getRemoved() check here as well

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServerImpl.java
alena1108: @sonchang not that at this point weights[i-1] already has the value set. If you just reset it to be "total-sum", you will lose this previously set value, and the final sum will be less than the total.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/resource/ServiceDiscoveryConfigItem.java
```diff
@@ -0,0 +1,131 @@
...
+    public static ServiceDiscoveryConfigItem getServiceConfigItemByRancherName(String internalName) {
+        for (ServiceDiscoveryConfigItem serviceItem : supportedServiceConfigItems) {
+            if (serviceItem.getRancherName() != null && serviceItem.getRancherName().equalsIgnoreCase(internalName)) {
+                return serviceItem;
+            }
+        }
+        return null;
+    }
+
+    public boolean isDockerComposeProperty() {
```
sonchang: Super minor.  Can we move this above the static functions so it's grouped with the rest of the properties?

alena1108: @sonchang yup

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServerImpl.java
sonchang: You're right

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -0,0 +1,246 @@
...
+        createServiceInstances(service, scale);
+        return null;
+    }
+
+    private void activateConsumedServices(List<? extends Integer> consumedServiceIds,
+            Map<String, Object> data) {
+        List<Service> consumedServices = new ArrayList<>();
+        if (consumedServiceIds != null) {
+            for (Integer consumedServiceId : consumedServiceIds) {
+                Service consumedService = objectManager.findOne(Service.class, SERVICE.ID, consumedServiceId,
```
sonchang: Minor: Can we add a comment line to note that we should probably bulk load the services in the future.  Definitely don't need to do it now but I think it's worth the optimization in the future.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    private Instance createInstance(Service service, int i) {
+        Environment env = objectManager.findOne(Environment.class, ENVIRONMENT.ID, service.getEnvironmentId());
+        String instanceName = String.format("%s_%s_%d", env.getName(), service.getName(), i + 1);
+        Instance instance = objectManager.findOne(Instance.class, INSTANCE.NAME, instanceName,
+                INSTANCE.REMOVED, null);
+        
+        if (instance == null || instance.getRemoved() != null) {
```
sonchang: Don't need the instance.getRemoved() check since it should be filtered out already

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/ServiceDiscoveryServer.java
```diff
@@ -0,0 +1,17 @@
+package io.cattle.platform.servicediscovery.service;
+
+import io.cattle.platform.core.model.Service;
+
+import java.util.AbstractMap.SimpleEntry;
+import java.util.List;
+import java.util.Map;
+
+public interface ServiceDiscoveryServer {
```
sonchang: This is more of a nitpick, but can we rename this to ServiceDiscoveryService.
When I see Server, I think of a daemon process.

alena1108: @sonchang I wanted to avoid the Service-Service tautology :-)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/resource/ServiceDiscoveryConfigItem.java
```diff
@@ -0,0 +1,131 @@
...
+    public static final ServiceDiscoveryConfigItem MEMLIMIT = new ServiceDiscoveryConfigItem("memory", "mem_limit");
+    public static final ServiceDiscoveryConfigItem PRIVILEGED = new ServiceDiscoveryConfigItem("privileged", "privileged");
+    public static final ServiceDiscoveryConfigItem RESTART = new ServiceDiscoveryConfigItem("restartPolicy", "restart");
+    public static final ServiceDiscoveryConfigItem STDINOPEN = new ServiceDiscoveryConfigItem("stdinOpen", "stdin_open");
+    public static final ServiceDiscoveryConfigItem TTY = new ServiceDiscoveryConfigItem("tty", "tty");
+    public static final ServiceDiscoveryConfigItem CPUSHARES = new ServiceDiscoveryConfigItem("cpuShares", "cpu_shares");
+    public static final ServiceDiscoveryConfigItem EXPOSE = new ServiceDiscoveryConfigItem("expose", "expose");
+    public static final ServiceDiscoveryConfigItem EXTERNALLINKS = new ServiceDiscoveryConfigItem(
+            "instanceLinks",
+            "external_links", true, true);
```
sonchang: We can use the other constructor since it's true, true

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServerImpl.java
```diff
@@ -0,0 +1,296 @@
...
+        List<Service> translateToInstances = new ArrayList<>();
+        List<? extends Integer> consumedServiceIds = (List<? extends Integer>) getServiceDataAsMap(service).get(
+                ServiceDiscoveryConfigItem.VOLUMESFROMSERVICE.getRancherName());
+
+        if (consumedServiceIds != null) {
+            for (Integer consumedServiceId : consumedServiceIds) {
+                Service consumedService = objectManager.findOne(Service.class, SERVICE.ID,
+                        consumedServiceId);
+                boolean found = false;
+                for (Service serviceToExport : servicesToExport) {
```
sonchang: Could be worth it to create a lookup map to use prior to iterating through consumedServiceIds.  Not sure how large these lists can get.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/ServiceDiscoveryServer.java
sonchang: Yeah.  I noticed that too, but I think it's still better.  Not a merge blocker to me though. :-)

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -144,4 +144,23 @@
     <!-- Cluster host map -->
     <process:process name="clusterhostmap.create" resourceType="clusterHostMap" start="requested" transitioning="creating" done="created" />
     <process:process name="clusterhostmap.remove" resourceType="clusterHostMap" start="created,creating" transitioning="removing" done="removed" />
+
+    <!-- Service Discovery Environment -->
+    <process:process name="environment.create" resourceType="environment" start="requested" transitioning="activating" done="active" />
+    <process:process name="environment.remove" resourceType="environment" start="active, activating, updating-active" transitioning="removing" done="removed" />
```
sonchang: Is there an 'updating-active' state?

alena1108: @sonchang nope, again, a leftover from the state transition I removed. Will fix.

sonchang:
@alena1108 LGTM

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/NetworkDaoImpl.java
```diff
@@ -72,4 +85,37 @@ public Nic getPrimaryNic(long instanceId) {
                         .and(NETWORK.REMOVED.isNull()))
                 .fetchInto(NetworkRecord.class);
     }
+
+    @Override
+    public Network getNetworkForObject(Object object) {
```
cjellick: After this is merged, I will use this method for native docker. I may need to add support for getting an unmanaged network (as opposed to assuming `KIND_HOSTONLY`).

Just bringing this up mainly for informational purposes.

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -144,4 +144,23 @@
...
+    <process:process name="environment.remove" resourceType="environment" start="active, activating" transitioning="removing" done="removed" />
+    
+    <!-- Service Discovery Service -->
+    <process:process name="service.create" resourceType="service" start="requested" transitioning="registering" done="inactive" />
+    <process:process name="service.activate" resourceType="service" start="inactive" transitioning="activating" done="active" />
+    <process:process name="service.deactivate" resourceType="service" start="active,activating" transitioning="deactivating" done="inactive" />
+    <process:process name="service.remove" resourceType="service" start="inactive, registering, active,activating" transitioning="removing" done="removed" />
+    
+    <!-- Service Discovery Service/Instance map -->
+    <process:process name="serviceexposemap.create" resourceType="serviceExposeMap" start="requested" transitioning="activating" done="active" />
```
cjellick: Do the resources need the word map in them? Does `serviceExpose` and `serviceConsume` sound any better?

Just a suggestion, you can take it or leave it.

cjellick:
You're introducing a resource called service we also have an account "kind" called service. It denotes an account that has special permissions to rancher. For example, go-machine-service interacts with the rancher api "service".

Here are two references to where the service account kind is setup:
https://github.com/rancherio/cattle/blob/47bac79faba126822048d9b4484749e7eb82b3b0/code/iaas/accounts/src/main/resources/META-INF/cattle/accounts/spring-accounts-context.xml#L62-L88

https://github.com/rancherio/cattle/blob/47bac79faba126822048d9b4484749e7eb82b3b0/resources/content/schema/service/service-auth.json

Should we be concerned about overloading the term? I think its more important for your resource to be called service, so we could change the account kind. I'm not fully convinced that anything needs to change, just wanted to bring it up for consideration. And if we did change the account kind, I imagine it could happen independent from this PR.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceAddRemoveLinkServiceValidationFilter.java
```diff
@@ -0,0 +1,64 @@
...
+    private void validateAction(long id, Map<String, Object> data, String action) {
+        Long consumedServiceId = (Long) data.get(ServiceDiscoveryConstants.FIELD_SERVICE_ID);
+        if (ACTIONS.get(action) == Operation.ADD) {
+            if (consumeMapDao.findNonRemovedMap(id, consumedServiceId) != null) {
+                ValidationErrorCodes.throwValidationError(ValidationErrorCodes.NOT_UNIQUE,
+                        ServiceDiscoveryConstants.FIELD_SERVICE_ID);
+            }
+        } else {
+            if (consumeMapDao.findMapToRemove(id, consumedServiceId) == null) {
+                ValidationErrorCodes.throwValidationError(ValidationErrorCodes.INVALID_OPTION,
```
cjellick: Would `INVALID_REFERENCE` be a more appropriate error code?

resources/content/db/service_discovery.xml
```diff
@@ -0,0 +1,213 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
cjellick: You'll rename before merge, right?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -0,0 +1,235 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        Service service = (Service) state.getResource();
+        List<Integer> consumedServiceIds = new ArrayList<>();
+        boolean activateConsumedServices = DataAccessor.fromMap(state.getData()).withScope(ServiceActivate.class)
+                .withKey(ServiceDiscoveryConstants.FIELD_ACTIVATE_CONSUMED_SERVICES).withDefault(false)
+                .as(Boolean.class);
+        if (activateConsumedServices) {
+            // the order is determined only based on volumes_from (service) attribute
```
cjellick: I don't understand this bit of logic. Might need you to explain it to me.

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
alena1108: @cjellick I already override it in the API (code/iaas/model/src/main/resources/META-INF/cattle/core-model/defaults.properties):

object.link.name.serviceconsumemaprecord.consumedserviceid.override=consumedservices
....
I prefer to keep it as a map in the DB as the word "map" reflects the purpose of the table.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceAddRemoveLinkServiceValidationFilter.java
alena1108: @cjellick you are right :)

resources/content/db/service_discovery.xml
alena1108: @cjellick yes, this name is temporary so I don't run into merge conflicts while doing rebase

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
alena1108: @cjellick Service launch order within the environment, doesn't matter in most of the cases as upon service activation we do DNS configuration as a late binding. Although if service1 uses service2's volumes (volumes_from parameter in docker-compose.yml), its important that the service2 is activated first. Parameter FIELD_ACTIVATE_CONSUMED_SERVICES in the handler indicates that this parameter has been set, and consumed services for the target service have to be activated first. Although I don't specifically extract the consumed service referenced via volumes_from; I just activate all the consumed services - might revise it later.

cjellick: OK. makes sense.

cjellick:
LGTM.  Squash, rebase, rename schema file and we can merge

PR #290, opened 2015-03-06
Machine extracted-config functionality

cjellick:
Adds the ability for go-machine-service to update machine resources
with a base64 encoded tar file that is the docker-machine's
configuration and the ability for the user to download this
configuration as a tar file from a link on the resource.


cjellick:
@vincent99 FYI The link for downloading machine config if available.

PR #304, opened 2015-03-12
container events

cjellick:


cjellick:
@ibuildthecloud I could use some feedback on this PR.

This creates aa /containerevents resource that the **agent** will post to. The thing I want your feedback on the most is the ContainerEventFilter (https://github.com/rancherio/cattle/pull/304/files#diff-0469ac116f45b2c10f4753864e8f318fR1).

Since an agent is posting an event that will ultimately cause resources to be created on behalf a of a real user, I do a lot of validation to ensure the agent owns the host it is reporting the event for.

The other tricky part is that these ContainerEvents will have the accountId of the agent, not the user account that owns the host. I'd prefer the accountId to be that of the user, but the API framework won't allow me to override accountId in my filter. And I think that is the right behavior. So I was thinking we either just leave the the accountId as the agent's account id or in a process handler, we update the account id on the container event to be that of the user's (which we could get through the Host).

Thoughts?

cjellick:
Closing because this is replaced by #315

PR #308, opened 2015-03-13
Swarm integration

sonchang:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/eventing/impl/AllocatorEventListenerImpl.java
```diff
@@ -51,11 +62,35 @@ protected void allocate(Event event) {
...
-                break;
+        // Check resourceType for instance and whether the host is a cluster or not.
+        // If resourceType is a cluster (host), do not bother with normal allocation process.
+
+        // Might be able to change this to: request.getType() == AllocationRequest.Type.INSTANCE although they check different things
+        if ("instance".equals(event.getResourceType())) {
+            final Instance instance = objectManager.loadResource(Instance.class, request.getResourceId());
+
+            // check requestedHostId to see whether it's a cluster or not
+            Long requestedHostId = DataAccessor.fields(instance).withKey(InstanceConstants.FIELD_REQUESTED_HOST_ID).as(Long.class);
```
ibuildthecloud: Let's try moving this to the InstanceAllocate class

resources/content/config-content/cluster/content-home/run_swarm.sh.ftl
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+if [ -e /var/run/swarm.pid ]
+then
+    kill -TERM $(cat /var/run/swarm.pid)
+fi
+/root/go/bin/swarm manage ${discoverySpec} --strategy random -host=0.0.0.0:${clusterServerPort} & echo $! > /var/run/swarm.pid
```
ibuildthecloud: assume swarm is in the path

ibuildthecloud: use monit for supervision

sonchang:
@ibuildthecloud Some new modifications since we last spoke.
- After moving the logic to InstanceActivate, I found out there was another step missing besides creating the instanceHostMap entry.  Basically, I needed to upset the nic with the subnet as well.  I noticed this when testing with Craig's managed network stuff.
- I've also changed to use monit and init.d for managing the swarm server
- A bunch of other refactors as well

What's still missing:
- Use real docker image for deploying swarm (rather than my personal one)

sonchang:
@ibuildthecloud Can you merge after the python-agent PR has been merged?  This now uses rancher/agent-instance:v0.3.0 instead of my personal docker image.

sonchang:
@ibuildthecloud Please merge

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/eventing/impl/AllocatorEventListenerImpl.java
```diff
@@ -25,6 +26,9 @@
     List<Allocator> allocators;
     EventService eventService;
 
+    @Inject
+    ObjectManager objectManager;
```
alena1108: Unused inject annotation?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceAllocate.java
```diff
@@ -30,6 +51,43 @@ public InstanceAllocate() {
...
+
+        // check requestedHostId to see whether it's a cluster or not
+        Long requestedHostId = DataAccessor.fields(instance).withKey(InstanceConstants.FIELD_REQUESTED_HOST_ID).as(Long.class);
+        if (requestedHostId != null) {
+            Host potentialCluster = objectManager.loadResource(Host.class, requestedHostId);
+            if (objectManager.isKind(potentialCluster, ClusterConstants.KIND)) {
+                log.info("Instance [{}] requested for a cluster: [{}]", instance, potentialCluster);
+
+                // create instanceHostMap binding with cluster.
+                objectManager.create(InstanceHostMap.class,
```
alena1108: We should check if the mapping already exists before creating it. For that, you can use genericMapDao.findNonRemoved method

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/ClusterHostMapDaoImpl.java
```diff
@@ -41,4 +47,26 @@ public ClusterHostMapRecord getClusterHostMap(Host cluster, Host host) {
...
+                .join(HOST_IP_ADDRESS_MAP)
+                    .on(IP_ADDRESS.ID.eq(HOST_IP_ADDRESS_MAP.IP_ADDRESS_ID))
+                .where(HOST_IP_ADDRESS_MAP.HOST_ID.eq(hostId)
+                    .and(HOST_IP_ADDRESS_MAP.REMOVED.isNull())
+                    .and(IP_ADDRESS.REMOVED.isNull()))
+            .fetchOneInto(IpAddressRecord.class);
+    }
+
+    @Override
+    public Host findHostByName(String name) {
```
alena1108: Looks like cattle allows duplicated names for the host, so this record might result in fetching just one record out of many. If there is a backend restriction on the hostName to be unique within the cluster, we define this restriction on the cattle side as well.

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+                    return null;
+                }
+            });
+        }
+    }
+
+    @Override
+    public Host getManagingHost(Host cluster) {
+        Long managingHostId = DataAccessor.fields(cluster).withKey(ClusterConstants.MANAGING_HOST).as(Long.class);
+        if (managingHostId == null) {
```
alena1108: If managing host can be set to null at any point, or the referenced managing host gets removed, we should handle it in the ProcessHandler. Perhaps by assigning another managing host to it?

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+        return ip;
+    }
+
+    private void start(final Instance agentInstance) {
+        if (InstanceConstants.STATE_STOPPED.equals(agentInstance.getState())) {
+            DeferredUtils.nest(new Callable<Object>() {
+                @Override
+                public Object call() throws Exception {
```
alena1108: "executeProcess" along with the initial state check here will not work when idempotent checks are on. Lets say the initial state check has passed, and the state got changed to Starting, but then the process has failed. In this case instance will get stuck in Starting state forever as the idempotent retry won't try to execute the process again as the object's state != Stopped.

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
```diff
@@ -0,0 +1,97 @@
...
+        return lockManager.lock(new ClusterLock(cluster), new LockCallback<HandlerResult>() {
+            @Override
+            public HandlerResult doWithLock() {
+                // check if this cluster's already managed by a specific host
+                Long hostId = DataAccessor.fields(cluster).withKey(ClusterConstants.MANAGING_HOST).as(Long.class);
+
+                // if not, find suitable host within cluster to manage the cluster server
+                // This actually doesn't check whether the current host is actually in the cluster. Which isn't really a requirement anyways.
+                if (hostId == null || !isHostActive(hostId)) {
+                    hostId = findSuitableHost(cluster);
```
alena1108: findSuitableHost can return null. Should we put a check for that here?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceAllocate.java
```diff
@@ -30,6 +51,43 @@ public InstanceAllocate() {
...
+                        INSTANCE_HOST_MAP.HOST_ID, requestedHostId,
+                        INSTANCE_HOST_MAP.INSTANCE_ID, instance.getId());
+
+                // update nic with active subnet
+                Nic nic = nicDao.getPrimaryNic(instance);
+                Network network = objectManager.loadResource(Network.class, nic.getNetworkId());
+                if (network != null) {
```
alena1108: If the network has multiple subnets, is always picking the first one is the correct behavior? Can we re-use NetworkAllocationCandidates allocator logic here?

alena1108:
1) We have to handle the cleanup on the following events:

* cluster.remove - all the host-cluster mappings should be gone as the result, and the Swarm agent has to be cleaned up from the host
* host.remove - all related host-cluster mapping(s) should be cleaned up. And if the host used to be the manager of the cluster, the management has to be reassigned to another host in that cluster.

2) Re-assigning managing host to cluster. Make sure that when the managing host is gone from the cluster - either as a result of host.removeHost from cluster, or host.removeHost - the new management host has to be reassigned. 

3) As Swarm agent instance can be removed via API, you probably need to add a logic for recreating it on some event. It can either be a periodic task checking the cluster health, or instance deployment in the cluster.

code/iaas/model/src/main/java/io/cattle/platform/core/constants/InstanceConstants.java
```diff
@@ -4,7 +4,9 @@
 
     public enum SystemContainer {
         NetworkAgent,
-        LoadBalancerAgent;
+        LoadBalancerAgent,
+        ClusterDockerAgent, // socat
```
alena1108: As we only have one system instance for the cluster right now, only one enum value should be listed.

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
sonchang: Actually, if it does return null, we still want to update the managingHost field to be null since that's more accurate.  We still check 'hostId' whether it's null or not in the next line and do the appropriate thing by deactivating the cluster

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
sonchang: Thanks.  Will be fixed in next commit

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceAllocate.java
sonchang: Thanks!

code/iaas/model/src/main/java/io/cattle/platform/core/constants/InstanceConstants.java
sonchang: Removed.  Other one will be added back in a future PR

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/ClusterHostMapDaoImpl.java
sonchang: I think I'll have to address this in a subsequent PR.  Maybe we can file an issue for it so we don't lose track of it.

alena1108: Son, can you file it and update this PR with the ticket #?

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private String getConnectionPrefix(Host cluster, Host managingHost) {
+        return objectManager.isKind(managingHost, "sim") ? "sim:" : "delegate:";
+    }
+
+    private String getImagePrefix(Host cluster, Host managingHost) {
+        return objectManager.isKind(managingHost, "sim") ? "sim:" : "docker:";
+    }
+
+    private Long getNetworkIds(Host managingHost) {
```
alena1108: I extracted a method for retrieving networkId and changed its argument to generic object, as a part of the PR https://github.com/rancherio/cattle/pull/286. So we don't have to repeat this logic for every object type in respective managers. If my PR gets merged first, you re-use this logic here. If not, it should be fixed in a separate PR.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceAllocate.java
sonchang: I think we should still be good here. This code was actually pulled from NetworkAllocationCandidates.  NetworkAllocationCandidates creates a bunch of AllocationCandidates with the different subnets available.  But ultimately, I believe the instance allocator just picks the first candidate that fits all the different criterias (the network constraint being one of the criteria)

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
alena1108: Then I would separate the cluster deactivation logic to its separate handler. The handler should be invoked when managing host is set to null for the cluster. And the latter can happen when:

1) host is deactivated
2) host is removed from cluster
3) host is removed from cattle

sonchang: This currently just schedules it for deactivation.  The deactivation logic already resides in HostDeactivate (since a cluster is also considered a host)

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/ClusterHostMapDaoImpl.java
sonchang: Created https://github.com/rancherio/cattle/issues/326 for this issue

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
alena1108: Sorry, rephrasing - scheduling deactivation logic (lets call it process #1) can be put to the separate handler invoked on managing host set to Null (process #2). Because process #2 itself can be triggered by:

1) host is deactivated
2) host is removed from cluster
3) host is removed from cattle
4) ... and probably more

Just to make the code cleaner, otherwise you will have to duplicate executing process #1 and scheduling process #2 in 1) 2) 3) and 4)

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/manager/ClusterManager.java
```diff
@@ -0,0 +1,77 @@
...
+                JooqUtils.getTableField(getMetaDataManager(), "instanceHostMap", InstanceHostMapTable.INSTANCE_HOST_MAP.HOST_ID);
+
+        TableField<?, Object> chmHostId =
+                JooqUtils.getTableField(getMetaDataManager(), "clusterHostMap", ClusterHostMapTable.CLUSTER_HOST_MAP.HOST_ID);
+        TableField<?, Object> chmRemoved =
+                JooqUtils.getTableField(getMetaDataManager(), "clusterHostMap", ClusterHostMapTable.CLUSTER_HOST_MAP.REMOVED);
+        TableField<?, Object> clusterId =
+                JooqUtils.getTableField(getMetaDataManager(), "clusterHostMap", ClusterHostMapTable.CLUSTER_HOST_MAP.CLUSTER_ID);
+
+        org.jooq.Condition cond1 = instanceId.eq(ihmInstanceId).and(ihmRemoved == null ? DSL.trueCondition() : ihmRemoved.isNull());
```
cjellick: I dont quite understand what you're doing here. The way I read this, you would end up with this kind of sql:
if ihmRemoved == null:
`
...where instance_id = ihm.instance_id and true
`
if ihmRemoved != null
`
...where instance_id = ihm.instance_id and ihm.removed is null
`
So the two questions/comments/issues I have are:
1. Are you doing the ternary + trueCondition() to keep that java code short or is there some other reason?
2. In what scenario would `ihmRemoved` ever be null? Doesn't it represent a table column that should always be there?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/manager/ClusterManager.java
```diff
@@ -0,0 +1,77 @@
...
+        org.jooq.Condition cond2 = ihmHostId.eq(chmHostId).and(chmRemoved == null ? DSL.trueCondition() : ihmRemoved.isNull());
```
cjellick: Do you mean ` chmRemoved.isNull())` instead of ` ihmRemoved.isNull())`?

cjellick: Why fully qualified name for `org.jooq.Condition`?

code/iaas/cluster-logic/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-framework-jooq</artifactId>
+            <version>${project.version}</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-iaas-agent-instance</artifactId>
+            <version>${project.version}</version>
+            <optional>true</optional>
```
cjellick: Dont think this should be optional.

code/iaas/cluster-logic/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-iaas-storage-service</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-framework-jooq</artifactId>
```
cjellick: Is this dep needed? Code didnt have compile warnings without it.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/manager/ClusterManager.java
sonchang: Cut/paste artifact from abstract class.  I can change this

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/manager/ClusterManager.java
sonchang: Much of this was also patterned after AbstractJooqResourceManager.
Looking at it now, I can definitely simplify it since we already know we have a removed field in the DB.  I guess in the abstract class, maybe that wasn't necessarily the case so they had that extra logic.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/manager/ClusterManager.java
sonchang: I did mean chmRemoved.isNull().  Thanks for catching this!!!

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivate.java
```diff
@@ -1,19 +1,65 @@
...
+        if (objectManager.isKind(state.getResource(), ClusterConstants.KIND)) {
+            return handleCluster(state, process);
+        } else {
+            return handleHost(state, process);
+        }
+    }
+
+    private HandlerResult handleCluster(ProcessState state, ProcessInstance process) {
+        final Host cluster = (Host)state.getResource();
+        lockManager.lock(new ClusterLock(cluster), new LockCallback<Instance>() {
```
cjellick: Could probably use `LockCallbackNoReturn`, but not a blocker. Up to you if you want to change.

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { "host.activate" };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        if (!objectManager.isKind(state.getResource(), ClusterConstants.KIND)) {
+            return new HandlerResult(false, (Map<Object, Object>)null);
```
cjellick: Could just return null or do you need to explicitly pass shouldContinue== false for some reason?

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
alena1108: Not a blocker, can be implemented as a part of the PR where you handle cluster/host cleanup and cluster's manager change use cases

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivatePreListener.java
```diff
@@ -0,0 +1,97 @@
...
+                return new HandlerResult(true, (Map<Object, Object>)null);
+            }
+        });
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private void updateManagingHost(Host cluster, Long managingHostId) {
+        Map<String, Object> newData = new HashMap<String, Object>();
+        newData.putAll(cluster.getData());
+        ((Map)newData.get("fields")).put(ClusterConstants.MANAGING_HOST, managingHostId);
```
cjellick: Is `DataUtils.getWriteableFields()` a 'safer' way to do this?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceAllocate.java
alena1108: @sonchang ok, should be good then. Below is the explanation why this new class instance.allocate was introduced (for other peoples reference - please correct if anything is missing)

Without Swarm:
=================
1) Instance.create API is called, Instance.create handler logic creates the nics/volumes db entries for the instances and schedules instance.start

2) Instance.start triggers allocators to kick off to allocate the host and the network for the container. Subnet got picked and allocated to a Nic and target host gets picked for the container as a part of this logic.

3) After allocation is done, instance.start handler triggers nic.activate for all container's nics. As a part of nic.activate, vnet gets allocated to the nic.

With Swarm
=================================
Everything above is applied, except the allocation logic. Allocation logic is moved to instance.allocate, and host/subnet are set explicitly on the container/nic in instance.allocate

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/ClusterManager.java
```diff
@@ -0,0 +1,22 @@
...
+import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.model.IpAddress;
+
+public interface ClusterManager {
+
+    Instance createClusterServerInstance(Host cluster);
+
+    Instance getClusterServerInstance(Host cluster);
+
+    Host getClusterForInstance(Instance clusterServerInstance);
```
cjellick: This method unused and possibly a hold over from when the plan was to create an agent.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceHostMapAgentProcessHandler.java
```diff
@@ -0,0 +1,108 @@
...
+                "instanceHostMap", "instance", "+data", "dockerContainer", "Names");
+
+        if (names != null) {
+            for (Object name : names) {
+                if (actualHost != null) break;
+
+                if (name == null) continue;
+
+                String[] nameParts = ((String)name).split("/");
+                if (nameParts.length > 2) {
```
cjellick: Just checking: you didnt want `>=` instead of `>` here, did you?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceHostMapAgentProcessHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    protected void postProcessEvent(EventVO<?> event, Event reply, ProcessState state, ProcessInstance process,
+            Object eventResource, Object dataResource, Object agentResource) {
+        Host actualHost = null;
+
+        List names = (List)CollectionUtils.getNestedValue(reply.getData(),
+                "instanceHostMap", "instance", "+data", "dockerContainer", "Names");
+
+        if (names != null) {
+            for (Object name : names) {
+                if (actualHost != null) break;
```
cjellick: What is the one-line if-break witchcraft?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceHostMapAgentProcessHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        return super.handle(state, process);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Override
+    protected void postProcessEvent(EventVO<?> event, Event reply, ProcessState state, ProcessInstance process,
+            Object eventResource, Object dataResource, Object agentResource) {
+        Host actualHost = null;
+
+        List names = (List)CollectionUtils.getNestedValue(reply.getData(),
```
cjellick: An alternative to the host name solution might be to always add the host uuid to the response event. It should always uniquely identify the host.

AgentResourceMonitorImpl makes use of it.
https://github.com/rancherio/cattle/blob/c552571c3d2003446cc5af9269563cc4da3258db/code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/resource/impl/AgentResourcesMonitorImpl.java

NOTE: host uuid and physical host uuid are two different things. You would not want physcial host.

cjellick: Or maybe that wont work. This response doesn't come from the real host, does it?

resources/content/config-content/cluster/content/etc/init.d/cluster
```diff
@@ -0,0 +1,39 @@
...
+### BEGIN INIT INFO
+# Provides:		cluster
+# Default-Start:	2 3 4 5
+# Default-Stop:		0 1 6
+# Short-Description:	swarm cluster
+# Description:		swarm cluster
+### END INIT INFO
+
+PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
+SWARM_ARG_FILE=/etc/swarm_args.conf
```
cjellick: Is the base of the /etc dir the best place for this?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceHostMapAgentProcessHandler.java
sonchang: Haha.  Must've been leftover code from a more complex time.  Thanks for finding it.  Fixed.

resources/content/config-content/cluster/content/etc/init.d/cluster
sonchang: Not sure.  There's a bunch of other config files currently in the base of /etc
I was tempted to maybe create a /etc/cluster directory and put it in there

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceHostMapAgentProcessHandler.java
sonchang: This response doesn't come from the actual host the container was deployed to.  It's still coming from the managing host (host where the swarm server is running).  This was an area I was planning to revisit to see if there's a better way.

sonchang:
Filed issue https://github.com/rancherio/cattle/issues/328 for detecting/resolving a deactivated or downed cluster server prior to deploying a container to it

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceHostMapAgentProcessHandler.java
sonchang: I can't seem to find the earlier comment regarding > 2, I just tried it out again and the reason it's > 2 rather than >=2 is because there's a leading / in the string.   So the first value is just empty.

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/process/ClusterUpdateConfig.java
```diff
@@ -0,0 +1,146 @@
...
+            return null;
+        }
+        final Host finalCluster = cluster;
+        final ProcessState finalProcessState = state;
+        lockManager.lock(new ClusterLock(cluster), new LockCallbackNoReturn() {
+            @Override
+            public void doWithLockNoResult() {
+                if (CommonStatesConstants.ACTIVE.equals(finalCluster.getState()) ||
+                        CommonStatesConstants.ACTIVATING.equals(finalCluster.getState())) {
+                    updateDiscoverySpec(finalCluster);
```
alena1108: We shouldn't load the discovery spec on every host add/activate for the cluster as this value is hardcoded anyway. Just do it once.

code/implementation/host-only-network/src/main/java/io/cattle/platform/networking/host/process/HostOnlyNicActivate.java
```diff
@@ -50,24 +55,28 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
             }
 
             createIgnoreCancel(vnet, null);
 
             boolean mapped = false;
             for (HostVnetMap map : mapDao.findNonRemoved(HostVnetMap.class, Vnet.class, vnet.getId())) {
-                if (map.getHostId().equals(host.getId())) {
+                if (map.getHostId().equals(actualHost.getId())) {
                     createThenActivate(map, state.getData());
                     mapped = true;
```
alena1108: Should we put a "break" here?

sonchang: Will add break

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/manager/ClusterManager.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.jooq.Condition;
+import org.jooq.Table;
+import org.jooq.TableField;
+import org.jooq.impl.DSL;
+
+public class ClusterManager extends AbstractJooqResourceManager {
```
alena1108: We should rename the class to ClusterResourceManager

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostActivate.java
```diff
@@ -1,19 +1,77 @@
...
+            @Override
+            public void doWithLockNoResult() {
+                // short term optimization to avoid updating cluster object unnecessarily
+                // since we're just currently only supporting file:// discoverySpec
+                if (StringUtils.isEmpty(DataAccessor.fieldString(cluster, ClusterConstants.DISCOVERY_SPEC))) {
+                    DataUtils.getWritableFields(cluster).put(ClusterConstants.DISCOVERY_SPEC, "file:///etc/cluster-hosts.conf");
+                    objectManager.persist(cluster);
+                }
+
+                Instance clusterServerInstance = clusterManager.createClusterServerInstance(cluster);
```
alena1108: This should be moved to the ClusterManagerImpl as creating the cluster service instance is a prereq for server config update

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/process/ClusterHostMapCreateRemoveHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { "clusterhostmap.create", "clusterhostmap.remove" };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        Long clusterId = ((ClusterHostMap) state.getResource()).getClusterId();
+        Host cluster = objectManager.loadResource(Host.class, clusterId);
```
alena1108: I see a potential problem here for the clusterhostmap.remove. If the host that is being removed is a master host, the updated config won't have this host specified. So you will end up updating the master with the config not having the master. Unless master host is reassigned by this point. I think we should add a pre-handler for clusterhostmap.remove checking if the host-to-remove is a master, then reassigning an another master, and only then updating the config. But you can do it as part of a PR that is gonna take care of all the cleanup/management reassignment cases. By now, just make sure that config update lacking master, doesn't result in the error on the backend.

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostRemove.java
```diff
@@ -0,0 +1,116 @@
...
+import io.cattle.platform.process.util.ProcessHelpers;
+
+import java.util.HashMap;
+import java.util.List;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+public class HostRemove  extends AbstractDefaultProcessHandler{
```
alena1108: Do you think it would make sense to make it a pre-listener to the host.remove? So it would be easier to separate cluster related logic if we ever decide to move swarm to its separate project.

resources/content/config-content/cluster/content/etc/init.d/cluster
alena1108: /etc/cluster would be the better place. As most of the existing config items - monit, haproxy, ipsec - have dedicated folders under /etc dir

sonchang: I'll move it

sonchang:
@alena1108 @cjellick Sorry for all the iterations.  Some of these bug fixes required quite a bit of refactoring.  But I think I have all the cases working now.  So no more changes from me unless you guys find something.

alena1108:
LGTM

cjellick:
LGTM

PR #309, opened 2015-03-17
Bigints should be 20

hibooboo2:


vincent99:
You have (implicitly) signed bigints, which cannot be more than 19 digits wide (the negative sign is not included).  Anwyay, that number is just a display width hint to pad out ASCII tables in the client, it doesn't actually change anything about how the numbers are stored or the range of values the column can hold. (I assume we don't use ZEROFILL anywhere).

hibooboo2:
Well vincent i only say they should be 20 because they are everywhere else in the database except in like 3 or 4 places where craig had recently made a table using this script.

ibuildthecloud:
@hibooboo2 Liquibase probably screws this all up with in generates the XML, I would guess.

PR #310, opened 2015-03-19
Update boot2docker script to use API

cjellick:


cjellick:
@hibooboo2 I updated the register-boot2docker script to use the API to get the "proper" `docker run` command. If you have the time/inclination, do you want to try this out and see if it works for you as well?

If you don't/didn't use register-boot2docker.sh as part of your workflow, don't worry about trying it out.

cjellick:
...It now requires jq. Not sure if that's a deal breaker.

tools/development/register-boot2docker.sh
```diff
@@ -11,10 +10,14 @@ fi
 # This is just here to make sure your environment is sane
 docker info
 
-CONSOLE_ARGS=""
-if [ -t 1 ]; then
```
cloudnautique: These allow interrupting the process while its running. Something ibuildthecloud does... Doesn't look like they are being set/used

cjellick: Ah, was curious about why those were needed. Thanks for clearing up. I can add them back in.

ibuildthecloud:
The curl commands don't need to be so ugly.  The api supports multi part forms so you can do, "curl -F key=value"

cjellick:
Ping @ibuildthecloud and @cloudnautique.
Fixed it up. Can you guys take another look?

cloudnautique:
We need Jq added to bootstrap. This script is called in ./scripts/ci

cjellick:
I incorrectly assumed that this script was only used for registering boot2docker instances. Will probably need to refactor more.

PR #311, opened 2015-03-19
Service Discovery: DNS config item/Json Answer file generation

alena1108:
@ibuildthecloud Darren, this PR introduces new config item - "dns" - and the logic for this config item update / dns answers.json file generation. The config file ends up on the network-agent instance. The file doesn't get applied, and there dns service is not installed on the network-agent. 

What goes to answers.json file:

* data generated based on instance links for service's instances
* data generated based on consumed services for service's instances
* for every instance of service "foo", dns records containing ips of all other instances of service "foo"

DNS data gets generated just for the containers that participate in the Service Discovery service. Regular instances are not affected.

### DIFF
--- a/resources/content/config-content/dns/content/etc/dns/answers.json.ftl
+++ b/resources/content/config-content/dns/content/etc/dns/answers.json.ftl
@@ -1,0 +1,19 @@
+{
+    "default": {
+        "recurse": ["8.8.8.8"],
+        "rancher.com.": ["1.2.3.4"]
+    },
+<#if dnsEntries?? && dnsEntries?has_content>
+    <#list dnsEntries as dnsEntry>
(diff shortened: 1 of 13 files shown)


resources/content/config-content/dns/content/etc/dns/answers.json.ftl
```diff
@@ -0,0 +1,19 @@
+{
+    "default": {
+        "recurse": ["8.8.8.8"],
+        "rancher.com.": ["1.2.3.4"]
```
vincent99: `rancher.com.` was just an example & should be removed.  The recurse addresses should probably be pulled from /etc/resolv.conf or something instead of being hardcoded to Google.

PR #316, opened 2015-03-23
Add flag to enable bootstrap on run

cloudnautique:
In the latest environment we need the ability to run MySQL with the run command. Also, if we exit the trap will attempt to shut down MySQL cleanly.


cjellick:
LGTM

PR #321, opened 2015-03-24
LB: merged setHosts/setTargets to a single API - "updateAll"

alena1108:
@vincent99 @ibuildthecloud


code/iaas/logic/src/main/java/io/cattle/platform/process/lb/LoadBalancerUpdateAll.java
```diff
@@ -0,0 +1,33 @@
...
+import io.cattle.platform.process.common.handler.AbstractObjectProcessHandler;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+@Named
+public class LoadBalancerUpdateAll extends AbstractObjectProcessHandler {
+
+    @Inject
+    LoadBalancerSetHosts setHosts;
```
cjellick: Having properties that begin with `set` seems like it could lead to confusion, since it clashes with POJO get/set naming convention.

Not a blocker, up to you if you want to change it. Just wanted to point it out.

cjellick:
Do you need a test that sets both hosts and targets in the same api call?

alena1108:
@cjellick you are right, I will add the test

alena1108:
@cjellick Craig, the test is in. @vincent99 Vince, can you also review it as I need 2 LGTMs for it to get merged?

cjellick:
LGTM ping @vincent99

vincent99:
I don't know about the java-yness specifics but this looks like the change we talked about.

will-chan:
@vincent99 @ibuildthecloud @alena1108 

The PR LTGM.

However, while I do understand why we need an updateall type API command for UI convenience, I do have two issues with this:

1. updateall basically re-maps everything so if I only needed to to append several items to an existing LB, I need to re-add all my previous targets/hosts.  You can argue you should use the individual addXXX/removeXXX commands but it is dangerous.  

2. if you are going to support a multi-item update command, why not also support that for the add/remove targets/hosts command?

vincent99:
Answering backwards..

2. Changing {add,remove}{Host,Target} to take an `array[reference[...]]` would be nice (but the UI won't need it).  Is this going to be idempotent, or does saying add addTarget twice cause 2 entries to show up?

1. Yes, if you only need to append, yes the multi-item add would be better than get/modify/set.

alena1108:
@vincent99 1) - it will be idempotent

PR #323, opened 2015-03-25
Native docker

cjellick:


code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/resource/impl/AgentResourcesMonitorImpl.java
```diff
@@ -49,6 +51,7 @@
     private static final String[] UPDATABLE_HOST_FIELDS = new String[] { HostConstants.FIELD_API_PROXY, HostConstants.FIELD_INFO };
 
     PingDao pingDao;
+    AgentDao agentDao;
```
alena1108: Should use @Inject instead of setting the var with setter? Less code

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/containerevent/ContainerEventFilter.java
```diff
@@ -0,0 +1,68 @@
...
+        return new Class<?>[] { ContainerEvent.class };
+    }
+
+    @Override
+    public Object create(String type, ApiRequest request, ResourceManager next) {
+        ContainerEvent event = request.proxyRequestObject(ContainerEvent.class);
+
+        Policy policy = ApiUtils.getPolicy();
+        Agent agent = null;
+        try {
```
alena1108: @cjellick how do we recover from situation when we have more than one agent?

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/dao/impl/PingInstancesMonitorDaoImpl.java
```diff
@@ -28,9 +26,9 @@
                     .on(INSTANCE.ID.eq(INSTANCE_HOST_MAP.INSTANCE_ID))
                 .where(INSTANCE.STATE.eq(InstanceConstants.STATE_RUNNING)
                         .and(AGENT.ID.eq(agentId)))
```
alena1108: Should we add Removed=null as a Agent search condition?

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -79,25 +79,22 @@ public void computeInstanceActivateReply(Event event) {
...
-    protected boolean different(long agentId, Set<String> knownInstances, Set<String> instances, boolean check) {
-        instances = new HashSet<String>(instances);
-        for (String instance : knownInstances) {
-            if (!instances.remove(instance)) {
-                if (check) {
+    
+    protected boolean handleUnreportedKnownInstances(long agentId, Map<String, String> knownInstances, 
+            ReportedInstances reportedInstances, boolean checkOnly) {
+
+        for (Map.Entry<String, String> knownInstance : knownInstances.entrySet()) {
```
alena1108: Can knownInstances be null? If so, need to add not null check before iterating

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -115,36 +112,49 @@ public boolean evaluate(ProcessState state, ProcessInstance processInstance, Pro
...
+    }
+
+    protected class ReportedInstances {
+        Set<String> byUuid = new HashSet<String>();
+        Set<String> byExternalId = new HashSet<String>();
     }
 
-    protected Set<String> load(Long agentId) {
-        return monitorDao.getHosts(agentId);
+    protected Map<String, String> load(Long agentId) {
```
alena1108: Autoboxing! :) should either check that the agentId is not null, or change the method signature to accept primitive long as thats what MonitorDao expects.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,207 @@
...
+                        INSTANCE.EXTERNAL_ID, externalId);
+                if ( instance == null ) {
+                    instance = objectManager.newRecord(Instance.class);
+                    instance.setKind(InstanceConstants.KIND_CONTAINER);
+                    instance.setAccountId(accountId);
+                    instance.setExternalId(externalId);
+                    setName(inspect, instance);
+                    setNetwork(accountId, inspect, instance);
+                    setImage(inspect, instance);
+                    setHost(event, instance);
```
alena1108: @cjellick is there a particular reason why we schedule instance restart on the same host instead of delegating this decision to allocators? I would set it only in case that user set requestedHostId on instance during its original start. Otherwise limiting it to the same host can cause failure situation to repeat - if the host is in some failure condition for example.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,207 @@
...
+                    DataAccessor.fields(instance).withKey(InstanceConstants.FIELD_NATIVE_CONTAINER).set(Boolean.TRUE);
+                    instance = objectManager.create(instance);
+                }
+
+                if ( !InstanceConstants.STATE_RUNNING.equals(ObjectUtils.getState(instance)) &&
+                        !InstanceConstants.STATE_STARTING.equals(ObjectUtils.getState(instance))) {
+                    try {
+                        getObjectProcessManager().scheduleProcessInstance(InstanceConstants.PROCESS_CREATE, instance, makeData());
+                    } catch (ProcessCancelException e) {
+                        // TODO Figure out if i need this catch
```
alena1108: I usually put try/catch ProcessCancelException when I schedule the process for the object w/o verifying its original state

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,207 @@
...
+                    }
+                }
+            }
+        });
+    }
+
+    public static boolean isNativeDockerStart(ProcessState state) {
+        Boolean nativeDockerStart = DataAccessor.fromMap(state.getData()).withScope(ContainerEventCreate.class)
+                .withKey(InstanceConstants.DATA_NATIVE_DOCKER_START).as(Boolean.class);
```
alena1108: You can add "withDefault=false" condition to this statement. It will set the value to false in case its null

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,207 @@
...
+        instance.setName(name);
+    }
+
+    void setNetwork(Long accountId, Map<String, Object> inspect, Instance instance) {
+        String networkKind = NetworkConstants.KIND_NETWORK;
+        if ( BooleanUtils.toBoolean(getRancherNetworkLabel(inspect)) ) {
+            networkKind = NetworkConstants.KIND_HOSTONLY;
+        }
+
+        List<? extends Network> networks = networkDao.getNetworksForAccount(accountId, networkKind);
```
alena1108: @cjellick you can reuse that generic method for finding generic object's network - I've added it to NetworkDao

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,207 @@
...
+    @SuppressWarnings("unchecked")
+    String getLabel(String key, Map<String, Object> inspect) {
+        try {
+            Map<String, Object> config = (Map<String, Object>)inspect.get(INSPECT_CONFIG);
+            List<String> envVars = (List<String>)config.get(INSPECT_ENV);
+            for ( String envVar : envVars ) {
+                if ( envVar.startsWith(key) ) {
+                    return envVar.substring(key.length());
+                }
+            }
```
alena1108: What can cause NPE here, and is it safe just to catch it silently?

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventInstanceLock.java
```diff
@@ -0,0 +1,11 @@
+package io.cattle.platform.process.containerevent;
+
+import io.cattle.platform.lock.definition.AbstractBlockingLockDefintion;
+
+public class ContainerEventInstanceLock extends AbstractBlockingLockDefintion {
+
+    public ContainerEventInstanceLock(Long accountId, String externalId) {
```
alena1108: Should it be primitive long instead for accountId?

resources/content/schema/base/container.json
```diff
@@ -189,6 +189,10 @@
         },
         "registryCredentialId":{
             "type":"reference[registryCredential]"
+        },
```
alena1108: @cjellick Do you think users might wanna do search by nativeContainer attribute? If so, we should probably move it to its own field in the DB instead of putting it to data field

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/containerevent/ContainerEventFilter.java
```diff
@@ -0,0 +1,68 @@
...
+            agent = objectManager.findOne(Agent.class, AGENT.ACCOUNT_ID, policy.getAccountId());
+        } catch (InvalidResultException e) {
+            // Found more than one agent, reject request just as if we didn't
```
sonchang: Is this a valid scenario?  If not, maybe we should wrap it into a RuntimeException so that it bubbles all the way up and forces us to figure out what went wrong?

tests/integration/tox.ini
```diff
@@ -4,7 +4,7 @@ envlist=flake8, py27
 [testenv]
 deps=-rrequirements.txt
 changedir=cattletest
-commands=py.test --durations=20 {posargs}
+commands=py.test --durations=20 {posargs} -k test_native_docker
```
sonchang: Is this supposed to be part of the PR?

cjellick: Nope. Thanks.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,207 @@
...
+import javax.inject.Named;
+
+import org.apache.commons.lang3.BooleanUtils;
+
+import com.netflix.config.DynamicBooleanProperty;
+
+@Named
+public class ContainerEventCreate extends AbstractDefaultProcessHandler {
+
+    private static final DynamicBooleanProperty MANAGE_NONRANCHER_CONTAIERS = ArchaiusUtil
```
sonchang: just minor typos

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -79,25 +79,22 @@ public void computeInstanceActivateReply(Event event) {
...
                     return true;
                 } else {
-                    restart(instance);
+                    // If this known instance was not reported, schedule a potential restart.
+                    restart(knownInstance.getKey());
                 }
             }
         }
 
-        if (!check && COMPLAIN.get()) {
```
sonchang: We should probably do all the additional code cleanup if we plan to permanently remove this (removing the 'ha.instance.state.report.unknown' property)

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
sonchang: I'm guessing the swallowing of the NPE is just temporary.  We could use CollectionUtils.getNestedValue(...) at least for INSPECT_CONFIG and INSPECT_ENV.  Don't think it'll iterate through the List though.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/dao/impl/PingInstancesMonitorDaoImpl.java
cjellick: I don't think so. For the context in which this is called, we really just want all running instances associated with an agent, regardless of the agent's state.

code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/resource/impl/AgentResourcesMonitorImpl.java
cjellick: will do

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/containerevent/ContainerEventFilter.java
cjellick: @alena1108 @sonchang I'll remove the try catch and let it bubble.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: will add null check

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: will do

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventPreCreate.java
```diff
@@ -0,0 +1,69 @@
...
+        // event's account id is set to the agent that submitted. This will
+        // change it to the actual user's account id.
+        // Checks to make sure agent's resource account id and the host's
+        // account id match
+        ContainerEvent event = (ContainerEvent)state.getResource();
+        Agent agent = null;
+        try {
+            agent = objectManager.findOne(Agent.class, AGENT.ACCOUNT_ID, event.getAccountId());
+        } catch (InvalidResultException e) {
+            // Found more than one agent, don't do anything.
```
sonchang: I guess another Exception I'm not sure whether we want to bubble up or not (or possibly log)?

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: good catch will fix.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: The container already exists on this host, we're effectively backfilling the container to that host in rancher.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: Going to use it, but looks like I'll need to change the sig to `Network getNetworkForObject(Object object, String kind)`

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: INSPECT_ENV can be legitimately null. I'll just change that to traditional null check.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventPreCreate.java
cjellick: Will bubble.

resources/content/schema/base/container.json
cjellick: I suppose it couldn't hurt. I'll switch it. Good suggestion!

alena1108:
LGTM

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -0,0 +1,197 @@
...
+import javax.inject.Named;
+
+import org.apache.commons.lang3.BooleanUtils;
+
+import com.netflix.config.DynamicBooleanProperty;
+
+@Named
+public class ContainerEventCreate extends AbstractDefaultProcessHandler {
+
+    private static final DynamicBooleanProperty MANAGE_NONRANCHER_CONTAINERS = ArchaiusUtil
```
cjellick: @ibuildthecloud I currently have the feature turned off by default and you would have to change a setting to turn it on. Would you rather it be on by default?

cjellick:
got verbal LGTM from @ibuildthecloud

thaJeztah:
Great to see this merged! Thanks!

cjellick:
Our pleasure @thaJeztah. This cut covers the most basic use case. You can refer to the commit (https://github.com/cjellick/cattle/commit/ce8a7ab15aa57706b12a90908c4149707c47a137) to see exactly what functionality that coves.

We're actively working on the next round of functionality.

thaJeztah:
I know, one step at a time. Still very much appreciated!

PR #327, opened 2015-03-26
Error handling framework for ExternalHandlers

will-chan:
This PR introduced error handling for ExternalHandlers.  It will still require additional PR from go-machine-service and go-rancher to make it work end to end.

Requires:
https://github.com/rancherio/go-rancher/pull/7
https://github.com/rancherio/go-machine-service/pull/20

### DIFF
--- a/code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/dao/ExternalHandlerDao.java
+++ b/code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/dao/ExternalHandlerDao.java
@@ -1,11 +1,13 @@
 package io.cattle.platform.extension.dynamic.dao;
 
 import io.cattle.platform.core.model.ExternalHandler;
+import io.cattle.platform.extension.dynamic.data.ExternalHandlerData;
 
 import java.util.List;
 
 public interface ExternalHandlerDao {
 
-    List<? extends ExternalHandler> getExternalHandler(String eventName);
+    List<? extends ExternalHandler> getExternalHandler(String processName);
+    List<? extends ExternalHandlerData> getExternalHandlerData(String processName);
 
 }
--- a/code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/impl/ExternalDynamicExtensionHandlerImpl.java
+++ b/code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/impl/ExternalDynamicExtensionHandlerImpl.java
@@ -1,11 +1,11 @@
 package io.cattle.platform.extension.dynamic.impl;
 
 import io.cattle.platform.core.constants.ExternalHandlerConstants;
-import io.cattle.platform.core.model.ExternalHandler;
 import io.cattle.platform.engine.handler.ProcessLogic;
 import io.cattle.platform.eventing.EventService;
 import io.cattle.platform.extension.dynamic.DynamicExtensionHandler;
 import io.cattle.platform.extension.dynamic.dao.ExternalHandlerDao;
+import io.cattle.platform.extension.dynamic.data.ExternalHandlerData;
 import io.cattle.platform.object.ObjectManager;
 import io.cattle.platform.object.meta.ObjectMetaDataManager;
 import io.cattle.platform.object.process.ObjectProcessManager;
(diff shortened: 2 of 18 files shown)


resources/content/db/changelog.xml
```diff
@@ -33,5 +33,6 @@
     <include file="db/core-026.xml"/>
     <include file="db/core-027.xml"/>
     <include file="db/core-028.xml"/>
+    <include file="db/errorhandling.xml"/>
```
will-chan: I'll make the correct change when we are ready to merge.  I didn't want to deal with merge-conflicts later in case someone else is going to check in db changes.

cjellick:
You should add tests to test_external_handler.py for this.

https://github.com/rancherio/cattle/blob/e8f6206a13a4a8590ec0d9e2a71ac17ade896b04/tests/integration/cattletest/core/test_external_handler.py

code/iaas/model/src/main/java/io/cattle/platform/core/constants/ExternalHandlerConstants.java
```diff
@@ -6,5 +6,6 @@
     public static final String FIELD_TIMEOUT = "timeoutMillis";
     public static final String FIELD_PROCESS_NAMES = "processNames";
```
cjellick: No longer used, right?

will-chan:
Yep, working on that now.

code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/dao/ExternalHandlerDao.java
```diff
@@ -1,11 +1,13 @@
...
 
 import io.cattle.platform.core.model.ExternalHandler;
+import io.cattle.platform.extension.dynamic.data.ExternalHandlerData;
 
 import java.util.List;
 
 public interface ExternalHandlerDao {
 
-    List<? extends ExternalHandler> getExternalHandler(String eventName);
+    List<? extends ExternalHandler> getExternalHandler(String processName);
```
cjellick: Old method not used any more right? I think you should delete it.

code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/process/ExternalHandlerActivate.java
```diff
@@ -25,42 +29,47 @@
...
 
-        List<?> list = accessor.as(List.class);
+        List<? extends ExternalHandlerProcessConfig> list = accessor.asList(jsonMapper, ExternalHandlerProcessConfig.class);
         if (list != null) {
-            for (Object obj : accessor.as(List.class)) {
-                for (String part : obj.toString().trim().split("\\s*,\\s*")) {
-                    eventNames.add(part);
+            for (ExternalHandlerProcessConfig config : list) {
+                String name = config.getName();
+                for (String part : name.toString().trim().split("\\s*,\\s*")) {
```
cjellick: Looking at the PR as a whole now and how you've replaced processNames with processNameConfigs, I think you should remove this comma separation bit.

Since a process configuration is a json map and not a simple string, the split logic wont make sense.

cjellick:
There is sample bash code that will no long work because of this change:
https://github.com/rancherio/cattle/blob/master/docs/examples/handler-bash/simple_handler.sh

You don't need to fix that script now, but you should put a note in there saying that the example no longer applies and then maybe open a separate issue to update the example.

resources/content/schema/base/externalHandler.json
```diff
@@ -14,6 +14,11 @@
             "required" : true,
             "includeInList" : false
         },
+        "externalHandlerProcessConfigs" : {
```
cjellick: the name is a little redundant. should just be processConfigs

cjellick: Im just talking about the property on externalHandler. The name of the type is fine and can stay externalHandlerProcessConfigs.

code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/handler/EventBasedProcessHandler.java
```diff
@@ -100,9 +102,17 @@ public void progress(Event event) {
...
             }
         });
-        Event response = eventService.callSync(request, options);
 
-        return postEvent(state, process, CollectionUtils.toMap(response.getData()));
+        try {
+            Event response = eventService.callSync(request, options);
+            return postEvent(state, process, CollectionUtils.toMap(response.getData()));
+        } catch (ExecutionException e) {
+            if (getOnError() != null && getOnError().length() > 0) {
```
cjellick: fyi `StringUtils.isEmpty()` will handle this check as one call for you.

code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/dao/ExternalHandlerDao.java
will-chan: Not used but is the default method to retrieve an ExternalHandler.  I don't think there's any harm to keep it there for now in case we can make use of it later.

code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/process/ExternalHandlerActivate.java
will-chan: it's actually possible to do this still with a comma-delimited name.  You can also register the same set of events sharing the same onerror method.

code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/impl/ExternalDynamicExtensionHandlerImpl.java
```diff
@@ -53,37 +54,37 @@
...
 
-        processHandler.setEventName(String.format("%s;handler=%s", eventName, handler.getName()));
-        processHandler.setName(handler.getName());
-        processHandler.setPriority(priority);
-        processHandler.setRetry(retries);
-        processHandler.setTimeoutMillis(timeout);
+        processHandler.setEventName(String.format("%s;handler=%s", eventName, handlerName));
+        processHandler.setName(handlerName);
+        processHandler.setPriority(handler.getPriority());
+        processHandler.setRetry(handler.getRetries());
```
cjellick: How are retries and timeoutMillis getting populated if they came from the data field?

cjellick: Nevermind github had an outdated diff.

cjellick:
Last thing I've found: there should be at least one test that sets `onError` in the processConfigs of an ExternalHandler and then assert that it shows up in the `ExternalHandlerExternalHandlerProcessMap`.

cjellick:
That test can be as simple as a three line change to `test_external_handler(admin_client)`:
```
 def test_external_handler(admin_client):
     name = '{}-{}'.format(TEST_HANDLER_PREFIX, random_str())
-    configs = [{'name': 'instance.start'}]
+    configs = [{'name': 'instance.start', 'onError': 'instance.stop'}]
     h = admin_client.create_external_handler(name=name,
                                              processConfigs=configs)

@@ -50,6 +50,8 @@ def test_external_handler(admin_client):
     maps = h.externalHandlerExternalHandlerProcessMaps()
     assert len(maps) == 1
     assert maps[0].state == 'active'
+    ehpMap = maps[0]
+    assert ehpMap.onError == 'instance.stop'
```

code/iaas/external-handler/src/main/java/io/cattle/platform/extension/dynamic/process/ExternalHandlerActivate.java
cjellick: While it is still possible, it makes less sense now. It was a convenient shortcut for http-form based submissions before, but having the json type embedded makes that solution impractical.

At any rate, I only mentioned it because you had asked about it previously. I'm ok with leaving it in for the time being.

will-chan:
Yeah, I was going to add a lot more since we don't have any checks on subscribing to events, but I can certainly add this bit for now.

sonchang:
I looked over the java + config side (didn't look at the python test).  Looks good to me.

PR #329, opened 2015-03-27
Fixes rancherio/rancher #341

hibooboo2:


alena1108:
LGTM @cjellick @vincent99 can any of you guys review it, so we have 2 LGTMs on this PR

vincent99:
:+1:

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/GithubOAuthImpl.java
```diff
@@ -41,7 +42,7 @@ public Account getAccount(ApiRequest request) {
             return null;
         }
         String token = request.getServletContext().getRequest().getHeader(AUTH_HEADER);
-        if (StringUtils.isEmpty(token)) {
+        if (StringUtils.isEmpty(token) || !token.startsWith(AUTH_TYPE)) {
```
vincent99: Pedantic: This should be a case-insensitive test, and check for/split on whitespace after the AUTH_TYPE.
`Authorization: bearer jwtdata` should match
`Authorization: bearerfoobar somethingelse` shouldn't

alena1108: And one more pedantic comment from me - if the token is not of an "Bearer" (Github) type, we should return null right away w/o retrieving the token from the request parameters. As this class is only responsible for account lookups for github auth requests.

String token = request.getServletContext().getRequest().getHeader(AUTH_HEADER);
if (token != null && !token.startsWith(AUTH_TYPE)) {
return null
} else if (token == null) {
token = request.getServletContext().getRequest().getParameter("token");
......
}

vincent99: @alena1108 This class should return an account if **either** there is a JWT in the `token` param OR in a  `Authorization: Bearer jwtdata` header.  So return immediately is no good.

The problem @hibooboo2 is fixing is that in certain cases the browser running the UI generates requests that contains both a `Authorization: Basic adsfasdf` and `?token=jwtdata` query param.  We want to take the query param and ignore the `Authorization` header in this class, because it is not a Bearer token, but it was not checking that part before so it never looked at the `token` query param.

alena1108: @vincent99 ok, I was confused exactly by the use case when "requests contain both a Authorization: Basic adsfasdf and ?token=jwtdata". I thought in this case the Authorization header should take precedence, and handling of such request should be delegated to BasicAuthImpl AccountLookup:

vincent99: If there are basic creds and no token query param it will end up in basicauthimpl.

The reason it happens is if you use API keys in the API UI the browser remembers the basic auth creds and sends it with other requests.  The UI usually overrides this with a bearer jwt, but /v1/subscribe is a websocket connection where I can't send arbitrary http headers.  So we put the token in the query instead, and basic gets sent too, and the net result is that no resource.change events get sent to update the UI as things change.

PR #331, opened 2015-03-31
Service Discovery: LB service

alena1108:
This PR adds support for the Load Balancer Service. There are also some additional refactoring changes (to explain those 55 changed files):

1) moved all load balancer logic related code to its own package cattle-iaas-lb-logic. This package is used when LB is configured via API as a standalone feature; and by service-discovery component when LB is created as a part of LB service

2) Moved service->instance map creation to its separate handler invoked on the instance.start. Had to do it as service instance start process gets scheduled differently for lb and a regular service

Load Balancer service is a part of Service Discovery feature. 

LB service create triggers:

* creating a Load Balancer with listener ports defined in launchConfig parameter. Other parameters from launch Config are ignored by now

LB service activate: 

* register <scale> number of hosts for the load balancer. Today we just pick <scale> number of hosts owned by the user. The pick is random
* Containers of services' consuming LB service, get registered as LB targets on the lbs.

LB service deactivate:

* hosts get removed from the load balancer, and lb instances get destroyed.

All regular services operations like scaleup/scaledown are also supported by the LB service.


alena1108:
@sonchang @cjellick Could you please review the PR?

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/process/LoadBalancerRemovePostListener.java
```diff
@@ -0,0 +1,72 @@
...
+
+        // remove the config
+        objectProcessManager.scheduleProcessInstance(LoadBalancerConstants.PROCESS_LB_CONFIG_REMOVE, config,
+                state.getData());
+        return null;
+    }
+
+    @Override
+    public int getPriority() {
+        // TODO Auto-generated method stub
```
cjellick: Remove TODO

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceCreate.java
```diff
@@ -0,0 +1,183 @@
...
+            }
+            
+            Integer publicPort = spec.getPublicPort();
+            int privatePort = spec.getPrivatePort();
+            if (publicPort == null) {
+                publicPort = privatePort;
+            }
+
+            LoadBalancerListener listenerObj = objectManager.findOne(LoadBalancerListener.class,
+                    LOAD_BALANCER_LISTENER.NAME, sdService.getLoadBalancerName(service) + "_" + publicPort,
```
cjellick: Are the names of environments, load balancers, and listeners all not-editable?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceDeactivate.java
```diff
@@ -22,15 +16,6 @@
...
-    ObjectManager objectManager;
-    
-    @Inject
-    ResourceMonitor resourceMonitor;
-
-    @Inject
-    ProcessProgress progress;
-
-    @Inject
     ServiceDiscoveryService sdServer;
```
cjellick: Is `sdServer` the right name for this? Did you mean `sdService`?
...Won't stop me from LGTMing, just wanted to point it out.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/dao/ServiceConsumeMapDao.java
```diff
@@ -12,4 +12,6 @@
     List<? extends ServiceConsumeMap> findConsumedServices(long serviceId);
 
     List<? extends ServiceConsumeMap> findMapsToRemove(long serviceId);
+    
+    List<? extends ServiceConsumeMap> findConsumedByServices(long serviceId);
```
cjellick: Is `findConsumingServices` a more appropriate name for this method?
Also, whether you change this name or not, this method and `findConsumedServices` are so similarly named, I think they warrant have a javadoc clarifying their intention. A couple one-liners like:
> Finds all the services that consume the specified service.
and
> Finds all the services that the specified service consumes.
would suffice.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceCreate.java
alena1108: Craig, initially I've made them not editable. But then Vince and Darren suggested I allow to edit them. So you are right, for lb service I should no longer rely on the name of the listener/config to be consistent over time. I will add serviceId to indicate they are the part of the service (like I already do for the load balancer itself)

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/dao/ServiceConsumeMapDao.java
alena1108: Sure, Craig, will change the name as well as add the java doc.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceDeactivate.java
alena1108: will change, not a big deal :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -328,9 +373,233 @@ public long getServiceNetworkId(Service service) {
...
+        }
+        startServiceInstances(instancesToStart);
+    }
+
+    private void startServiceInstances(List<Instance> instancesToStart) {
+        for (Instance instance : instancesToStart) {
+            scheduleServiceInstanceStart(instance);
+        }
+        for (Instance instance : instancesToStart) {
+            progress.checkPoint("start service instance " + instance.getUuid());
```
cjellick: Would it be better to use instance.name for the progress message?

alena1108: sure

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -328,9 +373,233 @@ public long getServiceNetworkId(Service service) {
...
+    }
+
+    private void stopServiceInstances(List<Instance> instances) {
+        for (Instance instance : instances) {
+            objectProcessManager.scheduleProcessInstance(InstanceConstants.PROCESS_STOP, instance, null);
+        }
+    }
+
+    @Override
+    public void scaleDownService(Service service, int requestedScale) {
```
cjellick: This method seems brittle in that if the instance names ever get out of sync with the count, the scale down won't happen.

Is that the intended behavior?

alena1108: @cjellick this will get fixed as a part of the separate PR. The original intention was to rely on the immutable service/environment names to be used for instances names. Then just today Vince/Darren suggested we let user to modify it. In this PR I've just moved this method from one class to another - it shouldn't affect regular service's behavior.

alena1108: @cjellick I gave it a second thought - it might be good to address this problem in this PR as the instance->service map creation affects both LB and regular service

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/api/action/LoadBalancerAddHostActionHandler.java
```diff
@@ -38,17 +31,8 @@ public Object perform(String name, Object obj, ApiRequest request) {
         LoadBalancer lb = (LoadBalancer) obj;
         long hostId = DataAccessor.fromMap(request.getRequestObject()).withKey(LoadBalancerConstants.FIELD_LB_HOST_ID)
                 .as(Long.class);
-        createLbHostMap(lb.getId(), hostId);
+        lbService.addHostToLoadBalancer(lb, hostId);
 
         return objectManager.reload(lb);
```
sonchang: Since addHostToLoadBalancer just creates the mapping entry, does it really change the lb instance?  I'm wondering whether the reload is necessary or not?

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/api/action/LoadBalancerConfigSetListenersActionHandler.java
```diff
@@ -83,9 +74,7 @@ private void removeOldListenerMaps(LoadBalancerConfig config, List<? extends Lon
         }
 
         for (LoadBalancerConfigListenerMap mapToRemove : mapsToRemove) {
-            objectProcessManager.scheduleProcessInstance(LoadBalancerConstants.PROCESS_LB_CONFIG_LISTENER_MAP_REMOVE,
-                    mapToRemove, null);
+            lbService.removeListenerFromConfig(config, mapToRemove.getLoadBalancerListenerId());
```
sonchang: Probably not a big deal, but with the new changes, we now do a DB lookup for each map since we're just passing the ID here.  I didn't check but I'm guessing nothing is locked so I don't know if we get much out of the findNonRemoved call in lbService.removeListenerFromConfig()

alena1108: We do if the method is called from another context, when just the listenerId is know (LoadBalancerConfigRemoveListenerActionHandler).

cjellick:
LGTM

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/api/action/LoadBalancerAddHostActionHandler.java
alena1108: Mapping creation triggers the LB instance creation/start. This logic is in LoadBalancerUpdateConfig handler.

sonchang: So is it the state field for the LB that changes or some other field?  Would the trigger be that quick since the create process for the mapping is scheduled rather than executed immediately.

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/service/LoadBalancerService.java
```diff
@@ -0,0 +1,24 @@
...
+
+    void removeHostFromLoadBalancer(LoadBalancer lb, long hostId);
+
+    void addListenerToConfig(LoadBalancerConfig config, long listenerId);
+
+    void removeListenerFromConfig(LoadBalancerConfig config, long listenerId);
+
+    void addTargetToLoadBalancer(LoadBalancer lb, long instanceId);
+
+    void addTargetToLoadBalancer(LoadBalancer lb, String ipAddress);
```
sonchang: Might want to rename to 'addTargetIpToLoadBalancer'.  I think it's a little risky to have the same name even though the method signature's different.  Plus, it'll match the removeTargetIpFromLoadBalancer() below

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/api/action/LoadBalancerAddHostActionHandler.java
alena1108: When host is added to the load balancer, load balancer state doesn't change. And we never use the lb state to signal to start lb instance + update haproxy content. Instead the latter actions get triggered by mapping creation - host add/remove, target add/remove. So LB should always be open for modifications, and all of them are scheduled in async manner.

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/service/LoadBalancerService.java
alena1108: Yup, very valid suggestion, will do

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/api/action/LoadBalancerAddHostActionHandler.java
sonchang: Thanks.  Just for my understanding, at this point, none of the fields in LB changes, but we're reloading just to be sure since the chain of events could potentially change something.

alena1108: Yup, like the latest transition message as an example.

sonchang: Thanks!  makes sense

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/process/LoadBalancerRemovePostListener.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { LoadBalancerConstants.PROCESS_LB_REMOVE };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        LoadBalancer lb = (LoadBalancer) state.getResource();
+        if (lb.getServiceId() == null) {
```
sonchang: Should this be != null?

alena1108: No. This handler takes care only of Service's load balancer. I should probably change the name of the handler to reflect that.

alena1108: And move it to the service discovery mvn package

sonchang: Thanks.  Yeah.  I was a little confused also because of the comment below.  So I thought it was the opposite.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceCreate.java
```diff
@@ -0,0 +1,185 @@
...
+
+        for (String port : portDefs) {
+            PortSpec spec = new PortSpec(port);
+            if (!port.contains("tcp")) {
+                // default to http unless defined otherwise in the compose file
+                spec.setProtocol("http");
+            }
+
+            if (listeners.containsKey(spec.getPrivatePort())) {
+                continue;
```
sonchang: I'm guessing this shouldn't happen in the normal case right?  So if it does happen, maybe we should log it prior to skipping?

alena1108: :) it can happen with idempotent checks in. As with this check on, the loop will never gonna have sequential successes for all the ports defined in the portDefs

sonchang: Ahh. Dang idempotent checks!!!

tests/integration/cattletest/core/test_svc_discovery_lb.py
```diff
@@ -0,0 +1,525 @@
...
+                                   environmentId=env.id,
+                                   launchConfig=launch_config)
+    service = client.wait_success(service)
+    assert service.state == "inactive"
+
+    # verify that the load balancer was created for the service
+    lb = client. \
+        list_loadBalancer(serviceId=service.id)
+    assert len(lb) == 1
+    assert lb[0].state == 'active'
```
sonchang: If the service is inactive, should the LB be active?

alena1108: Lb gets activated by default during its Create process.

sonchang:
LGTM

PR #334, opened 2015-04-01
Added new DO fields and made "name" a required parameter for machine.

will-chan:
Also made vpcId optional given that it's either vpcId or subnetId that
is required.

### DIFF
--- a/code/implementation/docker/machine/src/main/resources/schema/base/amazonec2Config.json.d/amazonec2.json
+++ b/code/implementation/docker/machine/src/main/resources/schema/base/amazonec2Config.json.d/amazonec2.json
@@ -12,8 +12,7 @@
         },
         "vpcId": {
             "type": "string",
-            "nullable": false,
-            "required": true
+            "nullable": true
         },
         "sessionToken": {
             "type": "string",
--- a/code/implementation/docker/machine/src/main/resources/schema/base/digitaloceanConfig.json.d/digitalocean.json
+++ b/code/implementation/docker/machine/src/main/resources/schema/base/digitaloceanConfig.json.d/digitalocean.json
@@ -16,6 +16,18 @@
             "type": "string",
             "nullable": false,
             "required": true
+        },
+        "ipv6": {
+            "type": "boolean",
+            "nullable": true
+        },
+        "privateNetworking": {
+            "type": "boolean",
+            "nullable": true
+        },
+        "backups": {
+            "type": "boolean",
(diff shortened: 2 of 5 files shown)


cjellick:
Assuming tests pass, LGTM

code/implementation/docker/machine/src/main/resources/schema/base/amazonec2Config.json.d/amazonec2.json
```diff
@@ -12,8 +12,7 @@
         },
         "vpcId": {
             "type": "string",
```
alena1108: We have to add an API filter to check if either vpcId or subnetId is specified, but not both

vincent99: Setting both seems to be valid in the driver (if not terribly useful, since subnet implies vpc).  There should be validation on our side that one or the other is set though if possible instead of going off and failing in machine.

cjellick: We've been taking the approach of only requiring the fields minimally required by docker-machine, Everything else, we allow machine to validate and then we propagate the error back up to the user.

Adding what you're suggeting isnt a bad idea though...especially as the drivers mature and stabalize.

Would you be comfortable with doing that in a separate PR further down the road?

vincent99: That's fine, I can add a check for that in the UI too.

alena1108: LGTM if we decide to move the validation part to the UI, or decide that we don't need it at all. @will-chan ?

will-chan: yeah, I think we should add the check in the UI but for now, since it's so specific to amazon, we'll let docker-machine handle it.  I don't want to get a place where docker-machine changes it's requirement on us and we have to change cattle just to accommodate it.

This does mean we might need to think about how far we want to consider supporting these "first class" drivers.  There's so much more we can do here before sending the values down to machine.

alena1108:
LGTM

cjellick:
Green build and 2 LGTMs. merging

PR #335, opened 2015-04-03
LB - change configUpdate processes to actionHandlers

alena1108:


alena1108:
@cjellick @sonchang Son, Craig, can you please review this PR. Its mostly refactoring work done for the Load Balancer, required by UI. The goal was - move all the processes responsible for updating haproxy config to ActionHandlers, so the API calls can be made on the object at any time of its lifecycle. As a result, the UI doesn't have to check the object's state before performing the call.

Those processes converted to handlers, are:

1) for loadBalancer
add/remove/setHost(s)
add/remove/setTarget(s)

2) for loadBalancerConfig
add/remove/setListener(s)

I had to change a bunch of tests because validation whether the operation was successful, no longer depend on the object's state the action is called for, but on the objects that were created as a result of action.

I also reverted one of my previous commits that were merged to master, and this commit makes the merged one obsolete. So I think we shouldn't squash the merge revert commit in this PR as seeing it in the history might be important. What do you think, @cjellick?

@vincent99 Vince, as LB/LBConfig state no longer depends on the state of host/targets/listeners bound to it, we have to display the state of host-lb, target-lb, config-listener maps to indicate what had been set, and what not (by the map state Active/Activating/Removing/Removed). Process failure will also be indicated on per map basis. We have to think on how to display/aggregate this info on the LoadBalancer/Config object level.

cjellick:
Yes, I agree, lets not squash.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddHostActionHandler.java
```diff
@@ -0,0 +1,54 @@
...
+        return lb;
+    }
+
+    protected void createLbHostMap(long lbId, long hostId) {
+        LoadBalancerHostMap lbHostMap = mapDao.findNonRemoved(LoadBalancerHostMap.class, LoadBalancer.class, lbId, Host.class, hostId);
+
+        if (lbHostMap == null) {
+            lbHostMap = objectManager.create(LoadBalancerHostMap.class, LOAD_BALANCER_HOST_MAP.LOAD_BALANCER_ID, lbId, LOAD_BALANCER_HOST_MAP.HOST_ID, hostId);
+        }
+        objectProcessManager.scheduleProcessInstance(LoadBalancerConstants.PROCESS_LB_HOST_MAP_CREATE, lbHostMap, null);
```
sonchang: I'm wondering if lbHostMap already exist, what state would it be in.  Just wondering whether the scheduleProcessInstance(...) call afterwards might error out since it might not be in a valid state for 'create'.
In ClusterAddHostActionHandler, I ended up throwing a ValidationErrorException when I saw the mapping already existed.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddTargetActionHandler.java
```diff
@@ -0,0 +1,68 @@
...
+
+        return lb;
+    }
+
+    private LoadBalancerTarget getLbTarget(LoadBalancer lb, Long instanceId, String ipAddress) {
+        LoadBalancerTarget target = null;
+        if (ipAddress != null) {
+            target = lbTargetDao.getLbIpAddressTarget(lb.getId(), ipAddress);
+        } else {
+            target = lbTargetDao.getLbInstanceTarget(lb.getId(), instanceId);
```
sonchang: quick question.  is this a valid case where ipAddress is null so we have to look it up by instanceId?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddHostActionHandler.java
alena1108: @sonchang I have an API filter checking for existing maps, and throw exception here. Once we get to the handlers, it means we pass the check. This not null check for create was added to support idempotent retries.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddTargetActionHandler.java
alena1108: @sonchang yes, its a valid use case. Thats why I put this if/else statement. IpAddress takes precedence; if its not passed, instanceId is considered.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddHostActionHandler.java
alena1108: @sonchang I also have a LoadBalancerTargetValidationFilter where I verify if either one or another passed in. But I am missing validation for "set" action there (exists only for add/remove). Will fix it.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddTargetActionHandler.java
alena1108: @sonchang I also check if either one or another passed in, in the API filter LoadBalancerTargetValidationFilter

tests/integration/cattletest/core/test_lb_config.py
```diff
@@ -234,44 +216,20 @@ def test_lb_config_set_listeners(admin_client, super_client):
     # set 2 listeners
     config = config.setlisteners(
         loadBalancerListenerIds=[listener1.id, listener2.id])
-    config = admin_client.wait_success(config)
     assert config.state == 'active'
```
sonchang: Is assert config.state == 'active' necessary?  I don't think setListeners(...) changes it right?

tests/integration/cattletest/core/test_lb_config.py
```diff
@@ -326,15 +284,8 @@ def test_lb_config_add_conflicting_listener(admin_client, super_client):
 
     # add listener to the config
     config = config.addlistener(loadBalancerListenerId=listener.id)
-    config = admin_client.wait_success(config)
-
     assert config.state == 'active'
```
sonchang: Doesn't look like we need to check state is 'active'

alena1108: you are right, I will remove it.

sonchang:
@alena1108 LGTM

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddTargetActionHandler.java
cjellick: @alena1108 Its not in this PR, but think LoadBalancerTargetValidationFilter should validate that a system container is not being added as a target? I was able to add the net agent.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
```diff
@@ -1,42 +1,50 @@
...
-        return new String[] { LoadBalancerConstants.PROCESS_LB_CONFIG_SET_LISTENERS };
+    public String getName() {
+        return LoadBalancerConstants.PROCESS_LB_CONFIG_SET_LISTENERS;
     }
 
     @Override
-    public HandlerResult handle(ProcessState state, ProcessInstance process) {
-        LoadBalancerConfig config = (LoadBalancerConfig) state.getResource();
-        List<? extends Long> newListenerIds = DataAccessor.fromMap(state.getData()).withKey(LoadBalancerConstants.FIELD_LB_LISTENER_IDS).asList(jsonMapper,
+    public Object perform(String name, Object obj, ApiRequest request) {
```
cjellick: For my clarification, setlisteners is just for adding listeners? its not supposed to basically clear out all current listeners and replace with the provided ones

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
```diff
@@ -46,7 +54,7 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
             // create a new set
             createNewListenerMaps(config, newListenerIds);
         }
-        return null;
+        return objectManager.reload(config);
```
cjellick: This is the only action handler thus far where you've reload the resource before returning. Do you think you should consistently reload it for all the handlers?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
cjellick: Bad question. Clearly, the intended behavior is to remove the listeners. When I ran it, they didn't go away. Let me make sure I did it correctly.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
sonchang: Actually, I was going to comment that I don't think it's necessary to reload.  But figured it doesn't do too much harm (at least until we need to worry about performance testing)

cjellick: Right, it doesn't really seem to need a reload. None of the state on the base resource is changing as a result of these actions.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerAddTargetActionHandler.java
alena1108: @cjellick  You can add whatever you want :) network container, specify an external ip - as long as those instances are visible to you. We are not putting restrictions on that.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
cjellick: I dont know what happened, removing a listener was resulting in a port.create process to kick off and that was failing. But I cannot repro.

alena1108: Its exactly for rewriting the set. To add/remove listener, we use add/removeListener API. In the future we will refactor those to accept the list of ids. Today it accepts a single id.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
alena1108: @sonchang  @cjellick  and you are both right! and I need to be more consistent for sure :)

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/lb/LoadBalancerConfigSetListenersActionHandler.java
alena1108: @cjellick yes, adding the listener does result in port creation + host-iptables config item update. If you see that failure again, can you send me the log?

cjellick:
LGTM

PR #336, opened 2015-04-03
HAProxy: use listener uuid for backend/frontend

alena1108:
As listener.name is not required, and can be null

@vincent99 can you pls review


vincent99:
LGTM

will-chan:
LGTM

PR #337, opened 2015-04-03
ServiceDiscovery:deactivate/remove commands are async now

alena1108:
Schedule deactivate/remove processes for corresponding
instances/services and return the result. Used to wait for removal
completion before

### DIFF
--- a/tests/integration/cattletest/core/test_svc_discovery.py
+++ b/tests/integration/cattletest/core/test_svc_discovery.py
@@ -184,7 +184,28 @@ def test_activate_services(super_client, admin_client, sim_context, nsp):
     assert service2.state == "active"
 
 
-def test_deactivate_service(super_client, admin_client, sim_context, nsp):
+def _validate_instance_stopped(admin_client, service, super_client):
+    instances = super_client. \
+        list_container(name="compose_" + service.name + "_" + "1")
+    assert len(instances) == 1
+    instance = instances[0]
+    wait_for_condition(
+        admin_client, instance, _resource_is_stopped,
+        lambda x: 'State is: ' + x.state)
+
+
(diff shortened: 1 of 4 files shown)


alena1108:
@sonchang @cjellick can you please review this PR? I would need it merged in order to start working on service.scaleDown functionality

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -295,21 +311,13 @@ def test_remove_inactive_service(super_client, admin_client, sim_context, nsp):
...
 
     # remove service
-    service = wait_success(admin_client, service.remove(), 120)
+    service = wait_success(admin_client, service.remove())
     assert service.state == "removed"
-    instances = super_client. \
-        list_container(name="compose_" + service.name + "_" + "1")
-    assert len(instances) == 1
-    assert instances[0].state == "removed"
-    instance_service_map = super_client. \
```
sonchang: Not a merge blocker, but was there any particular reason we were checking the state of the mapping before?

sonchang:
LGTM

tests/integration/cattletest/core/test_svc_discovery.py
alena1108: I can re-add this check later. I removed it because making sure the instance is removed, is sufficient enough; and with the async changes I had to add more logic for making sure that the mapping is gone. wait_for_condition can't handle map objects because they don't support "reload" method

sonchang: actually, i don't think it's even necessary to add it back in.  but was
wondering whether there was any particular reason in the first place

cjellick:
LGTM

PR #339, opened 2015-04-06
Added accountId to LB map tables

alena1108:
@sonchang @vincent99 can you please review? The checkin adds missing accountId to LB related maps table - as the UI needs to subscribe to the resource.change events on map records, and it doesn't work when accountId is not set on the map object.


sonchang:
LGTM

alena1108:
@sonchang @vincent99 Just figured out that putting accountId to the mapping tables has some impact on the way resources map to each other. Apparently when more than 2 foreign keys are present in *map table (and it will be 3 with added account_id), the resources are no longer listed under each other "links" section, and fail to be discovered via jooq "mappedChildren" method.

So hold on with reviewing/merging, I will talk to Darren tomorrow about what fix should be put in.

alena1108:
Found the cause for this behavior: in DefaultObjectMetaDataManager when *map tables are being processed, method findMappings() skips mapping tables having more than 2 Foreign keys. So possible fixes would be:

1) Don't setup foreign key for the account_id field in the mapping table
2) Fix DefaultObjectMetaDataManager to accept more than 2 foreign keys for *map table (not sure what would be the impact of that)
3) Provide a diff solution for Event publishing for mapped object type and don't add account_id to map table. Extract the account_id from the object that is being referenced via foreign key?

@ibuildthecloud Darren, need your input on that - would do you think would be the correct fix?

sonchang:
I'm leaning towards 2 or 3.  I guess for #2, I recall we already do some
naming convention hackery but I don't think we want to continue doing
that.  I haven't looked at the code for #3 so I don't know the extent of
the code changes necessary, but I think I like this option the best.

alena1108:
@sonchang spoke with Darren - he says that acocunt_id should be there in map, otherwise nobody else but the admin will not be able to even see those mappings via API/UI. I will have to exclude the account_id from the list of foreign keys being processed by Jooq manager. It will be configurable.

sonchang:
LGTM.  I like the exclusion list idea

cjellick:
LGTM

PR #342, opened 2015-04-08
Scale up/down support for Service Discovery

alena1108:
WIP, scaleUP is supported, scaleDown is yet to be added

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
@@ -1,6 +1,5 @@
 package io.cattle.platform.servicediscovery.process;
 
-import static io.cattle.platform.core.model.tables.EnvironmentTable.ENVIRONMENT;
 import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
 import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.SERVICE_EXPOSE_MAP;
 import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
@@ -8,7 +7,6 @@ import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.dao.GenericMapDao;
 import io.cattle.platform.core.dao.GenericResourceDao;
-import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.core.model.Image;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.model.Service;
@@ -28,6 +26,7 @@ import io.cattle.platform.object.util.DataUtils;
 import io.cattle.platform.process.common.handler.AbstractObjectProcessHandler;
 import io.cattle.platform.process.progress.ProcessProgress;
 import io.cattle.platform.servicediscovery.api.constants.ServiceDiscoveryConstants;
+import io.cattle.platform.servicediscovery.api.dao.ServiceExposeMapDao;
 import io.cattle.platform.servicediscovery.resource.ServiceDiscoveryConfigItem;
 import io.cattle.platform.servicediscovery.service.ServiceDiscoveryService;
 import io.cattle.platform.storage.service.StorageService;
@@ -68,6 +67,9 @@ public class ServiceActivate extends AbstractObjectProcessHandler {
     @Inject
     ServiceDiscoveryService sdServer;
 
+    @Inject
+    ServiceExposeMapDao svcExposeDao;
+
     @Override
     public String[] getProcessNames() {
         return new String[] { ServiceDiscoveryConstants.PROCESS_SERVICE_ACTIVATE,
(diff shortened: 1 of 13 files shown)


alena1108:
@cjellick @sonchang can you pls review this PR? it provides a support for scaling service up and down, meaning increasing/reducing number of service instances.

cjellick:
What is the functional difference between an active service and an inactive one?

Asking in the context of Vince's standard lifecycle issue. https://github.com/rancherio/rancher/issues/384

alena1108:
@cjellick Active service has service's instances started. Inactive service's instances are stopped

https://github.com/rancherio/rancher/wiki/Service-Discovery-API-reference

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -68,15 +67,40 @@
...
+                ServiceDiscoveryConstants.PROCESS_SERVICE_UPDATE };
     }
 
     @Override
     @SuppressWarnings("unchecked")
     public HandlerResult handle(ProcessState state, ProcessInstance process) {
         Service service = (Service) state.getResource();
+
+        // on inactive service update, do nothing
+        if (process.getName().equalsIgnoreCase(ServiceDiscoveryConstants.PROCESS_SERVICE_UPDATE)
```
cjellick: Why allow inactive service update if it is a no-op? (This may just be me missing the big picture)

alena1108: @cjellick well, I just couldn't find a reason of why disallow it :) its just a simple object's property change when inactive. On the service activation, number of service's instances will be = scale

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -180,21 +203,39 @@ public Object call() throws Exception {
...
+        }
+
+        return image == null ? null : image.getId();
+    }
+
+
+    protected Instance createInstance(Service service, int i, Map<String, Object> launchConfigData, Long imageId,
+            List<Long> networkIds) {
+        String instanceName = sdServer.getInstanceName(service, i);
         Instance instance = objectManager.findOne(Instance.class, INSTANCE.NAME, instanceName,
```
cjellick: Does `ServiceExposeMapDao.getServiceInstance` do the same thing (or close enough to the same thing that it could/should be used here?

resources/content/schema/base/service.json
```diff
@@ -9,7 +9,10 @@
             "type": "int",
             "required": false,
             "nullable": true,
-            "default": 1
+            "default": 1,
+             "attributes" : {
+                "scheduleUpdate" : true
```
cjellick: For my own education, does these mean that any time this field is changed, an update process will be scheduled?

cjellick: (And this may explain why you have the inactive update)

alena1108: @cjellick yes. Without this attribute specified, no handler will be invoked on db resource change.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
alena1108: @cjellick not really. At this point there might be no map as there is no service instance was created yet. First we create an instance, then register it in instance->service (serviceExpose) map

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -789,6 +789,82 @@ def test_remove_environment_w_active_svcs(super_client,
...
+def test_valiate_service_scaleup_scaledown(super_client,
+                                           admin_client, sim_context, nsp):
+    env = admin_client.create_environment(name="compose")
+    env = admin_client.wait_success(env)
+    assert env.state == "active"
+
+    image_uuid = sim_context['imageUuid']
+    launch_config = {"imageUuid": image_uuid}
+
+    service = super_client.create_service(name=random_str(),
```
cjellick: Why do you need to use super_client for this?
My understanding is that we like to restrict the use of super_client solely to system level resources that a normal user (or admin) can access.

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -789,6 +789,82 @@ def test_remove_environment_w_active_svcs(super_client,
...
+    # scale down the service
+    service = admin_client.update(service, scale=2)
+    service = admin_client.wait_success(service, 120)
+    assert service.state == "active"
+    _validate_compose_instance_start(service, super_client, "1")
+    _validate_compose_instance_start(service, super_client, "2")
+    _validate_instance_removed(service, super_client, "3")
+    _validate_instance_removed(service, super_client, "4")
+
+
```
cjellick: Want a test to verify the no-op behavior of updating a deactivated service?
I don't think you _need_ it and its absence wont stop me from LGTMing, but just wanted to point it out.

tests/integration/cattletest/core/test_svc_discovery.py
alena1108: @cjellick Because there are certain parameters that I don't expose to an end user, and that are needed in order to run against simulator. Like networkId, agent URI predicate, etc

tests/integration/cattletest/core/test_svc_discovery.py
alena1108: @cjellick sure, will add the test

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -68,15 +67,40 @@
...
     ServiceDiscoveryService sdServer;
 
+    @Inject
+    ServiceExposeMapDao svcExposeDao;
+
     @Override
     public String[] getProcessNames() {
-        return new String[] { ServiceDiscoveryConstants.PROCESS_SERVICE_ACTIVATE };
+        return new String[] { ServiceDiscoveryConstants.PROCESS_SERVICE_ACTIVATE,
+                ServiceDiscoveryConstants.PROCESS_SERVICE_UPDATE };
```
sonchang: I guess I'm not able to comment on a non-modified line in github (since I wanted to comment on the class name above).
Since we're adding 'update', I'm wondering whether we should modify the class name too to avoid confusion.  Super minor, so not merge blocking.

sonchang:
LGTM

cjellick:
LGTM

PR #343, opened 2015-04-10
Addressed various LB bugs

alena1108:
1) handle instance.restart when update haproxy config
2) unbind public ports when there are no active lb targets
3) fixed the bug when duplicated ports were created for the same lb
instance

Also bug fix for instance.stop: don't try to deactivate purged/removed
ports

### DIFF
--- a/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
+++ b/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
@@ -8,14 +8,15 @@ import io.cattle.platform.configitem.version.ConfigItemStatusManager;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.constants.PortConstants;
+import io.cattle.platform.core.dao.GenericResourceDao;
 import io.cattle.platform.core.dao.IpAddressDao;
 import io.cattle.platform.core.dao.LoadBalancerDao;
+import io.cattle.platform.core.dao.LoadBalancerTargetDao;
 import io.cattle.platform.core.dao.NetworkDao;
 import io.cattle.platform.core.model.Agent;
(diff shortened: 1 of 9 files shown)


alena1108:
https://github.com/rancherio/rancher/issues/504
https://github.com/rancherio/rancher/issues/503
https://github.com/rancherio/rancher/issues/441
https://github.com/rancherio/rancher/issues/430

cjellick:
Are you able to add test cases for these bugs?

#504 seems like it could be a straight-forward integration test, buy the rest would require more complex validation test scenarios.

If you say it's not worth the effort, I'll agree and give the thumbs up, but just thought I should pose the question.

alena1108:
@cjellick #3 was reproducible only with idempotent checks off. I already had a test for this case, and now I always run it for checks on and off

As a part of this PR, the test is added for #2.
Unfortunately #1 can't be covered with integration test as it requires backend haproxy config verification.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
```diff
@@ -177,45 +190,69 @@ public boolean evaluate(Instance obj) {
...
+                            if (listenerPorts.contains(port.getPublicPort()) && hasActiveInstances) {
+                                portsToCreate.put(port.getPublicPort(), port);
+                            } else {
+                                portsToRemove.put(port.getPublicPort(), port);
+                            }
+                        }
+
+                        if (hasActiveInstances) {
+                            for (Integer listenerPort : listenerPorts) {
+                                if (existingPorts.containsKey(listenerPort)) {
```
sonchang: Just out of curiosity.  Previously we were checking portsToCreate.  In the new changes, we're checking existingPorts which is a superset of portsToCreate.

alena1108: You are right. That was an unneeded change. But shouldn't break anything

sonchang:
LGTM

cjellick:
LGTM

PR #344, opened 2015-04-11
Enable Read access for lb map object types

alena1108:
loadBalancerConfigListenerMap
loadBalancerHostMap

As we rely on the map states to reflect haproxy config changes, we should expose those objects to an end user.

@sonchang it will probably be similar to your host->cluster map

@sonchang @cjellick can you please review?


sonchang:
LGTM.  I'm still wondering whether there's a better way to expose progress without exposing the map resource itself.

cjellick:
LGTM

PR #349, opened 2015-04-15
Docker inspect transformer

cjellick:
Class to transform a docker inspect to a rancher container instance.


code/implementation/docker/common/src/main/java/io/cattle/platform/docker/constants/DockerInstanceConstants.java
```diff
@@ -5,7 +5,29 @@
...
+    public static final String FIELD_DOMAIN_NAME = "domainName";
+    public static final String FIELD_USER = "user";
+    public static final String FIELD_MEMORY = "memory";
+    public static final String FIELD_MEMORY_SWAP = "memorySwap";
+    public static final String FIELD_CPU_SHARES = "cpuShares";
+    public static final String FIELD_CPU_SET = "cpuSet";
+    public static final String FIELD_TTY = "tty";
+    public static final String FIELD_STDIN_OPEN = "stdinOpen";
+    public static final String FIELD_DIRECTORY = "directory";
+    public static final String FIELD_COMMAND = "command";
```
alena1108: Craig, I know we also have command_arg parameter in Rancher, along with command. Do these 2 rancher parameters get translated to a single one - command - on the docker side?

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instancehostmap/DockerPostInstanceHostMapActivate.java
```diff
@@ -99,21 +108,31 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
     }
 
-    @SuppressWarnings({ "unchecked", "rawtypes" })
+    @SuppressWarnings({ "unchecked" })
+    protected void nativeDockerBackPopulate(Instance instance) {
+        Map<String, Object> inspect = (Map<String, Object>)instance.getData().get(FIELD_DOCKER_INSPECT);
+        if (inspect == null || instance.getNativeContainer() == null || !instance.getNativeContainer().booleanValue()) {
+            return;
+        }
+        transformer.transform(inspect, instance);
```
alena1108: Craig, not sure if it should be a point of concern, but I see that we update instance and then update the same instance with the volumes as a part of 2 separate calls. Would it cause any problem if one of the transactions fail? And if yes, we should probably wrap up these 2 updates within a single transaction

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/constants/DockerInstanceConstants.java
cjellick: We _just_ eliminated that: we have a single command parameter that is an array.
https://github.com/rancherio/cattle/pull/370

alena1108: Nice, didn't realize that PR was to fix this problem. Thanks!

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instancehostmap/DockerPostInstanceHostMapActivate.java
cjellick: @alena1108 That's a good point to bring up, but I don't think these need to be wrapped up in a single transaction. Backfilling the container fields (or failing to) doesn't impact creating volumes and vice versa. I got @ibuildthecloud's opinion on it as well, to make sure I wasn't missing the mark.

alena1108: Sure, as long as it doesn't impact anything. Thanks for the confirmation!

alena1108:
LGTM

ibuildthecloud:
LGTM

PR #355, opened 2015-04-17
serviceDiscovery/lb bug fixes

alena1108:
1) Convert restarPolicy to format docker-compose understands: 

https://github.com/rancherio/rancher/issues/548

2) Allow only [a-zA-Z0-9] for service/environment name as those are the chars that can be used for Docker containers' names

3) fixed the typo in load balancer resource name.

4) Fixed private registry support - used to not work when credentials were required

https://github.com/rancherio/rancher/issues/572

5) Expose cpuSet and requestedHostId in service

https://github.com/rancherio/rancher/issues/576
https://github.com/rancherio/rancher/issues/549

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
@@ -1,5 +1,6 @@
 package io.cattle.platform.servicediscovery.process;
 
+import static io.cattle.platform.core.model.tables.ImageTable.IMAGE;
 import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
 import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.SERVICE_EXPOSE_MAP;
 import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
@@ -166,7 +167,9 @@ public class ServiceActivate extends AbstractObjectProcessHandler {
     private void createServiceInstances(Service service, int scale) {
         List<Instance> instancesToStart = new ArrayList<>();
         Map<String, Object> launchConfigData = sdServer.buildLaunchData(service);
-        Long imageId = getImage(String.valueOf(launchConfigData.get(InstanceConstants.FIELD_IMAGE_UUID)));
+        Object registryCredentialId = launchConfigData.get(ServiceDiscoveryConfigItem.REGISTRYCREDENTIALID.getRancherName());
+        Long imageId = getImage(String.valueOf(launchConfigData.get(InstanceConstants.FIELD_IMAGE_UUID)),
+                registryCredentialId != null ? (Integer) registryCredentialId : null);
         List<Long> networkIds = getServiceNetworks(service);
         for (int i = 0; i < scale; i++) {
             Instance instance = createInstance(service, i, launchConfigData, imageId, networkIds);
@@ -211,10 +214,16 @@ public class ServiceActivate extends AbstractObjectProcessHandler {
         return ntwkIds;
     }
 
-    protected Long getImage(String imageUuid) {
+    protected Long getImage(String imageUuid, Integer registryCredentialId) {
         Image image;
         try {
             image = storageService.registerRemoteImage(imageUuid);
+            if (image == null) {
+                return null;
+            }
+            if (registryCredentialId != null) {
+                objectManager.setFields(image, IMAGE.REGISTRY_CREDENTIAL_ID, registryCredentialId);
+            }
         } catch (IOException e) {
             throw new IllegalStateException("Failed to get image [" + imageUuid + "]");
         }
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/RancherConfigToComposeFormatter.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/RancherConfigToComposeFormatter.java
@@ -1,0 +1,7 @@
+package io.cattle.platform.servicediscovery.service.impl;
+
+import io.cattle.platform.servicediscovery.resource.ServiceDiscoveryConfigItem;
+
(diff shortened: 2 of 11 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceActivate.java
```diff
@@ -211,10 +213,16 @@ public Object call() throws Exception {
         return ntwkIds;
     }
 
-    protected Long getImage(String imageUuid) {
+    protected Long getImage(String imageUuid, Object registryCredentialId) {
```
sonchang: Quick question.  Is the registryCredentialId a Long or could it also be something else?

alena1108: Its supposed to be Long, but I put object to simply avoid casting (and not null checks) when fetching this field from launchConfig. As objectManager accepts type Object in setFields.

alena1108: I can change getImage signature to accept launchConfig - it would probably be more clear. Then we retrieve imageId and registryCredential inside the method. Whatever you prefer.

sonchang: Hmm.  What if we casted it in createServiceInstances() and changed getImage to accept a Long?

alena1108: I can do that too

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/RancherConfigToComposeFormatter.java
```diff
@@ -0,0 +1,7 @@
+package io.cattle.platform.servicediscovery.service.impl;
+
+import io.cattle.platform.servicediscovery.resource.ServiceDiscoveryConfigItem;
+
+public interface RancherConfigToComposeFormatter {
+    public Object format(ServiceDiscoveryConfigItem item, Object valueToTransform);
```
sonchang: Can we add a comment or javadoc to mention that it returns null if no special formatting is necessary?  I actually think moving forward, we should probably do that for almost all interfaces.  But I'm sure people will want to debate it.

For this function, my initial thought was it would just return valueToTransform itself if no special formatting was done.  So I was confused when it returned null for almost all items.  Until I saw how it was being used in the caller.

sonchang:
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/RancherConfigToComposeFormatter.java
alena1108: Good suggestion. I will return the non-transformed value instead

alena1108: Actually I'm taking my words back. Returning the null indicates that the formatter didn't apply, and we should continue looping through the rest of formatters. The first not-null response indicates that the formatter modified the data, and the loop breaks.

sonchang: Ah.  I'm good with that too.  I guess if we can document the interface, that'd be great!

resources/content/schema/base/environment.json
```diff
@@ -1,7 +1,8 @@
 {
     "resourceFields": {
         "name":{
-            "required": true
+            "required": true,
+            "validChars": "[a-zA-Z0-9]"
```
cjellick: Do you want minLength=1 as well?

alena1108: good idea

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -12,6 +12,11 @@ def nsp(super_client, sim_context):
     return nsp
 
 
+@pytest.fixture(scope='module')
```
cjellick: What is your intention in making this scoped to the module? Because by making it module scope, you're saying "Run this once for the module and cache and re-use the result", right?

Also, really, you aren't using it as a fixture, you're just calling it like a normal function, so you can probably just drop the fixture decorator all together.

alena1108: ok, will fix

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/RancherImageToComposeFormatter.java
```diff
@@ -0,0 +1,19 @@
+package io.cattle.platform.servicediscovery.service.impl;
+
+import io.cattle.platform.servicediscovery.resource.ServiceDiscoveryConfigItem;
+
+public class RancherImageToComposeFormatter implements RancherConfigToComposeFormatter {
+    private static final String IMAGE_PREFIX = "docker:";
```
cjellick: Is there anything to be concerned here with your integration tests that use sim_contexts and have image uuids that look like `sim:foo:bar`?

You're integration tests do asserts like:
```
assert service.launchConfig.imageUuid == image_uuid
assert container.imageUuid == image_uuid
```

I guess the only issue is simply that you aren't testing this format logic in the integration test, which isnt a big deal

alena1108: At the moment, we don't have tests for docker-compose.yml export. I think it should be added to the validation set though - export the config, and try to run it with docker-compose. Testing it with the data generated on simulator setup, would break anyway.

cjellick:
LGTM

PR #356, opened 2015-04-17
Fix bug for exec/logs/stats for native containers

cjellick:


code/implementation/docker/common/src/main/java/io/cattle/platform/docker/util/DockerUtils.java
```diff
@@ -33,4 +35,15 @@ public static Host getHostFromContainer(ObjectManager objectManager, Instance in
         return getHostFromContainer(objectManager, instance, DockerHostConstants.KIND_DOCKER);
     }
 
+    /**
+     * The docker API allows for identification of a container by name or id.
+     * This method will return the docker container id if it is available.
+     * If it isn't available, the rancher container uuid will be returned, since that is currently used
+     * as the name for containers that do not have external id set.
```
hibooboo2: I would change this so that it is more clear that external id given by docker to a container.

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/util/DockerUtils.java
```diff
@@ -33,4 +35,18 @@ public static Host getHostFromContainer(ObjectManager objectManager, Instance in
...
     }
 
+    /**
+     * The docker API allows for identification of a container by name or id.
+     * This method will return the docker container id (which is stored in the
+     * externalId field of the instance) if it is available. If it isn't
+     * available, the instance uuid will be returned, since that is currently
+     * used as the name for docker containers that do not have the docker id
+     * set.
+     *
```
cjellick: Message update @hibooboo2. Make more sense?

hibooboo2: Yes it does. I think you could maybe explain why there would be a uuid but not an id. But i don't it matters to much.

tests/integration/cattletest/core/test_host_stats.py
```diff
@@ -54,8 +54,8 @@ def test_stats_container(admin_client, docker_context):
     stats_access = container.stats()
 
     assert stats_access.token.index('.') > 0
```
alena1108: @cjellick is this test just for native containers, Craig? And if its for both - regular, and native containers - does externalId get set to UUID for regular containers?

cjellick: @alena1108 Yes, externalId now gets set for all containers via this line:
https://github.com/rancherio/python-agent/blob/1b87b33d26dcb0bc875b4982a2de29e0bc1deb54/cattle/plugins/docker/compute.py#L508.

cjellick: I should add that the fallback to uuid in my `getDockerIdentifier()` is for backwards compatibility (to support containers created prior to the initial native-docker commit).

alena1108:
LGTM

sonchang:
LGTM

PR #357, opened 2015-04-17
hopefully fixes intermittent cluster TFs. looks like test_host_purge rem...

sonchang:
...oves main host
@alena1108 @cjellick 
I tried to take this into account when I first wrote the test, but looks like there are still issues with it.
What I noticed when I took a closer look was that test_host_purge() looked like it was wiping out the main  host (non-fake one).  In my case, boot2docker.  But I guess locally, it re-registered (amazing and oddly very quickly).  That's why I didn't see the TFs when running locally.  But it would explain why the failures are intermittent and also why the failures sometimes occur in different tests.
I'm taking the test out temporarily unless I have more time to research our test simulator framework better.  It's passed 2 drone tests now.

### DIFF
--- a/tests/integration/cattletest/core/test_cluster.py
+++ b/tests/integration/cattletest/core/test_cluster.py
@@ -96,6 +96,9 @@ def test_cluster_add_remove_host_actions(admin_client, super_client,
         lambda x: len(x.hosts()))
 
 
+# temporarily skipping since this was inadvertently deleting the
+# real host causing downstream TFs
+@pytest.mark.skipif('True')
 def test_host_purge(admin_client, super_client):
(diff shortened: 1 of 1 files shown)


alena1108:
@sonchang I believe there is a way to mark the test as "to skip" to exclude it from the pytest...but if you feel like its better to remove it, I'm fine with either

sonchang:
Added skip annotation with comments instead of deleting test

cjellick:
LGTM

alena1108:
LGTM

sonchang:
Had to change annotation slightly.  It was skipping all the cluster tests before.

PR #358, opened 2015-04-20
Make stateTransitions visible

cjellick:
The `processDefinition.stateTransitions` field is set as readable, but the resource to which it points (`stateTransition`) is not. So, the field doesn't show up in the resource. This fixes that.

### DIFF
--- a/resources/content/schema/admin/admin-auth.json
+++ b/resources/content/schema/admin/admin-auth.json
@@ -97,6 +97,8 @@
         "processDefinition.resourceType" : "r",
         "processDefinition.stateTransitions" : "r",
 
+        "stateTransition" : "r",
+
         "processExecution" : "r",
         "processExecution.log" : "r",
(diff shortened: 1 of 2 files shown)


resources/content/schema/admin/admin-auth.json
```diff
@@ -97,6 +97,8 @@
         "processDefinition.resourceType" : "r",
         "processDefinition.stateTransitions" : "r",
 
+        "stateTransition" : "r",
```
alena1108: Craig, should we also set the R access on the stateTransition object's fields? like we do for other resources. Or for this particular use case, setting access rights on object itself is enough?

sonchang:
LGTM

resources/content/schema/admin/admin-auth.json
sonchang: I was always a little curious about this as well

alena1108:
LGTM

PR #363, opened 2015-04-23
service discovery bug fixes

alena1108:
Addresses various Service Discovery bugs as well as refactored the way we handle service instance.create. Now instance and service->instance map objects get created in a single transaction. 

Bugs that should be fixed as a part of this PR:

1) When generate service instance name, take into account that existing instances might be renamed. So if the service with scale=3 has instances named:

* env_service_1
* randomName
* env_service_3

and scale up to 4 was requested, the new instance with name env_service_3 should be started

https://github.com/rancherio/rancher/issues/561
https://github.com/rancherio/rancher/issues/624

2) Don't automatically remove service-instance map when instance gets removed (as the instance can get restored)

Service link to removed instance can be removed only if activate/update operation is called on the service; or when instance gets purged.

https://github.com/rancherio/rancher/issues/527
https://github.com/rancherio/rancher/issues/570

3) Renaming environment and service should rename corresponding instances names. The name change doesn't get propagated to the backend instances till the service deactivated/activated again

https://github.com/rancherio/rancher/issues/551

4) Scale down should start stopped instances if they are within the scale

https://github.com/rancherio/rancher/issues/621

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
@@ -35,6 +35,7 @@ public class ServiceCreateValidationFilter extends AbstractDefaultResourceManage
             ValidationErrorCodes.throwValidationError(ValidationErrorCodes.INVALID_CHARACTERS,
                     "name");
         }
+
         return super.create(type, request, next);
     }
 }
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
@@ -3,16 +3,40 @@ package io.cattle.platform.servicediscovery.dao.impl;
 import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
 import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.SERVICE_EXPOSE_MAP;
 import io.cattle.platform.core.constants.CommonStatesConstants;
+import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.model.Service;
+import io.cattle.platform.core.model.ServiceExposeMap;
 import io.cattle.platform.core.model.tables.records.InstanceRecord;
+import io.cattle.platform.core.model.tables.records.ServiceExposeMapRecord;
 import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
+import io.cattle.platform.deferred.util.DeferredUtils;
+import io.cattle.platform.object.ObjectManager;
+import io.cattle.platform.object.process.ObjectProcessManager;
+import io.cattle.platform.object.process.StandardProcess;
 import io.cattle.platform.servicediscovery.api.dao.ServiceExposeMapDao;
+import io.cattle.platform.servicediscovery.service.ServiceDiscoveryService;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+import javax.inject.Inject;
 
 public class ServiceExposeMapDaoImpl extends AbstractJooqDao implements ServiceExposeMapDao {
+
+    @Inject
+    ObjectManager objectManager;
+
+    @Inject
+    ObjectProcessManager objectProcessManager;
+
+    @Inject
+    ServiceDiscoveryService sdService;
+
     @Override
-    public List<? extends Instance> listNonRemovedInstancesForService(long serviceId) {
+    public List<? extends Instance> listActiveServiceInstances(long serviceId) {
         return create()
                 .select(INSTANCE.fields())
                 .from(INSTANCE)
@@ -26,7 +50,7 @@ public class ServiceExposeMapDaoImpl extends AbstractJooqDao implements ServiceE
     }
 
     @Override
-    public Instance getServiceInstance(long serviceId, String instanceName) {
+    public Instance getActiveServiceInstance(long serviceId, String instanceName) {
         List<? extends Instance> instances = create()
                 .select(INSTANCE.fields())
                 .from(INSTANCE)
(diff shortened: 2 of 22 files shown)


alena1108:
@cjellick @sonchang could you please review this PR? all the fixes are described in the summary; there is a separate commit per each bullet.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -42,4 +67,74 @@ public Instance getServiceInstance(long serviceId, String instanceName) {
...
+        }
+        return instances;
+    }
+
+    @Override
+    public void updateEnvironmentName(Environment env) {
+        List<? extends Service> services = objectManager.mappedChildren(env, Service.class);
+        List<Instance> instances = new ArrayList<>();
+        for (Service service : services) {
+            if (service.getRemoved() == null && !service.getState().equalsIgnoreCase(CommonStatesConstants.REMOVED))
```
sonchang: I wonder should we filter out CommonStatesConstants.REMOVING as well.  Probably not a big deal since we're just updating the instance names.

alena1108: Yup, good idea. Will fix

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceDiscoveryInstanceStartPostListener.java
```diff
@@ -48,31 +50,34 @@
...
-                    ServiceDiscoveryConstants.FIELD_SERVICE_ID);
-            if (requestedServiceId != null) {
-                serviceId = requestedServiceId.longValue();
+            // for regular, non lb instance, the instance->service map gets created as a part of the instance creation
+            // (within the same transaction), so it should exist at this point
+            List<? extends ServiceExposeMap> instanceServiceMap = mapDao.findNonRemoved(ServiceExposeMap.class,
+                    Instance.class,
+                    instance.getId());
+            if (instanceServiceMap.isEmpty()) {
+                return null;
```
sonchang: Do we need to be concerned if the instance exists but there is no mapping?

alena1108: No. Service instance and its map gets created in transaction.

sonchang: Just because I know I'm going to forget in the future, can we add a comment here.  Something like: "not service instance"?

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
```diff
@@ -4,6 +4,8 @@
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.iaas.api.filter.common.AbstractDefaultResourceManagerFilter;
 import io.cattle.platform.object.ObjectManager;
+import io.cattle.platform.object.util.DataUtils;
```
sonchang: Unused imports?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -42,4 +66,95 @@ public Instance getServiceInstance(long serviceId, String instanceName) {
         }
         return instances.get(0);
     }
+
+    @Override
+    public Instance createServiceInstance(Map<Object, Object> properties, Service service, String instanceName) {
+        Map<String, Object> props = objectManager.convertToPropertiesFor(Instance.class,
+                properties);
+        final Instance instance = objectManager.create(Instance.class, props);
+            objectManager.create(ServiceExposeMap.class,
```
sonchang: white space

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -42,4 +66,95 @@ public Instance getServiceInstance(long serviceId, String instanceName) {
...
+                    SERVICE_EXPOSE_MAP.INSTANCE_ID,
+                    instance.getId(), SERVICE_EXPOSE_MAP.SERVICE_ID, service.getId());
+        DeferredUtils.nest(new Callable<Object>() {
```
sonchang: I don't understand why we use DeferredUtils here

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
alena1108: removed, thx!

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceUpdatePostListener.java
```diff
@@ -9,14 +9,19 @@
...
 import io.cattle.platform.util.type.Priority;
 
+import java.util.Map;
+
 import javax.inject.Inject;
 import javax.inject.Named;
 
+/**
+ * this handler takes care of
+ * a) scaling down the service
```
sonchang: No longer handles scaling down since that's done in pre handler

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
alena1108: Darren requested to put it in. He wanted Create event to be generated before the process completes (and DeferredUtils.nest does it ); otherwise there were some delays in scheduling the start process.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceUpdatePostListener.java
alena1108: Will be done, thanks!

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -381,9 +386,104 @@ public long getServiceNetworkId(Service service) {
...
+                if (finalCount < requestedCount) {
+                    if (!usedIds.contains(i)) {
+                        serviceNames.add(generateServiceInstanceName(service, i));
+                        finalCount++;
+                        usedIds.add(i);
+                    }
+                }
+            }
+            // only after we got all "gap" names, get the rest
+            Collections.sort(usedIds);
```
sonchang: I'm a little confused about this part.  At this point, usedIds should be complete (original + any holes filled including potentially new ones if requestedCount is higher than before).  But finalCount will just be the # of holes filled (by holes, i'm once again including potentially new ones).

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -381,9 +386,104 @@ public long getServiceNetworkId(Service service) {
...
-        return String.format("%s_%s_%d", env.getName(), service.getName(), order + 1);
+        // get all existing instances to check if the name is in use by the instance of the same service
+        List<Integer> usedIds = new ArrayList<>();
+        // list all the instances
+        List<? extends ServiceExposeMap> instanceServiceMaps = exposeMapDao.getNonRemovedServiceInstanceMap(service
+                .getId());
+        
+        for (ServiceExposeMap instanceServiceMap : instanceServiceMaps) {
+            Instance instance = objectManager.loadResource(Instance.class, instanceServiceMap.getInstanceId());
+            if (isServiceGeneratedName(service, instance)) {
```
sonchang: seems a little bit fragile (esp. with the potential service renaming, although I can't checked locking yet).  but i'm ok if we fix edge cases in a later PR

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
alena1108: finalCount reflects number of instances we return; not just the number of gaps filled. And we do have to proceed in case finalCount < requested count. Think about situation like that:

Updated scale=5
Existing instances: env_svc_1, randomName1, env_svc3

So first we fill up the gap by creating the instance env_svc2. And then we create env_svc4 instance - after line 412

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
alena1108: Shouldn't get affected by the service renaming as update/activate operations are guarded by States. If the service is being activated/updated, another service update can't be executed.

But you are right - its not guarded against environment update. We should def address it in a subsequent PR

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -925,6 +876,233 @@ def test_set_service_links(super_client, admin_client,
...
+
+    instance_service_map = super_client. \
+        list_serviceExposeMap(serviceId=service.id, instanceId=instance2.id)
+    assert len(instance_service_map) == 1
+    wait_for_condition(
+        super_client, instance_service_map[0], _resource_is_active,
+        lambda x: 'State is: ' + x.state)
+
+    # 3. activate the service. The map should be gone
+    service.activate()
```
sonchang: Do we need to check whether a new instance and mapping is created to preserve the scale?

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -754,29 +698,33 @@ def test_remove_environment_w_active_svcs(super_client,
...
+def _validate_instance_start(service, super_client, name):
     instances = super_client. \
-        list_container(name="compose_" + service.name
-                            + "_" + number, state="running")
+        list_container(name=name)
     assert len(instances) == 1
     return instances[0]
 
 
 def test_valiate_service_scaleup_scaledown(super_client,
```
sonchang: oddly, noticed the misspelling of 'validate'

sonchang:
LGTM

tests/integration/cattletest/core/test_svc_discovery.py
alena1108: I do have a test case for that; this one was meant to test only destroyed instance case

tests/integration/cattletest/core/test_svc_discovery.py
alena1108: thanks! :) will fix

tests/integration/cattletest/core/test_svc_discovery.py
sonchang: ok

cjellick:
LGTM

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -925,6 +876,227 @@ def test_set_service_links(super_client, admin_client,
...
+    instance = wait_success(super_client, instance.stop())
+    assert instance.state == 'stopped'
+    instance = wait_success(super_client, instance.remove())
+    assert instance.state == 'removed'
+    return instance
+
+
+def test_destroy_service_instance(super_client,
+                                  admin_client, sim_context, nsp):
+    env = admin_client.create_environment(name=random_str())
```
alena1108: Sure, thanks for pointing it out

PR #365, opened 2015-04-23
first pass at cattle changes for SSL

sonchang:
Generally how things work:
Before, when we activated a cluster in HostActivate.java, we just deployed out a cluster agent: clusterManager.activateCluster()
Now, prior to calling clusterManager.activateCluster(), we call clusterManager.checkSslAgentInstances().
This essentially generates all the necessary cert/keys and deploys out the socat-ssl agents to the hosts within the cluster.

host1: docker (/var/run/docker.sock) <=> socat-ssl tcp:2377 =\
                                                                                                   } swarm <=> docker-py/pyagent
host2: docker (/var/run/docker.sock) <=> socat-ssl tcp:2377 =/

Things in this PR:
- Moved 'certificateId' out of the data fields originally specified for a cluster and into an actual column 'certificate_id' in the host table.  This is populated for both hosts and clusters.  For hosts, it stores the certs used for socat to act as a proxy.  For clusters, it stores the certs for swarm (these certs act as both client and server.  client to talk to the socat servers.  server for docker to connect to).

Things that are still missing:
It's still missing several key actions that I'll have to add later such as: 1) allow users to upload their own CA cert/key 2) download link for downloading client cert/keys 3) maybe an action to regenerate all certs

Also, I need to add the accountId into the mapping table similar to what alena did for the UI.


code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/process/lock/CertificateAuthorityLock.java
```diff
@@ -0,0 +1,11 @@
+package io.cattle.iaas.cluster.process.lock;
+
+import io.cattle.platform.lock.definition.AbstractBlockingLockDefintion;
+
+public class CertificateAuthorityLock extends AbstractBlockingLockDefintion {
+
+    public CertificateAuthorityLock(Long accountId) {
```
alena1108: Minor, but can we change the argument type to use primitive long instead? So null values don't slip in

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
```diff
@@ -235,33 +263,397 @@ public boolean evaluate(Instance obj) {
...
+            objectManager.persist(cluster);
+        }
 
         if (hostId == null) {
             processManager.scheduleStandardProcess(StandardProcess.DEACTIVATE, cluster, null);
         } else {
+            try {
+                createClusterCertificates(cluster);
+                createClientCertificates(cluster);
+            } catch (IdempotentRetryException e) {
```
alena1108: Son, is there a reason why we handle exceptions this way? Why are we catching/rethrowing?

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
```diff
@@ -235,33 +263,397 @@ public boolean evaluate(Instance obj) {
...
+        GeneralNames subjectAltName = new GeneralNames(new DERSequence(new GeneralName[] {
+                new GeneralName(GeneralName.iPAddress, ipAddress.getAddress()),
+                new GeneralName(GeneralName.iPAddress, "127.0.0.1")
+        }));
+        serverCertGen.addExtension(X509Extensions.SubjectAlternativeName, false, subjectAltName);
+
+        X509Certificate serverCrt = serverCertGen.generateX509Certificate(caKeyPair.getPrivate());
+        String serverCert = SshKeyGen.writePemObject(serverCrt);
+
+        // create cert record
```
alena1108: Would it make sense to wrap 2 updates below in a single transaction? so certificate gets created and the host gets updated with the cert in a transaction with the single point of failure

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
```diff
@@ -235,33 +263,397 @@ public boolean evaluate(Instance obj) {
...
+
+            Host managingHost = getManagingHost(cluster);
+            IpAddress ipAddress = clusterHostMapDao.getIpAddressForHost(managingHost.getId());
+
+            PKCS10CertificationRequest swarmCSR = new PKCS10CertificationRequest(
+                    "SHA256withRSA",
+                    new X500Principal("CN="),
+                    swarmKeyPair.getPublic(),
+                    null,
+                    swarmKeyPair.getPrivate());
```
alena1108: As the above, we should probably wrap it in transaction. It would make sense to extract this method, place it in Dao, and enable the transaction on it. As it looks very similar to the one called from the previous method

code/iaas/cluster-logic/src/main/resources/META-INF/cattle/system-services/cluster-logic-defaults.properties
```diff
@@ -1,4 +1,6 @@
-cluster.instance.name=Cluster Server
 cluster.insecuredocker.port=2375
 cluster.securedocker.port=2377
-cluster.config.items=cluster
\ No newline at end of file
+cluster.config.items=cluster
+cluster.ssl.instance.name=SSL Agent
+cluster.ssl.config.items=ssl-socat
+ssl-socat.deploy.wait.time.millis=120000
```
alena1108: new line at the end of the file :) guilty myself for missing it all the time

code/iaas/cluster-logic/src/main/java/io/cattle/iaas/cluster/service/impl/ClusterManagerImpl.java
sonchang: The catching and rethrowing of the IdempotentRetryException was to allow the idempotent checks to work properly.  When we have idempotent checks=true, the framework throws an IdempotentRetryException (that's not meant to be caught here).

I might be able to change the catching of Exception to be more specific. It was Exception because at one point, a lot of the functions in SshKeyGen just threw Exception rather than something more specific.

alena1108: Yes, I get why we rethrow idempotentRetry exception here; I'm just not clear why we are catching the rest and wrapping it up in a RuntimeException. As it can potentially convert CheckedExeptions into Runtime ones..

ibuildthecloud:
We are going to take a different approach for Swarm.  Closing for now.

PR #366, opened 2015-04-23
instance.stop - don't deactivate purged child objects

alena1108:
And LB fixes

    1) fixed the tests to always wait for LB instance to come to final state,
     otherwise simulator cleanup can result in error
    2) lb.removeHost - don't trigger LB-agent instance cleanup if the host
    itself is in removing/removed state. Instance cleanup will be handled by
    generic host cleanup process where all instances on host are marked as
    removed
    3) LB config update - create LB-agent instance(s) only on active hosts

### DIFF
--- a/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerInstanceRemovePostListener.java
+++ b/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerInstanceRemovePostListener.java
@@ -57,12 +57,15 @@ public class LoadBalancerInstanceRemovePostListener extends AbstractObjectProces
     private void deleteLoadBalancerInstance(Instance instance) {
         Agent lbAgent = objectManager.loadResource(Agent.class, instance.getAgentId());
 
-        // try to remove first
-        try {
-            objectProcessManager.scheduleStandardProcess(StandardProcess.REMOVE, lbAgent, null);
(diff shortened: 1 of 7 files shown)


alena1108:
@ibuildthecloud Darren, can you pls review. This PR takes care of Lb integration test fixes (see more info in the first comment)

ibuildthecloud:
Let me think this through a bit more.  The logic seems brittle when we start looking for specific states.

ibuildthecloud:
What exactly does this fix?  Do certain errors go away or does it run faster??

alena1108:
@ibuildthecloud there are 3 fixes:

1) When host is removed from the LB as a result of host.remove, don't try to cleanup LB system instance , but rather rely on the default host cleanup logic to mark the instance as removed.I also removed the test checking that lb instance was gone after host removal, because I believe it should be tested in host lifecycle specific test where we verify that all instances on removed host are marked as removed. That test used to take a lot of time because I was waiting for the instance to be gone.

2) When apply haproxy configuration on lb instances, skip the hosts that are in Removed/Removing states

3) The checks I've put to instance.stop handler. This handler is invoked directly form instance.removed, so its quite possible to run into situation when instance failed to remove for the first time (some of its resources were removed/purged); and for a subsequent removal instance.stop will be called, and it will error out when we try to deactivate already purged resources.

ibuildthecloud:
LGTM, should I merge?

alena1108:
Sure.

PR #375, opened 2015-04-25
[WIP] Fix project bugs

hibooboo2:
This pr addresses 
Issue rancherio/rancher#607
Issue rancherio/rancher#606
Issue rancherio/rancher#605
Issue rancherio/rancher#595
Add integration test to verify rancherio/rancher#596 works in the api.

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/ExternalIdHandler.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/ExternalIdHandler.java
@@ -4,8 +4,8 @@ import io.cattle.platform.api.auth.ExternalId;
 
 public interface ExternalIdHandler {
 
-    public ExternalId transform(ExternalId externalId);
+    ExternalId transform(ExternalId externalId);
     
-    public ExternalId untransform(ExternalId externalId);
+    ExternalId untransform(ExternalId externalId);
 
 }
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
@@ -3,8 +3,10 @@ package io.cattle.platform.iaas.api.auth.dao.impl;
 import static io.cattle.platform.core.model.tables.AccountTable.*;
 import static io.cattle.platform.core.model.tables.CredentialTable.*;
 import static io.cattle.platform.core.model.tables.ProjectMemberTable.*;
+
 import io.cattle.platform.api.auth.ExternalId;
 import io.cattle.platform.archaius.util.ArchaiusUtil;
+import io.cattle.platform.core.constants.AccountConstants;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.ProjectConstants;
 import io.cattle.platform.core.dao.GenericResourceDao;
@@ -15,14 +17,14 @@ import io.cattle.platform.core.model.tables.records.ProjectMemberRecord;
 import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
 import io.cattle.platform.iaas.api.auth.ProjectLock;
 import io.cattle.platform.iaas.api.auth.dao.AuthDao;
+import io.cattle.platform.iaas.api.auth.github.GithubUtils;
 import io.cattle.platform.iaas.api.auth.github.resource.Member;
 import io.cattle.platform.lock.LockCallback;
 import io.cattle.platform.lock.LockManager;
 import io.cattle.platform.object.ObjectManager;
+import io.cattle.platform.object.meta.ObjectMetaDataManager;
 import io.cattle.platform.object.process.ObjectProcessManager;
 import io.cattle.platform.object.process.StandardProcess;
-import io.github.ibuildthecloud.gdapi.exception.ClientVisibleException;
-import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -33,6 +35,8 @@ import java.util.Set;
 
 import javax.inject.Inject;
 
+import io.cattle.platform.object.util.ObjectUtils;
+import io.cattle.platform.process.common.handler.AbstractObjectProcessLogic;
 import org.apache.commons.lang3.StringUtils;
 import org.jooq.Condition;
 import org.jooq.SelectQuery;
@@ -59,7 +63,7 @@ public class AuthDaoImpl extends AbstractJooqDao implements AuthDao {
         return create()
                 .selectFrom(ACCOUNT)
                 .where(ACCOUNT.STATE.eq(CommonStatesConstants.ACTIVE)
-                        .and(ACCOUNT.KIND.eq("admin")))
+                        .and(ACCOUNT.KIND.eq(AccountConstants.ADMIN_KIND)))
                 .orderBy(ACCOUNT.ID.asc()).limit(1).fetchOne();
     }
 
@@ -69,7 +73,7 @@ public class AuthDaoImpl extends AbstractJooqDao implements AuthDao {
                 .selectFrom(ACCOUNT)
                 .where(
                         ACCOUNT.ID.eq(id)
-.and(ACCOUNT.STATE.ne(CommonStatesConstants.PURGED))
+                                .and(ACCOUNT.STATE.ne(CommonStatesConstants.PURGED))
                                 .and(ACCOUNT.REMOVED.isNull())
                 ).fetchOne();
     }
@@ -132,12 +136,6 @@ public class AuthDaoImpl extends AbstractJooqDao implements AuthDao {
         return resourceDao.createAndSchedule(Account.class, objectManager.convertToPropertiesFor(Account.class, properties));
     }
 
-    public Account createDefaultProject(Account account) {
-        Account defaultProject = createProject(account.getName() + ProjectConstants.PROJECT_DEFAULT_NAME, null);
-        objectManager.setFields(account, ACCOUNT.PROJECT_ID, defaultProject.getId());
-        return defaultProject;
-    }
-
     @Override
     public Account getAccountByUuid(String uuid) {
         return create()
@@ -184,7 +182,7 @@ public class AuthDaoImpl extends AbstractJooqDao implements AuthDao {
             projects.addAll(create()
                     .selectFrom(ACCOUNT)
                     .where(ACCOUNT.KIND.eq(ProjectConstants.TYPE)
-                            .and(ACCOUNT.STATE.eq(CommonStatesConstants.ACTIVE)))
+                            .and(ACCOUNT.REMOVED.isNull()))
                     .orderBy(ACCOUNT.ID.asc())
                     .fetch());
             return projects;
@@ -328,7 +326,7 @@ public class AuthDaoImpl extends AbstractJooqDao implements AuthDao {
                 HashSet<Member> create = (HashSet<Member>) ((HashSet<Member>) members).clone();
                 HashSet<Member> delete = (HashSet<Member>) ((HashSet<Member>) otherPreviosMembers).clone();
                 for (Member member : members) {
-                    if (delete.remove(member)){
+                    if (delete.remove(member)) {
                         create.remove(member);
                     }
                 }
@@ -365,37 +363,43 @@ public class AuthDaoImpl extends AbstractJooqDao implements AuthDao {
     }
 
     @Override
-    public Account setDefaultProject(Account project, long accountId) {
-        if (project == null) {
-            throw new ClientVisibleException(ResponseCodes.BAD_REQUEST, "Failed", "No project specified.", null);
-        }
-        Account account = getAccountById(accountId);
-        if (accountId != project.getId() && project.getKind().equalsIgnoreCase(ProjectConstants.TYPE)
-                && !account.getKind().equalsIgnoreCase(ProjectConstants.TYPE)) {
-            int updateCount = create()
-                    .update(ACCOUNT)
-                    .set(ACCOUNT.PROJECT_ID, project.getId())
-                    .where(ACCOUNT.ID
-                            .eq(accountId))
-                    .execute();
-
-            if (1 != updateCount) {
-                throw new ClientVisibleException(ResponseCodes.NOT_MODIFIED, "Failed", "Failed to update account default project.", null);
+    public void ensureAllProjectsHaveNonRancherIdMembers(ExternalId externalId) {
+        //This operation is expensive if there are alot of projects and members however this is
+        //only called when auth is being turned on. In most cases this will only be called once.
+        Member newMember = new Member(externalId, "owner");
+        Set<ExternalId> externalIds = new HashSet<>();
+        externalIds.add(new ExternalId(String.valueOf(getAdminAccount().getId()), ProjectConstants.RANCHER_ID));
+        List<Account> allProjects = getAccessibleProjects(externalIds, true, null);
+        for(Account project: allProjects){
+            List<? extends ProjectMember> members = getActiveProjectMembers(project.getId());
+            boolean hasNonRancherMember = false;
+            for (ProjectMember member: members){
+                if (!member.getExternalIdType().equalsIgnoreCase(ProjectConstants.RANCHER_ID)){
+                    hasNonRancherMember = true;
+                } else {
+                    deactivateThenRemove(member);
+                }
             }
-            return project;
-        } else {
-            throw new ClientVisibleException(ResponseCodes.BAD_REQUEST, "Failed", "Cannot Use your account as a default project", null);
-
+            if (!hasNonRancherMember) {
+                createProjectMember(project, newMember);
+            }
         }
     }
 
-    @Override
-    public Account getDefaultProject(Account account) {
-        Account project = getAccountById(account.getProjectId());
-        if (project == null){
-            throw new ClientVisibleException(ResponseCodes.FORBIDDEN, "DefaultProjectNotFound", "Current Default Project Not Found. Please Select Another.", null);
+    private void deactivateThenRemove(ProjectMember member) {
+        Object state = ObjectUtils.getPropertyIgnoreErrors(member, ObjectMetaDataManager.STATE_FIELD);
+
+        if (CommonStatesConstants.ACTIVE.equals(state)) {
+            objectProcessManager.executeStandardProcess(StandardProcess.DEACTIVATE, member, null);
+            member = objectManager.reload(member);
         }
-        return project;
+
+        if (CommonStatesConstants.PURGED.equals(state)) {
+            return;
+        }
+
+        objectProcessManager.executeStandardProcess(StandardProcess.REMOVE, member, null);
+        return;
     }
 
     @Override
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubTokenHandler.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubTokenHandler.java
@@ -2,6 +2,7 @@ package io.cattle.platform.iaas.api.auth.github;
 
 import io.cattle.platform.api.auth.ExternalId;
 import io.cattle.platform.archaius.util.ArchaiusUtil;
+import io.cattle.platform.core.constants.AccountConstants;
 import io.cattle.platform.core.model.Account;
 import io.cattle.platform.iaas.api.auth.TokenHandler;
 import io.cattle.platform.iaas.api.auth.dao.AuthDao;
@@ -12,13 +13,10 @@ import io.cattle.platform.object.ObjectManager;
 import io.cattle.platform.token.TokenService;
 import io.cattle.platform.util.type.CollectionUtils;
 import io.github.ibuildthecloud.gdapi.context.ApiContext;
-import io.github.ibuildthecloud.gdapi.exception.ClientVisibleException;
 import io.github.ibuildthecloud.gdapi.request.ApiRequest;
-import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -29,13 +27,9 @@ import java.util.Set;
 import javax.inject.Inject;
 
 import org.apache.commons.lang3.ObjectUtils;
-import org.apache.commons.lang3.StringUtils;
 
-import com.google.common.base.Function;
-import com.google.common.collect.Collections2;
 import com.netflix.config.DynamicBooleanProperty;
 import com.netflix.config.DynamicLongProperty;
-import com.netflix.config.DynamicStringProperty;
 
 public class GithubTokenHandler implements TokenHandler {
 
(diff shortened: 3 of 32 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/ExternalIdHandler.java
```diff
@@ -4,8 +4,8 @@
 
 public interface ExternalIdHandler {
 
-    public ExternalId transform(ExternalId externalId);
+    ExternalId transform(ExternalId externalId);
```
alena1108: Did we remove public because we want this interface to be available only within auth-logic package?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -365,37 +360,28 @@ public ProjectMember createProjectMember(Account project, Member member) {
...
+        Set<ExternalId> externalIds = new HashSet<>();
+        externalIds.add(new ExternalId(String.valueOf(getAdminAccount().getId()), ProjectConstants.RANCHER_ID));
+        List<Account> allProjects = getAccessibleProjects(externalIds, true, null);
+        for(Account project: allProjects){
+            List<? extends ProjectMember> members = getActiveProjectMembers(project.getId());
+            boolean hasNonRancherMember = false;
+            for (ProjectMember member: members){
+                if (!member.getExternalIdType().equalsIgnoreCase(ProjectConstants.RANCHER_ID)){
+                    hasNonRancherMember = true;
+                    objectProcessManager.executeStandardProcess(StandardProcess.DEACTIVATE, member, null);
```
alena1108: James, could you briefly explain the purpose of this method?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -184,7 +179,7 @@ public void updateAccount(Account account, String name, String kind, String exte
             projects.addAll(create()
                     .selectFrom(ACCOUNT)
                     .where(ACCOUNT.KIND.eq(ProjectConstants.TYPE)
-                            .and(ACCOUNT.STATE.eq(CommonStatesConstants.ACTIVE)))
+                            .and(ACCOUNT.STATE.ne(CommonStatesConstants.PURGED)))
```
alena1108: Should we also check for Purging state?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
```diff
@@ -125,4 +148,58 @@ public String getTeamOrgById(String id) {
...
+            case "unrestricted":
+                break;
+            default:
+                throw new ClientVisibleException(ResponseCodes.UNAUTHORIZED);
+        }
+        return true;
+    }
+
+    protected String getWhitelistedUser(List<String> idList) {
+        if (idList == null) {
```
alena1108: We should probably add "|| idList.isEmpty" condition:

if (idList == null || idList.isEmpty) {
return null
}

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
```diff
@@ -125,4 +148,58 @@ public String getTeamOrgById(String id) {
...
+            return null;
+        }
+        if (StringUtils.equals(WHITELISTED_USERS.get(), "*")) {
```
alena1108: Curious - is it possible to have null idList passed in, while WHITELISTED_USERS="*"? If yes, why null idList takes precedence (as we return null before checking whiltelisted users)

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
```diff
@@ -125,4 +148,58 @@ public String getTeamOrgById(String id) {
         Map<String, String> teamToOrg = (Map<String, String>) jsonData.get("teamToOrg");
         return  teamToOrg.get(id);
     }
+
+    public boolean isAllowed(List<String> idList, Set<ExternalId> externalIds){
+        boolean whiteListed = getWhitelistedUser(idList) != null;
```
alena1108: is getWhiteListedUser being called just from this method? If so - and we care only if the user is whitelisted or not - , would it make sense to change the return type of getWhiteListedUser to boolean? Its also a big unclear why the method with signature:

String getWhitelistedUser(List<String> idList

returns only one id from the idList passed in:

for (String id : idList) {
+            if (whitelist.contains(id)) {
+                return id;
+            }
+        }

I'm sure there is a logic behind it, but could you please elaborate it?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/ProjectResourceManager.java
```diff
@@ -84,7 +86,7 @@ public Object listInternal(SchemaFactory schemaFactory, String type, Map<Object,
                     throw new ClientVisibleException(ResponseCodes.NOT_FOUND);
                 }
             }
-            policy.grantObjectAccess(project);
+            giveProjectAccess(project);
```
alena1108: We should probably pass policy object as an argument to giveProjectAccess call as we already retrieved it (line 70), and there is no need to call it for the second time (line 127)

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/ExternalIdHandler.java
hibooboo2: Removed public because it is redundant when defined in an interface.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
vincent99: While access control is disabled there are no projectMember entries create for projects.  So when you turn auth on this creates a projectMember entry for each project that has none, making the github user who just turned on auth the owner

hibooboo2: This is for the use case of turning auth of and on to make sure admin has access to all projects that don't have a github user that can access them when auth is turned on.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
hibooboo2: Untill we do the refactor we can leave this as is. Since we are removing purged and purging.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
hibooboo2: This method was written just to gram the first id from the list that is whitelisted. I didn't change it from the original impl as I didn't want to spend time reading it and seeing it if needed to be changed as it already worked. I can refactor it to make it better but didn't see the need.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
hibooboo2: As far as list being empty it will never be empty.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
alena1108: ok, we can always do it later

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
hibooboo2: Null idList takes precedence because if your list is null then you don't have any one to say is white-listed. There for you get no access.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/ProjectResourceManager.java
hibooboo2: I liked the simplicity of this method signature and since it is the same pointer and doesn't make a new object i went for that way. But I could do that as well.

alena1108: Having 2 arguments doesn't make it look complicated :) Plus we make it more simple by avoiding unnecessary calls.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
alena1108: I still think we should change it so we don't run into race conditions before the refactor changes come in.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
vincent99: Is whitelist=* any different than mode=unrestricted?  Seems like it could just be removed

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -365,37 +360,28 @@ public ProjectMember createProjectMember(Account project, Member member) {
...
+        List<Account> allProjects = getAccessibleProjects(externalIds, true, null);
+        for(Account project: allProjects){
+            List<? extends ProjectMember> members = getActiveProjectMembers(project.getId());
+            boolean hasNonRancherMember = false;
+            for (ProjectMember member: members){
+                if (!member.getExternalIdType().equalsIgnoreCase(ProjectConstants.RANCHER_ID)){
+                    hasNonRancherMember = true;
+                    objectProcessManager.executeStandardProcess(StandardProcess.DEACTIVATE, member, null);
+                    objectProcessManager.executeStandardProcess(StandardProcess.REMOVE, member, null);
+                }
```
alena1108: can we use objectProcessManager.deactivateThenRemove to replace 2 calls on lines 376 and 377? This method already handles checking if the start state is valid for both deactivate and remove

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
alena1108: On a second thought, are you ok with this method returning Removed and Removing entries? If you want any non-removed entry being returned, you can just return records with null Removed field

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
hibooboo2: No that is not blocking.

hibooboo2: Ill look i thought it was non blocking. This needs to be blocking.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/ProjectResourceManager.java
hibooboo2: Complicated may have been the wrong word. But i will change it.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
hibooboo2: This is true. Removing.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
alena1108: If you decide to continue keeping it in this method, you should check for the valid states, otherwise you will run into ProcessCancelException if either deactivate or removed is called when the start state is invalid for those processes.

vincent99:
Note: https://github.com/rancherio/ui/pull/146 needs to be merged to coincide with this

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -365,37 +363,43 @@ public ProjectMember createProjectMember(Account project, Member member) {
...
+        List<Account> allProjects = getAccessibleProjects(externalIds, true, null);
+        for(Account project: allProjects){
+            List<? extends ProjectMember> members = getActiveProjectMembers(project.getId());
+            boolean hasNonRancherMember = false;
+            for (ProjectMember member: members){
+                if (!member.getExternalIdType().equalsIgnoreCase(ProjectConstants.RANCHER_ID)){
+                    hasNonRancherMember = true;
+                    deactvateThenRemove(member);
+                }
+                if (hasNonRancherMember) break;
```
sonchang: how many non-rancher_id members can there be in a project?  did you intend to just remove the first one found?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -365,37 +363,43 @@ public ProjectMember createProjectMember(Account project, Member member) {
...
-
         }
     }
 
-    @Override
-    public Account getDefaultProject(Account account) {
-        Account project = getAccountById(account.getProjectId());
-        if (project == null){
-            throw new ClientVisibleException(ResponseCodes.FORBIDDEN, "DefaultProjectNotFound", "Current Default Project Not Found. Please Select Another.", null);
+    private void deactvateThenRemove(ProjectMember member) {
```
sonchang: minor: typos

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -365,37 +363,43 @@ public ProjectMember createProjectMember(Account project, Member member) {
...
+        Object state = ObjectUtils.getPropertyIgnoreErrors(member, ObjectMetaDataManager.STATE_FIELD);
```
sonchang: Any reason why we're not just calling member.getState()?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
hibooboo2: I think i have typo here Hold on.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
hibooboo2: This was how the deactivateThenRemove was originally implemented.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubTokenHandler.java
```diff
@@ -47,20 +48,16 @@
...
-    private static final DynamicStringProperty WHITELISTED_ORGS = ArchaiusUtil.getString("api.auth.github.allowed.orgs");
-    private static final DynamicStringProperty WHITELISTED_USERS = ArchaiusUtil.getString("api.auth.github.allowed.users");
-    private static final DynamicStringProperty ACCESS_MODE = ArchaiusUtil.getString("api.auth.github.access.mode");
 
     @Inject
     ProjectResourceManager projectResourceManager;
     @Inject
     ObjectManager objectManager;
+    @Inject
+    GithubUtils githubUtils;
```
sonchang: Very minor nitpick.  Typically when I see Utils in the class name, I think of static functions rather than a singleton or bean.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -365,37 +363,43 @@ public ProjectMember createProjectMember(Account project, Member member) {
...
-            int updateCount = create()
-                    .update(ACCOUNT)
-                    .set(ACCOUNT.PROJECT_ID, project.getId())
-                    .where(ACCOUNT.ID
-                            .eq(accountId))
-                    .execute();
-
-            if (1 != updateCount) {
-                throw new ClientVisibleException(ResponseCodes.NOT_MODIFIED, "Failed", "Failed to update account default project.", null);
+    public void allProjectsHaveNonRancherIdMember(ExternalId externalId) {
```
sonchang: minor: possibly rename to something like "ensureAllProjectsHaveNonRancherIdMembers()".  otherwise, the method signature is a little confusing.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubTokenHandler.java
hibooboo2: I believe that this is not static because it has a few classes internally it uses that are injected and as a result needs to be a bean. I could be wrong i can rename it if you would like to something like GithubHelper

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubUtils.java
```diff
@@ -51,6 +65,10 @@ public String validateAndFetchAccountIdFromToken(String token) {
     @SuppressWarnings("unchecked")
     public Set<ExternalId> externalIds(String token) {
         Map<String, Object> jsonData = getJsonData(token);
+        return externalIds(jsonData);
+    }
+
+    private Set<ExternalId> externalIds(Map<String, Object> jsonData) {
```
sonchang: possibly rename to getExternalIdsFromJson() or internalGetExternalIds().  or possibly rename the line in line 66 as getExternalIdsFromToken() to avoid confusion with the other getExternalIds() below (line 125)

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/ProjectMemberResourceManager.java
```diff
@@ -90,15 +98,27 @@ protected Object listInternal(SchemaFactory schemaFactory, String type, Map<Obje
         return new Class<?>[0];
     }
 
-    public List<ProjectMember> setMembers(Account project, List<Map<String, String>> members) {
+    public List<ProjectMember> setMembers(Account project, List<Map<String, String>> membersProvided) {
+        List<Map<String, String>> members = membersProvided;
         List<ProjectMember> membersCreated = new ArrayList<>();
+        HashSet<Member> membersTransformed = new HashSet<>();
```
sonchang: If you don't need HashSet specific things, just declare as Set for membersTransformed.  No biggie though

ibuildthecloud:
LGTM

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/ProjectMemberResourceManager.java
```diff
@@ -64,20 +73,19 @@ protected Object listInternal(SchemaFactory schemaFactory, String type, Map<Obje
...
-            return members;
+            members =  authDao.getActiveProjectMembers(Long.valueOf(projectId));
+        } else {
+            members = authDao.getActiveProjectMembers(policy.getAccountId());
         }
-        List<? extends ProjectMember> members = authDao.getActiveProjectMembers(policy.getAccountId());
+        List<ProjectMember> membersToReturn = new ArrayList<>();
         for (ProjectMember member: members){
             member = untransform(member);
+            membersToReturn.add(member);
```
sonchang: haven't dug further, but it looks like untranform() can return a null.  are we ok with this?  or should an exception get thrown instead of returning null in untransform()

hibooboo2: We want that.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/TokenResourceManager.java
```diff
@@ -65,6 +65,6 @@ private Token getToken(ApiRequest request) throws IOException {
 
     @Override
     protected Object listInternal(SchemaFactory schemaFactory, String type, Map<Object, Object> criteria, ListOptions options) {
-        return new Token(null, null, null, null, SECURITY.get(), GITHUB_CLIENT_ID.get(), null, null, null);
+        return new Token(null, null, null, null, SECURITY.get(), GITHUB_CLIENT_ID.get(), null, null);
```
sonchang: minor, but typically something like this deserves a new constructor

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/ProjectMemberResourceManager.java
hibooboo2: This allows easy masking of member types. I could check for null and not add nulls to the list though.

hibooboo2: Example: When auth is off we use rancher_ids. This makes projects not require github. But when auth is on we use github externalIds and as a result we still respect the rancher_ids but when you list members they should be ignored if any exist.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/TokenResourceManager.java
hibooboo2: I agree

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/AdminAuthLookUp.java
```diff
@@ -29,35 +29,36 @@
...
+        return getAccountAccessInternal(projectId);
+    }
+
+    private AccountAccess getAccountAccessInternal(String projectId){
+        Account admin = authDao.getAdminAccount();
+        if (admin == null){
+            return null;
+        }
+        Account project = null;
+        if (projectId != null && !projectId.isEmpty()) {
```
sonchang: haven't had time to look at functional change, but it looks like we've taken out the:
!projectId.equalsIgnoreCase(ProjectConstants.USER)
check.  is this intentional?

sonchang:
LGTM if the changes I pointed out were intentional.  Otherwise, I don't mind tackling the minor things later.  I don't want to block people.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/AdminAuthLookUp.java
hibooboo2: Yes it is. Removing the USER constant as a requirement. If you don't send a project id the request is fulfilled as the user.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubTokenHandler.java
sonchang: Yeah.  That's why I'm suggesting a rename (since *Util classes typically are more just a bunch of static util functions).  More just a naming thing.

hibooboo2: So would GithubHelper be good?

PR #378, opened 2015-04-27
Fixed SD test to work with async exposeMap create

alena1108:
@ibuildthecloud Darren, this PR is to fix the test failure you observed yesterday on http://drone.rancher.io/github.com/rancherio/cattle/master/045185267a1362818df7eff6b6fe271b6a119e5d?


tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -131,7 +131,9 @@ def test_activate_single_service(super_client, admin_client, sim_context, nsp):
         list_serviceExposeMap(serviceId=service.id)
 
     assert len(instance_service_map) == 1
-    assert instance_service_map[0].state == 'active'
+    wait_for_condition(
```
ibuildthecloud: just curious, we do you use this long form? does `x = client.wait_success(x)` not work?

PR #386, opened 2015-04-29
More bug fixes for Service Discovery

alena1108:
1) SD external_links:  follow container:link semantic in config

https://github.com/rancherio/rancher/issues/631

2) When service is removed, cleanup service maps referencing service-to-remove as:

* consuming service
* consumed service

https://github.com/rancherio/rancher/issues/638
https://github.com/rancherio/rancher/issues/643

3) Don't allow null value for scale parameter:

https://github.com/rancherio/rancher/issues/634

4) LB AppCookieStickinessPolicy: marked cookie/maxLength/timeout as required in API as those are required by haproxy

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceConsumeMapDaoImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceConsumeMapDaoImpl.java
@@ -47,7 +47,7 @@ public class ServiceConsumeMapDaoImpl extends AbstractJooqDao implements Service
     }
 
     @Override
-    public List<? extends ServiceConsumeMap> findMapsToRemove(long serviceId) {
+    public List<? extends ServiceConsumeMap> findConsumedMapsToRemove(long serviceId) {
         return create()
                 .selectFrom(SERVICE_CONSUME_MAP)
                 .where(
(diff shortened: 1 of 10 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -273,7 +273,7 @@ private void populateExternalLinksForService(Service service, Map<String, Object
                         null);
                 String instanceName = getInstanceName(instance);
                 if (instanceName != null) {
-                    instanceLinksWithNames.put(linkName, instanceName);
+                    instanceLinksWithNames.put(instanceName, linkName);
```
sonchang: darn.  sorry I missed this the first time around

sonchang:
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
alena1108: no worries, totally my bad, should have paid more attention

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceConsumeMapDaoImpl.java
```diff
@@ -66,4 +66,15 @@ public ServiceConsumeMap findNonRemovedMap(long serviceId, long consumedServiceI
...
     }
 
+    @Override
+    public List<? extends ServiceConsumeMap> findConsumingMapsToRemove(long serviceId) {
+        return create()
+                .selectFrom(SERVICE_CONSUME_MAP)
+                .where(
+                        SERVICE_CONSUME_MAP.CONSUMED_SERVICE_ID.eq(serviceId)
+                                .and((SERVICE_CONSUME_MAP.REMOVED.isNull().
+                                        or(SERVICE_CONSUME_MAP.STATE.eq(CommonStatesConstants.REMOVING))))).
```
cjellick: Want to make sure I understand this query, you want to get maps "where removed is null or state is removing."
Is that right?
Seems to me like it should be "where removed is null and state is not removing"?

alena1108: Here I'm finding the maps To remove; so I specifically looking for maps with removed flag null, or state=Removing (records against which Remove process was called once). I follow the logic from genericMapDao here

cjellick:
LGTM

PR #388, opened 2015-04-30
Native docker lifecycle events

cjellick:


code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -61,76 +68,109 @@
...
 
-        ContainerEvent event = (ContainerEvent)state.getResource();
-        Map<String, Object> inspect = (Map<String, Object>)DataUtils.getFields(event).get(FIELD_DOCKER_INSPECT);
+        final ContainerEvent event = (ContainerEvent)state.getResource();
 
-        String rancherUuid = getRancherUuidLabel(inspect);
-        if ( rancherUuid != null ) {
-            return null;
-        }
+        HandlerResult result = lockManager.lock(new ContainerEventInstanceLock(event.getAccountId(), event.getExternalId()), new LockCallback<HandlerResult>() {
```
ibuildthecloud: In the ContainerEventInstanceLock I would probably hash the external id when constructing the lock key.  You can even just take the hashCode().  The reason is that this information could be sent from external so could have bad values in it.  For example, a "/" in a lock id can cause problems on ZooKeeper.

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/InstanceDaoImpl.java
```diff
@@ -59,4 +62,27 @@ public boolean isOnSharedStorage(Instance instance) {
                 .fetchInto(InstanceRecord.class);
     }
 
+    @Override
+    public Instance getInstanceByExternalIdOrUuid(Long accountId, String externalId, String uuid) {
+        Condition accountCondition = INSTANCE.ACCOUNT_ID.eq(accountId).and(INSTANCE.REMOVED.isNull());
```
ibuildthecloud: This may be a convoluted situation but something to think about.  On the cattle side we look for external id first and then uuid.  On the python-agent side we look for uuid first and then external id.  Could this cause problems?  I'm not sure I can think of any, but just bringing it up.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -61,76 +68,109 @@
...
+            @Override
+            public HandlerResult doWithLock() {
+                Map<String, Object> inspect = (Map<String, Object>)DataUtils.getFields(event).get(FIELD_DOCKER_INSPECT);
```
ibuildthecloud: Want to use io.cattle.platform.util.type.CollectionUtils.toMap(Object) here to a little more type safety?

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -61,76 +68,109 @@
...
-                        !InstanceConstants.STATE_STARTING.equals(ObjectUtils.getState(instance))) {
-                    try {
-                        getObjectProcessManager().scheduleProcessInstance(InstanceConstants.PROCESS_CREATE, instance, makeData());
-                    } catch (ProcessCancelException e) {
-                        // ignore
-                    }
-                }
-            }
-        });
+        if (!STATE_RUNNING.equals(ObjectUtils.getState(instance)) && !STATE_STARTING.equals(ObjectUtils.getState(instance))) {
```
ibuildthecloud: Why do you use ObjectUtils.getState() over just instance.getState()?

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -159,10 +199,10 @@ void setImage(Map<String, Object> inspect, Instance instance) {
...
         // Somewhat of a hack, but needed for testing against sim contexts
-        if ( !imageUuid.matches(IMAGE_KIND_PATTERN) ) {
+        if (!imageUuid.matches(IMAGE_KIND_PATTERN)) {
             imageUuid = IMAGE_PREFIX + imageUuid;
         }
-        DataAccessor.fields(instance).withKey(InstanceConstants.FIELD_IMAGE_UUID).set(imageUuid);
+        DataAccessor.fields(instance).withKey(FIELD_IMAGE_UUID).set(imageUuid);
         Image image = null;
         try {
             image = storageService.registerRemoteImage(imageUuid);
```
ibuildthecloud: io.cattle.platform.storage.service.StorageService.registerRemoteImage(String) is problematic.  I can't imagine the account_id is being set properly.  Also the lookup of the image will most likely cross accounts too.  This isn't specific to native docker but we need to fix.  registerRemoteImage() should really take an account id so that look ups are scoped and the proper account id gets set.

When registerRemoteImage() is ran during an API call, the account_id does get set properly because there is a thread local context variable that has the account_id in it.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -61,76 +68,109 @@
...
+                    }
 
-        return null;
+                    if (instance == null) {
+                        return null;
+                    }
+
+                    String state = ObjectUtils.getState(instance);
+                    if (EVENT_START.equals(status)) {
+                        if (STATE_RUNNING.equals(state) || STATE_STARTING.equals(state))
```
alena1108: There is also Restarting state in cattle...should we check against it as well?

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -61,76 +68,109 @@
...
+                            return null;
+
+                        objectProcessManager.scheduleProcessInstance(PROCESS_START, instance, makeData());
+                    } else if (EVENT_STOP.equals(status) || EVENT_DIE.equals(status)) {
+                        if (STATE_STOPPED.equals(state) || STATE_STOPPING.equals(state))
+                            return null;
+
+                        objectProcessManager.scheduleProcessInstance(PROCESS_STOP, instance, makeData());
+                    } else if (EVENT_DESTROY.equals(status)) {
+                        if (REMOVED.equals(state) || REMOVING.equals(state) || PURGED.equals(state) || PURGING.equals(state))
```
alena1108: Instead of checking for state being Purged or purging, we can also check for removed flag not being null. But its up to you, minor thing

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: @ibuildthecloud 
How does io.cattle.platform.core.dao.StoragePoolDao.findExternalActivePools() get scoped to a particular account? It doesn't have an account id in the query.

Called here
https://github.com/rancherio/cattle/blob/c552571c3d2003446cc5af9269563cc4da3258db/code/iaas/storage-service/src/main/java/io/cattle/platform/storage/service/impl/StorageServiceImpl.java#L38

cjellick: It appears the answer is that it doesn't. I can see images getting associated to storage pools that are not owned by the account that owns the image. This is outside of native-docker. This seems like a pretty big problem.

cjellick: Further down the rabbit hole, what _should_ the correct behavior of findExternalActivePools() be? Should it take hostId into account? Currently does not.

ibuildthecloud: This is all outside the scope of native docker.  I'll take up this work when I do one network per account.  You can merge as is.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: I kind of like that all these checks are consistently looking at the state. So, I think I'll leave it as-is if there is no other reason to change it.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: Looked into making this change, but CollectionUtils.toMap(Object) returns an empty map when the input is null. The input in this case the docker inspect and while a a null value is valid (no inspect is available for destroy events), an empty non-null map is not and it causes NPEs further down. And I'd rather just check whether or not inspect is null than its children as well.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -61,76 +67,104 @@
...
-                    instance.setAccountId(accountId);
-                    instance.setExternalId(externalId);
-                    instance.setNativeContainer(true);
-                    setName(inspect, instance);
-                    setNetwork(inspect, instance);
-                    setImage(inspect, instance);
-                    setHost(event, instance);
-                    instance = objectManager.create(instance);
+            public HandlerResult doWithLock() {
+                Map<String, Object> inspect = (Map<String, Object>)DataUtils.getFields(event).get(FIELD_DOCKER_INSPECT);
```
cjellick: @ibuildthecloud You suggested using CollectionUtils.toMap here, but the comment and my reply got hidden when I pushed my updates.  Re-including my reply:

Looked into making this change, but CollectionUtils.toMap(Object) returns an empty map when the input is null. The input in this case the docker inspect and while a a null value is valid (no inspect is available for destroy events), an empty non-null map is not and it causes NPEs further down. And I'd rather just check whether or not inspect is null than its children as well.

ibuildthecloud: That makes sense.  The only reason I brought it up is that this input theoretically comes from the user.  As long as the user can't get a non-map object into this field, I'm fine.

ibuildthecloud:
As long as the user can't cause a classcastexception.  LGTM

PR #391, opened 2015-05-01
Load Balancer refactoring to work with services/new allocator

alena1108:
This PR provides support for scaling up the load balancer as a part of LB service, and delegate the host picking part for haproxy instance start, to allocator. Before that user had to specify the host explicitly in loadBalancer.addHost API command (this behavior is still supported). Now you can say scale=3 on the load balancer service, and service.activate would launch 3 haproxy containers on the hosts picked by allocator, considering public ports' and other constraints. 

Implementation changes:

1) When no host is specified while scaling up the lb, record with null host_id is generated in load_balancer_host_map. Once lb instance is started on a particular host, the map is updated with that hostId.
2) LB agent uri is no longer constructed from loadBalancer/hostId. Instead its constructed from loadBalancerId/hostMapId

### DIFF
--- a/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
+++ b/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
@@ -32,6 +32,8 @@ import io.cattle.platform.lock.LockCallbackNoReturn;
 import io.cattle.platform.lock.LockManager;
 import io.cattle.platform.object.resource.ResourceMonitor;
 import io.cattle.platform.object.resource.ResourcePredicate;
+import io.cattle.platform.object.util.DataAccessor;
+import io.cattle.platform.object.util.DataUtils;
 import io.cattle.platform.process.common.handler.AbstractObjectProcessLogic;
 import io.cattle.platform.util.type.Priority;
 
@@ -45,6 +47,8 @@ import java.util.Set;
 
 import javax.inject.Inject;
 import javax.inject.Named;
+
+import org.jooq.tools.StringUtils;
 
 import com.netflix.config.DynamicStringListProperty;
 
@@ -208,8 +212,13 @@ public class LoadBalancerUpdateConfig extends AbstractObjectProcessLogic impleme
             List<? extends Instance> lbInstances = lbInstancesMap.get(lbId);
             // surround by lock
             for (final Instance lbInstance : lbInstances) {
-                final boolean hasActiveInstances = (targetDao.getLoadBalancerActiveInstanceTargets(lb.getId()).size() + targetDao
+                boolean lbHasActiveTargets = (targetDao.getLoadBalancerActiveInstanceTargets(lb.getId()).size() + targetDao
                         .getLoadBalancerActiveIpTargets(lb.getId()).size()) > 0;
+                List<String> ports = DataUtils.getFieldList(lbInstance.getData(), InstanceConstants.FIELD_PORTS,
+                        String.class);
+                boolean lbInstanceHasPortsSet = (ports != null && !ports.isEmpty());
+                final boolean hasActiveInstances = lbHasActiveTargets || lbInstanceHasPortsSet;
+
                 final Map<Integer, Port> portsToCreate = new HashMap<Integer, Port>();
                 final Map<Integer, Port> portsToRemove = new HashMap<Integer, Port>();
                 lockManager.lock(new LoadBalancerInstancePortLock(lbInstance), new LockCallbackNoReturn() {
--- a/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
+++ b/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
@@ -5,11 +5,12 @@ import io.cattle.platform.agent.instance.factory.AgentInstanceFactory;
 import io.cattle.platform.archaius.util.ArchaiusUtil;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.InstanceConstants;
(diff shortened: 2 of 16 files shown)


alena1108:
@ibuildthecloud @cjellick @sonchang could you please review

sonchang:
I'll take a look

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
```diff
@@ -122,7 +141,7 @@ public Instance getLoadBalancerInstance(LoadBalancer loadBalancer, long hostId)
         return lbInstance;
     }
 
-    protected Agent getLoadBalancerAgent(LoadBalancer loadBalancer, long hostId) {
+    protected Agent getLoadBalancerAgent(LoadBalancer loadBalancer, LoadBalancerHostMap hostId) {
```
sonchang: we should change to hostMap instead of hostId

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
```diff
@@ -154,12 +173,12 @@ protected Agent getLoadBalancerAgent(LoadBalancer loadBalancer, long hostId) {
         return instances;
     }
 
-    private List<Long> populateHosts(LoadBalancer loadBalancer, Long... hostIds) {
-        List<Long> hosts = new ArrayList<Long>();
-        if (hostIds.length == 0) {
-            hosts = lbInstanceDao.getLoadBalancerHosts(loadBalancer.getId());
+    private List<LoadBalancerHostMap> populateHostMaps(LoadBalancer loadBalancer, LoadBalancerHostMap... hostMaps) {
+        List<LoadBalancerHostMap> hosts = new ArrayList<>();
```
sonchang: we should probably rename 'hosts'.  Not sure what's a good name since we already have 'hostMaps'.
Actually, looking at this again, the purpose of this function seems a little strange.  It's not quite lazy loading, but has some similarities from line 179

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
```diff
@@ -216,4 +235,30 @@ public IpAddress getLoadBalancerInstanceIp(Instance lbInstance) {
...
+    @Override
+    public LoadBalancerHostMap getLoadBalancerHostMapForInstance(Instance lbInstance) {
+        if (!isLbInstance(lbInstance)) {
+            return null;
+        }
+        Agent agent = objectManager.loadResource(Agent.class, lbInstance.getAgentId());
+
+        // get lb id from agent uri
+        String uri = agent.getUri();
+        String[] result = uri.split("hostMapId");
```
sonchang: hmm.  why not split by "hostMapId="
?

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
```diff
@@ -208,8 +212,13 @@ private void updatePublicPorts(ProcessState state, Map<Long, List<? extends Inst
             List<? extends Instance> lbInstances = lbInstancesMap.get(lbId);
             // surround by lock
             for (final Instance lbInstance : lbInstances) {
-                final boolean hasActiveInstances = (targetDao.getLoadBalancerActiveInstanceTargets(lb.getId()).size() + targetDao
+                boolean lbHasActiveTargets = (targetDao.getLoadBalancerActiveInstanceTargets(lb.getId()).size() + targetDao
                         .getLoadBalancerActiveIpTargets(lb.getId()).size()) > 0;
+                List<String> ports = DataUtils.getFieldList(lbInstance.getData(), InstanceConstants.FIELD_PORTS,
+                        String.class);
+                boolean lbInstanceHasPortsSet = (ports != null && !ports.isEmpty());
+                final boolean hasActiveInstances = lbHasActiveTargets || lbInstanceHasPortsSet;
```
sonchang: I'm not quite sure I understand this.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
alena1108: You are right, before hostMaps came in, this function made sense because it was translating hostids into hostmaps. Now we might not need it.

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
alena1108: You are right :) will fix

code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/process/LoadBalancerUpdateConfig.java
alena1108: Before this refactoring, lb ports binding to host were dynamic; and we were opening/closing them only when LoadBalancer had active (Running) targets. Now to make it work with the new allocator, initial set of listener ports is set on the lb system instance itself, and we have to open the ports in this case on the host even when there are no active targets exist in the load balancer. Therefore to determine if the ports have to be opened, we check 2 conditions:

* lbHasActiveTargets
* lbInstanceHasPortsSet

If either one of them true, we open the ports

sonchang: thx!

PR #392, opened 2015-05-01
Service discovery bug fixes

alena1108:
1) Missing state transitions are added; that would fix:

https://github.com/rancherio/rancher/issues/678
https://github.com/rancherio/rancher/issues/676

2) Environment name is unique within the account, not globally

https://github.com/rancherio/rancher/issues/659

3) Disallow to remove the lb that was created as a part of LB service. This type of lb can be removed only when its service is removed

4) Don't try to scale down lb service as a part of cleanup, if the lb is gone (the cleanup is called as a part of service.remove process)

https://github.com/rancherio/rancher/issues/699

5) LB service: handle the case when consumed svc link is created before lb
    service was fully created

https://github.com/rancherio/rancher/issues/702

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/EnvironmentCreateValidationFilter.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/EnvironmentCreateValidationFilter.java
@@ -1,18 +1,23 @@
 package io.cattle.platform.servicediscovery.api.filter;
 
-import static io.cattle.platform.core.model.tables.EnvironmentTable.ENVIRONMENT;
 import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.iaas.api.filter.common.AbstractDefaultResourceManagerFilter;
-import io.cattle.platform.object.ObjectManager;
+import io.cattle.platform.object.meta.ObjectMetaDataManager;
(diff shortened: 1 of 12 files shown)


alena1108:
@sonchang @cjellick could you please review?

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/EnvironmentCreateValidationFilter.java
```diff
@@ -29,7 +29,7 @@ public Object create(String type, ApiRequest request, ResourceManager next) {
         }
 
         Environment existingEnv = objectManager.findOne(Environment.class, ENVIRONMENT.NAME, env.getName(),
```
ibuildthecloud: When looking up resource in the API layer it is far better to use `io.github.ibuildthecloud.gdapi.request.resource.ResourceManager` because that does proper auth checking

ibuildthecloud: You get a resource manager by injecting `io.github.ibuildthecloud.gdapi.request.resource.ResourceManagerLocator` and then use list()

alena1108:
@ibuildthecloud

cjellick:
LGTM

PR #396, opened 2015-05-04
Allow for inactive port

sonchang:


sonchang:
@ibuildthecloud @alena1108

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -333,7 +333,7 @@ public void setMapDao(GenericMapDao mapDao) {
                         .and(INSTANCE.REMOVED.isNull())
                         .and(INSTANCE.STATE.in(InstanceConstants.STATE_STARTING, InstanceConstants.STATE_RESTARTING, InstanceConstants.STATE_RUNNING))
                         .and(INSTANCE_HOST_MAP.REMOVED.isNull())
-                        .and(PORT.STATE.in(CommonStatesConstants.ACTIVE, CommonStatesConstants.ACTIVATING))
+                        .and(PORT.STATE.notIn(CommonStatesConstants.REMOVED, CommonStatesConstants.REMOVING, CommonStatesConstants.PURGED))
```
alena1108: Should we also include Purging state?

sonchang: Shoot.  I'll add purging in here.  thanks for catching this.  so many states for port :-P

alena1108:
LGTM, but there is a bug in lb_service test that is fixed as a part of pending https://github.com/rancherio/cattle/pull/391. By now you can just mark that test as skipped; and I will re-enable it once my stuff gets merged.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -331,9 +331,10 @@ public void setMapDao(GenericMapDao mapDao) {
                         .on(INSTANCE_HOST_MAP.INSTANCE_ID.eq(INSTANCE.ID))
                     .where(INSTANCE_HOST_MAP.HOST_ID.eq(hostId)
                         .and(INSTANCE.REMOVED.isNull())
+                        .and(INSTANCE.ID.ne(instanceId))
                         .and(INSTANCE.STATE.in(InstanceConstants.STATE_STARTING, InstanceConstants.STATE_RESTARTING, InstanceConstants.STATE_RUNNING))
                         .and(INSTANCE_HOST_MAP.REMOVED.isNull())
-                        .and(PORT.STATE.in(CommonStatesConstants.ACTIVE, CommonStatesConstants.ACTIVATING))
+                        .and(PORT.STATE.notIn(CommonStatesConstants.REMOVED, CommonStatesConstants.REMOVING, CommonStatesConstants.PURGED, CommonStatesConstants.PURGING))
```
ibuildthecloud: Instead of listing all these states I think PORT.REMOVED.isNull() would be sufficient

sonchang: What about ones in the transitioning state like REMOVING?

ibuildthecloud: I'm not totally sure about that one.  At what point do we set removed=NOW()?  Is it at the start or end of the remove process.  I think.... it's at the start.

PR #397, opened 2015-05-04
Service Healthcheck

alena1108:


alena1108:
@ibuildthecloud Darren, this PR is for a healthcheck config generation

code/iaas/agent-instance/src/main/resources/META-INF/cattle/system-services/agent-instance-defaults.properties
```diff
@@ -50,3 +51,11 @@ serviceconsumemap.create.agentInstanceProvider.dnsService.apply=hosts
 serviceconsumemap.create.agentInstanceProvider.dnsService.increment=hosts
 serviceexposemap.create.agentInstanceProvider.dnsService.apply=hosts
 serviceexposemap.create.agentInstanceProvider.dnsService.increment=hosts
+
+nic.activate.agentInstanceProvider.dhcpService.apply=healthcheck
```
ibuildthecloud: why on dhcpservice?  maybe we should add another service type for this.

alena1108: wanted to avoid adding the extra services as I remember you were saying that you want to refactor network service code at some point...let me know what you think, and I can add an extra service if needed.

PR #398, opened 2015-05-05
Use labels in native container code

cjellick:
Use labels instead of environment variables to detect rancher uuid and rancher networking.

***Dont merge. This requires some coordination because it requires that we upgrade to 1.6.***
I'll coordinate with @ibuildthecloud and @cloudnautique on when it can be merged.


sonchang:
LGTM.  Just curious.  Is the fallback to checking environment variables temporary?

cjellick:
@sonchang yes. temporary backwards compatibility

cjellick:
updated to io.rancher.container.*

PR #400, opened 2015-05-06
snapshot: idempotency support

yasker:
Update volmgr to v0.1.5

### DIFF
--- a/resources/content/schema/base/snapshot.json
+++ b/resources/content/schema/base/snapshot.json
@@ -4,10 +4,6 @@
             "required": true,
             "nullable": false
         },
-        "managedSnapshotUUID": {
-            "type": "string",
-            "nullable": true
-        },
         "managedVolumeUUID": {
             "type": "string",
             "nullable": true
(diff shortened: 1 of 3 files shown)


yasker:
Updated with rebase and bump up rancher/agent version to v0.6.5

alena1108:
LGTM

resources/content/schema/base/snapshot.json
```diff
@@ -4,10 +4,6 @@
             "required": true,
             "nullable": false
         },
-        "managedSnapshotUUID": {
```
ibuildthecloud: Why did this go away?  Was it never really used by cattle?  Only python-agent?

yasker: Because it would be always identical to "snapshot.uuid" now. I've add ability to specific snapshot_uuid when creating snapshot using volmgr.

ibuildthecloud: Awesome

yasker: And now snapshot.uuid and "managedVolumeUUID" in snapshot are needed for restore volumes.

PR #408, opened 2015-05-07
DNS: include only Running resolve targets

alena1108:
So only reachable instance would appear as resolve targets in dns configuration

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
@@ -1,7 +1,7 @@
 package io.cattle.platform.configitem.context.dao.impl;
 
-import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
 import static io.cattle.platform.core.model.tables.InstanceLinkTable.INSTANCE_LINK;
+import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
 import static io.cattle.platform.core.model.tables.IpAddressNicMapTable.IP_ADDRESS_NIC_MAP;
 import static io.cattle.platform.core.model.tables.IpAddressTable.IP_ADDRESS;
 import static io.cattle.platform.core.model.tables.NicTable.NIC;
@@ -10,6 +10,7 @@ import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.SERVICE
 import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
 import io.cattle.platform.configitem.context.dao.DnsInfoDao;
 import io.cattle.platform.configitem.context.data.DnsEntryData;
+import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.constants.IpAddressConstants;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.model.InstanceLink;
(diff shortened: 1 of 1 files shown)


alena1108:
@ibuildthecloud

alena1108:
https://github.com/rancherio/rancher/issues/737

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
```diff
@@ -89,7 +93,9 @@ protected DnsEntryData map(List<Object> input) {
                         .and(targetNicIpTable.REMOVED.isNull())
                         .and(clientNic.REMOVED.isNull())
                         .and(targetNic.REMOVED.isNull())
-                        .and(instanceLink.REMOVED.isNull()))
+                        .and(instanceLink.REMOVED.isNull())
+                        .and(targetInstance.STATE.in(InstanceConstants.STATE_RUNNING,
```
ibuildthecloud: I don't think RESTARTING is necessary

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
```diff
@@ -160,7 +169,9 @@ protected DnsEntryData map(List<Object> input) {
                         .and(serviceConsumeMap.REMOVED.isNull())
                         .and(clientServiceExposeMap.REMOVED.isNull())
                         .and(targetServiceExposeMap.REMOVED.isNull())
-                        .and(targetService.REMOVED.isNull()))
+                        .and(targetService.REMOVED.isNull())
+                        .and(targetInstance.STATE.in(InstanceConstants.STATE_RUNNING,
+                                InstanceConstants.STATE_STARTING, InstanceConstants.STATE_RESTARTING)))
```
ibuildthecloud: I don't think RESTARTING is necessary

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
```diff
@@ -226,7 +240,9 @@ protected DnsEntryData map(List<Object> input) {
                         .and(clientNic.REMOVED.isNull())
                         .and(targetNic.REMOVED.isNull())
                         .and(clientServiceExposeMap.REMOVED.isNull())
-                        .and(targetServiceExposeMap.REMOVED.isNull()))
+                        .and(targetServiceExposeMap.REMOVED.isNull())
+                        .and(targetInstance.STATE.in(InstanceConstants.STATE_RUNNING,
+                                InstanceConstants.STATE_STARTING, InstanceConstants.STATE_RESTARTING)))
```
ibuildthecloud: Would you believe it... I don't think RESTARTING is necessary

alena1108: I believe you, can remove. I just remember that for the standalone LB I used to include only Starting/Running containers as haproxy config, and for the restart case, the containers weren't registered just because at the moment of config generation they were at Restarting state. But may be for this use case the trigger point for config update - nic.activate - would ensure that instance is in Starting state. Let me know what you think

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
alena1108: Just tested; works well w/o Restarting state, so removing

cjellick:
LGTM!

PR #409, opened 2015-05-07
Fixed seting ports on lb instance

alena1108:
for the lb instance the public port and source port are always = to the source port specified on the lb listener. The target port of the lb listener is the one that is set for the instance lb is balancing


alena1108:
https://github.com/rancherio/rancher/issues/742

sonchang:
LGTM

cjellick:
LGTM

PR #410, opened 2015-05-07
Choose Default Credential

hibooboo2:


code/implementation/docker/common/src/main/java/io/cattle/platform/docker/constants/DockerStoragePoolConstants.java
```diff
@@ -0,0 +1,8 @@
+package io.cattle.platform.docker.constants;
+
+public class DockerStoragePoolConstants {
+
+    public static final String DOCKER_KIND = "docker";
+    public static final String DOCKER_FORMAT = "docker";
```
alena1108: could we just default one parameter to another? or is there a chance that docker_format and docker_kind would be different?

code/implementation/docker/storage/src/main/java/io/cattle/platform/docker/storage/DockerImageCredentialLookup.java
```diff
@@ -0,0 +1,49 @@
...
+        Credential toReturn = null;
+        for(Object cred: credentials){
+            if (!(cred instanceof Credential)){
+                continue;
+            }
+            Credential credential = (Credential) cred;
+            if (!credential.getKind().equalsIgnoreCase("registryCredential")){
+                continue;
+            }
+            if (credential.getRegistryId().equals(registryId)){
```
alena1108: is there a chance registryId can be null at this point (when the code never hit line 25)? If so, we should probably handle it.

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/constants/DockerStoragePoolConstants.java
hibooboo2: This was a refactor of fields used before. Was told by darren that he had them as two for a reason.

code/implementation/docker/storage/src/main/java/io/cattle/platform/docker/storage/DockerImageCredentialLookup.java
hibooboo2: You are write. Addressed it.

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/constants/DockerStoragePoolConstants.java
alena1108: ok, sure, just wanted to check

alena1108:
LGTM

code/iaas/storage-service/src/main/java/io/cattle/platform/storage/api/filter/ExternalTemplateInstanceFilter.java
```diff
@@ -55,14 +60,34 @@ public Object create(String type, ApiRequest request, ResourceManager next) {
...
             Image image = storageService.registerRemoteImage(uuid);
             if (image == null) {
                 return null;
             }
-            if (registryCredentialId != null) {
-                objectManager.setFields(image, IMAGE.REGISTRY_CREDENTIAL_ID, registryCredentialId);
+            Long id = instance.getRegistryCredentialId();
+            if (id == null && image.getFormat().equalsIgnoreCase(DockerStoragePoolConstants.DOCKER_FORMAT)) {
+                String type = schemaFactory.getSchemaName(StoragePool.class);
+                List<?> storagePools = locator.getResourceManagerByType(type).list(type, new HashMap<>(), new ListOptions());
```
sonchang: This question is a bit for my own understanding.  Was there a particular reason we chose to retrieve a list of storagePools and credentials using the ResourceManager instead of ObjectManager?

I haven't dug into the code deeper, but it looks like the ResourceManager is closer to the public API (hence, the ListOptions for pagination, etc...) so it could potentially be performing more checks/filtering than ObjectManager which probably provides a more low level API.

code/implementation/docker/storage/src/main/java/io/cattle/platform/docker/storage/DockerImageCredentialLookup.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+public class DockerImageCredentialLookup implements ImageCredentialLookup {
+
+    private static final String REGISTRY = "registry";
+    private static final String REGISTRY_CREDENTIAL = "registryCredential";
+
+    @Override
+    public Credential getDefaultCredential(String uuid, List<?> storagePools, List<?> credentials) {
+        String serverAddress = DockerImage.parse(uuid).getServer();
```
sonchang: DockerImage.parse() seems like it could return a null in various cases, so we might want to check for that unless you're absolutely positive this can't happen ;-)

sonchang:
Besides the questions/comments already mentioned, LGTM

code/implementation/docker/storage/src/main/java/io/cattle/platform/docker/storage/DockerImageCredentialLookup.java
hibooboo2: You are right this method was written under the assumption that uuid is always valid. as per where the code is called i will add a null check.

code/iaas/storage-service/src/main/java/io/cattle/platform/storage/api/filter/ExternalTemplateInstanceFilter.java
hibooboo2: Objectmanager will not give me the list that i need. While ResourceManager will.

sonchang: How did you try fetching it with ObjectManager?  Or what was the difference?

ibuildthecloud: @son ResourceManager does authorization checks based on the API logic.  ObjectManager is more raw database access.  When in the API tier and you need to look up objects it's safe to use ResourceManager because of the authorization checks.  Imagine that one day we have more complex authorization like IAM.

Also, ResourceManager mimics API requests of GET, POST, PUT, etc.  So by calling `list()` you are getting the equivalent of `/v1/storagepools` which is already scoped to objects the user is allow to access.

hibooboo2: @sonchang sorry for my incomplete reply I will work on providing more thorough responses in the future.

sonchang: Cool.  That answers my original question.  I was concerned we were accessing a higher level API (one that does pagination, etc...) but from a much lower level entry point.  But if that's the only entrypoint we have to get the appropriate auth checks, I'm good with it.

sonchang:
LGTM

PR #412, opened 2015-05-07
Fix Bug with Turning Auth on.

hibooboo2:
Only change it if the rancher id is admin.
Use current account for projects with no members.


alena1108:
LGTM

sonchang:
LGTM

PR #413, opened 2015-05-08
Service Discovery: set label on service containers

alena1108:
As a part of this PR:

1) "labels" field (Map<String, String>) is introduced on the "container" object. So labels can be set through this field on per standalone container basis, as well as on the services' containers as their launchConfig is of type "container". All affinity rules user defines in the compose, get translated into lables on launchConfig. Expected format: "affinity:container_label:io.rancher.service==name of the service you want to have an affinity with" - for affinity rules

2) When ServiceDiscovery manager builds launch data for container, it copies all labels defined on services launch config, and set them per each service container ("labels" field on the container)+ adds one more extra label with key="io.rancher.service" and value="name of the service the container belongs to"

@ibuildthecloud Darren, let me know if the above is what you expected, or if anything is missing. @will-chan suggested that we use service name, not uuid for defining the service container. The reason is - at the moment compose file is defined, user doesn't know service's uuid, only name; so we stick to name so we don't have to do any name to uuid translation

@sonchang Son, let me know if the format for labeling the service container (key="io.rancher.service" and value=<name of the service the container belongs to>) is what the allocator expects.

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
@@ -44,4 +44,6 @@ public class ServiceDiscoveryConstants {
     public static final String LINK_RANCHER_COMPOSE_CONFIG = "rancherComposeConfig";
     public static final String LINK_COMPOSE_CONFIG = "composeConfig";
 
+    public static final String LABEL_SERVICE_NAME = "io.rancher.service.name";
+
 }
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
@@ -20,7 +20,6 @@ import io.cattle.platform.servicediscovery.service.ServiceDiscoveryService;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Callable;
 
 import javax.inject.Inject;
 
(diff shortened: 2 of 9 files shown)


alena1108:
I separated putting container.labels field to a separate PR: https://github.com/rancherio/cattle/pull/415

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -78,7 +78,7 @@ public Instance createServiceInstance(Map<Object, Object> properties, Service se
             @Override
             public Object call() throws Exception {
                 objectProcessManager.scheduleStandardProcess(StandardProcess.CREATE, instance, null);
-                return null;
+                return instance;
```
sonchang: Since we don't do anything with the return value, I wonder should we change this to DeferredUtils.nest(new Runnable() { ...}) instead

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
```diff
@@ -16,6 +16,7 @@
     public static final String FIELD_LAUNCH_CONFIG = "launchConfig";
     public static final String FIELD_LOAD_BALANCER_CONFIG = "loadBalancerConfig";
     public static final String FIELD_ENVIRIONMENT_ID = "environmentId";
+    public static final String FIELD_CONTAINER_LABEL = "io.rancher.service";
```
ibuildthecloud: This should probably be `io.rancher.service.name`

alena1108: sure, will fix

ibuildthecloud: Other small thing. FIELD_ is used for field names

alena1108: will rename to SERVICE_CONTAINER_LABEL

PR #414, opened 2015-05-08
Cache Github AccessToken on as account data.

hibooboo2:


cjellick:
@hibooboo2  What do you do with the token once it's saved on the account? Where do you read it?

hibooboo2:
For use with the saas and to allow us to talk to github on behalf of the user in the future.

cjellick:
@hibooboo2 I would say that `access_token` should probably be `accessToken` then, to follow proper naming conventions.

FWIW, you also inconsistently use camelCase and underscore_based naming in the the jsonData object. Is there an explicit reason for that?

cjellick:
And is this a field in the account schema? I don't think you should be setting a field without adding it as a field in the schema.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubTokenHandler.java
```diff
@@ -103,6 +105,9 @@ public Token getGithubAccessToken(ApiRequest request) throws IOException {
                     userAccountInfo.getAccountName()));
         }
         account = objectManager.reload(account);
```
ibuildthecloud: remove this reload

ibuildthecloud: Make L92 less scary

hibooboo2:
It is not a field in The schema right now. I could add it, but it won't be used by any of the consumers of the api.

PR #417, opened 2015-05-09
Support sync and import

cjellick:
Sync up all containers in ping monitor.


code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -36,13 +52,30 @@
...
+    @Inject
     PingInstancesMonitorDao monitorDao;
+    @Inject
     ObjectManager objectManager;
+    @Inject
     ObjectProcessManager processManager;
-    LoadingCache<Long, Map<String, String>> instanceCache = CacheBuilder.newBuilder().expireAfterWrite(CACHE_TIME.get(), TimeUnit.MILLISECONDS).build(
-            new CacheLoader<Long, Map<String, String>>() {
+
+    LoadingCache<Long, Map<String, Instance>> instanceCache = CacheBuilder.newBuilder().expireAfterWrite(CACHE_TIME.get(), TimeUnit.MILLISECONDS)
```
cjellick: @ibuildthecloud Note that I'm now putting instances in this cache. Think they are too big memory-wise? If so, I could change to a custom pojo that has just the fields I need (would be four or five fields).

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -1,7 +1,17 @@
 package io.cattle.platform.ha.monitor.impl;
 
+import static io.cattle.platform.core.constants.ContainerEventConstants.*;
```
cjellick: So much changed in this class, you might be better off _not_ looking at the diff and just looking at the updated version on its own.

cjellick:
Added a second commit to fix a bug where docker restarts weren't being properly respected.

cjellick:
For the record, that bug is here:
https://github.com/rancherio/rancher/issues/745

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/dao/impl/PingInstancesMonitorDaoImpl.java
```diff
@@ -4,18 +4,21 @@
...
 import java.util.Map;
 
 public class PingInstancesMonitorDaoImpl extends AbstractJooqDao implements PingInstancesMonitorDao {
 
     @Override
-    public Map<String, String> getInstances(long agentId) {
-        Map<String, String> result = create()
-                .select(INSTANCE.UUID, INSTANCE.EXTERNAL_ID)
+    public Map<String, Instance> getInstances(long agentId) {
+        List<Instance> instances = create()
```
alena1108: Craig, here we assume that its gonna be just one agent per host (as we do join based on hostId). This wouldn't work once we start supporting multiple networks per account. We might think about making one extra join on vnetId - we can implement it later, just something to note for the future.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -73,36 +113,173 @@ public void computeInstanceActivateReply(Event event) {
...
-                    if (checkOnly) {
-                        return true;
-                    } else {
-                        // If this known instance was not reported, schedule a potential restart.
-                        restart(knownInstance.getKey());
-                    }
+    /*
+     * If checkOnly is true, will raise a ContainersOutOfSync exception, indicating this should be reran with checkOnly set to false.
+     */
+    void syncContainers(Map<String, Instance> knownInstances, ReportedInstances reportedInstances, long agentAccountId, long agentId, long hostId,
```
alena1108: Just curious - do we apply any lock when this operation is called? I see that the changes is refactoring of existing code, so my questions applies to original implementation. In this PR I see that we lock on per individual instance basis when we operate on it in ContainerEventCreate.java, but not when we determine if the objects are out of sync

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -73,36 +113,173 @@ public void computeInstanceActivateReply(Event event) {
...
+    }
+
+    void determineSyncActions(Map<String, Instance> knownInstances, ReportedInstances reportedInstances, Map<String, ReportedInstance> needsSynced,
+            Map<String, String> syncActions, Set<String> needsHaRestart, boolean checkOnly) {
+        Map<String, Instance> inRancher = new HashMap<String, Instance>(knownInstances);
+        Map<String, ReportedInstance> onHost = new HashMap<String, ReportedInstance>(reportedInstances.byExternalId);
+        for (Map.Entry<String, ReportedInstance> reported : reportedInstances.byUuid.entrySet()) {
+            Instance ki = knownInstances.get(reported.getKey());
+            if (ki != null) {
+                onHost.remove(reported.getValue().getExternalId());
```
alena1108: Minor, but lines 152-155 seem to be identical to 169-172. May be it would make sense to extract it to some helper method.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -73,36 +113,173 @@ public void computeInstanceActivateReply(Event event) {
...
+                // This is the one place we can't use addSyncAction, since we don't have (and can't construct) a ReportedInstance.
+                if (checkOnly) {
+                    throw new ContainersOutOfSync();
+                }
+                needsHaRestart.add(ki.getUuid());
+            } else {
+                ReportedInstance ri = new ReportedInstance();
+                ri.setExternalId(ki.getExternalId());
+                ri.setUuid(ki.getUuid());
+                addSyncAction(needsSynced, syncActions, ri, EVENT_STOP, checkOnly);
```
alena1108: Craig, just curious - we always stop user containers if they are running in cattle, but stopped on the backend, disregarding their Restart Policy?

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
alena1108: Update - I guess we should be fine if the syncContainers is called from within a single threaded task (PingInstanceMonitorImpl for example), so just wanted to make sure that thats the case.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -197,10 +203,8 @@ void setNetwork(Map<String, Object> inspect, Instance instance) {
         }
     }
 
-    @SuppressWarnings("unchecked")
-    void setImage(Map<String, Object> inspect, Instance instance) {
-        Map<String, Object> config = (Map<String, Object>)inspect.get(INSPECT_CONFIG);
-        String imageUuid = (String)config.get(INSPECT_IMAGE);
+    void setImage(ContainerEvent event, Instance instance) {
+        String imageUuid = event.getExternalFrom();
```
sonchang: Can you explain this change?

cjellick: docker events include the image that was used for the container and this value is identical to what is in the inspect. We don't have the inspect for a container why doing the the sync via ping, but we do have it, so it is more straightforward to set it on the event as opposed to trying to construct a "fake" inspect. Does that make sense @sonchang ?

sonchang: Thanks!  makes sense

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: The restartPolicy that is exposed to the user in rancher is passed through to docker, so we don't do anything with it.
There is another field that we don't expose to users called `instanceTriggeredStopAction` and this is what controls the restart of system containers. I could have keyed off of that field instead, but went with keying off of system container. Maybe the right thing to do?

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: Sure. nice catch.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/dao/impl/PingInstancesMonitorDaoImpl.java
cjellick: @alena1108 there's the network agent and then there's the rancher agent. This is keying off of the rancher agent.

I'm under the impression that there will be only rancher agent per host, which is what handles the pinging.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -56,12 +90,18 @@ public void pingReply(Ping ping) {
...
+        Host host = hosts.get(reportedInstances.hostUuid);
 
-        if (handleUnreportedKnownInstances(agentId, knownInstances, reportedInstances, true)) {
+        try {
+            syncContainers(knownInstances, reportedInstances, agent.getAccountId(), agent.getId(), host.getId(), true);
+        } catch (ContainersOutOfSync e) {
             knownInstances = load(agentId);
             instanceCache.put(agentId, knownInstances);
-            handleUnreportedKnownInstances(agentId, knownInstances, reportedInstances, false);
+            syncContainers(knownInstances, reportedInstances, agent.getAccountId(), agent.getId(), host.getId(), false);
```
sonchang: This question is more for my understanding, but ultimately, the state of the host or rancher can change after the ping data or after we refresh the knownInstances in the cache.  So the syncActions that we generate can potentially be out of date by the time they're executed.  I noticed in the python-agent changes, we handle this already.  Is this correct?

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/dao/impl/PingInstancesMonitorDaoImpl.java
alena1108: You are right. I was thinking about the network agent, dropping this question :)

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
alena1108: I think keying off instanceTriggeredStopAction for system containers would be the better approach. Craig, the question about restartPolicy - basically if this parameter is defined, we delegate the container restart - if existed - to docker, and then we will sync up its state later. Let me know if my assumption is correct.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: Good question. Yes, theoretically, the syncActions can be out of date by the time they're executed. We do our best to avoid this, but there is a small chance of it happening still.

To mitigate, we:
 - don't do any syncing for an instance if it is in a transitioning state
 - we do additional state checks in ContainerEventCreate as well.
- we allow processes scheduled via ContainerEventCreate to be rejected if the container turns out to be in a different state (by catching and swallowing the ProcessCancelException which occurs when you schedule a process and the resource is not in a valid start state for that process)
 - We do as much as possible via ContainerEventCreate since those are no-op events that won't affect the actual container state. The exception to this is system containers, which need ha restarts, but rancher is the system of record for them, so what we decide to do in the check should be accurate.

So, even with all that, there is small chance we could miss schedule something. In this case, Darren said that his philosophy is that that's ok because it's a small chance and the containers should eventually sync back up (possibly having to wait up to 15 minutes for the knownInstances cache to be invalidated).

code/iaas/ha/src/test/java/io/cattle/platform/ha/monitor/impl/PingInstanceMonitorImplTest.java
```diff
@@ -0,0 +1,262 @@
...
+        needsHaRestart = new HashSet<String>();
+        reportedInstances = new ReportedInstances();
+    }
+
+    String setupSync(String key, boolean byExternalId, boolean sysCon, String knownInstanceState, String reportedInstanceState) {
+        key = fullKey(key, byExternalId, sysCon);
+        String uuid = key + "-uuid";
+        String externalId = key + "-extid";
+
+        String instaceEid = byExternalId ? externalId : null;
```
sonchang: I know it's a sickness.  but typos

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: Your assumption is correct regarding restartPolicy @alena.
And I see your point about `instanceTriggeredStopAction`. I'll switch to keying off of that.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
sonchang: Sounds good to me!

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: ok. wont change.

code/iaas/ha/src/test/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImplTest.java
```diff
@@ -0,0 +1,274 @@
...
+        assertSyncAction(externalIdCC, EVENT_INSTANCE_FORCE_STOP);
+        assertSyncAction(externalIdDD, EVENT_INSTANCE_FORCE_STOP);
+        
+        assertSyncAction(externalIdEE, EVENT_CREATE);
+        assertSyncAction(externalIdFF, EVENT_CREATE);
+        asserDoNothing(externalIdGG, uuidGG);
+        asserDoNothing(externalIdHH, uuidHH);
+    }
+
+    void asserDoNothing(String externalId, String uuid) {
```
sonchang: sorry.  another typos

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
alena1108: Craig, did you plan to switch to using this parameter, in this PR? or the subsequent one. I still see that we use .getSystemContainer to identify whether the container needs to be restarted

cjellick: @alena1108 I added `hasInstanceTriggeredStopConfigured()` to the check. So, we'll do the HA_RESTART either if it is a system container or it has that property set (to something other than stop, which they all have by default).

I did this because we never want ContainEvents to manage system containers. This is because ContainerEvents assume that the host is the source of truth for the intended state of the container, but for system containers, Rancher is _always_ the source of truth.

code/iaas/ha/src/test/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImplTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+    @Test
+    public void testDetermineSyncActions() {
+        // container running in rancher, stopped on host
+        String externalIdA = setupSync("rancher-running-host-stopped", false, false, STATE_RUNNING, STATE_STOPPED);
+        String externalIdC = setupSync("rancher-running-host-stopped", true, false, STATE_RUNNING, STATE_STOPPED);
+        String externalIdB = setupSync("rancher-running-host-stopped", false, true, STATE_RUNNING, STATE_STOPPED);
+        String externalIdC1 = setupSync("rancher-running-host-stopped", true, true, STATE_RUNNING, STATE_STOPPED);
+        String externalIdC2 = setupSync("rancher-running-host-stopped-triggered", false, false, "restart", STATE_RUNNING, STATE_STOPPED);
+        String externalIdC3 = setupSync("rancher-running-host-stopped-triggered", true, false, "restart", STATE_RUNNING, STATE_STOPPED);
```
cjellick: @alena1108 FYI, these two "rancher-running-host-stopped-triggered" cases are the tests i added to test the instanceTriggeredStop + not-a-system-container logic change.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -56,12 +90,18 @@ public void pingReply(Ping ping) {
         }
 
         long agentId = Long.parseLong(ping.getResourceId());
-        Map<String, String> knownInstances = instanceCache.getUnchecked(agentId);
+        Map<String, Instance> knownInstances = instanceCache.getUnchecked(agentId);
+
+        Agent agent = objectManager.loadResource(Agent.class, agentId);
+        Map<String, Host> hosts = agentDao.getHosts(agent.getId());
```
ibuildthecloud: I would deal with the situation when an agent has no hosts.  Also cache the response so that we don't query for each ping

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -73,36 +113,182 @@ public void computeInstanceActivateReply(Event event) {
...
+            restart(uuid);
+        }
+
+        for (Map.Entry<String, ReportedInstance> syncEntry : needsSynced.entrySet()) {
+            ReportedInstance ri = syncEntry.getValue();
+            String syncAction = syncActions.get(syncEntry.getKey());
+            if (EVENT_INSTANCE_FORCE_STOP.equals(syncAction)) {
+                forceStop(ri.getExternalId(), agentId);
+            } else if (HA_RESTART.equals(syncAction)) {
+                String uuid = ri.getInstance() != null ? ri.getInstance().getUuid() : ri.getUuid();
```
ibuildthecloud: let's just assume ri.getInstance() is never null (or throw illegalstateexception)

PR #422, opened 2015-05-11
Add userPreferences Resource.

hibooboo2:
This resource is to allow accounts to have a list preferences with name and value where name is unique per account.


sonchang:
LGTM

resources/content/schema/base/userPreference.json
```diff
@@ -0,0 +1,7 @@
+{
+  "resourceFields": {
+    "name": {
```
alena1108: is the name supposed to be unique across accounts? or on per account basis?

hibooboo2: per account.

alena1108:
LGTM

PR #430, opened 2015-05-13
Service Discovery Replication controller

alena1108:
This PR adds:

* refactoring to support replication controller model. ReplicationController is meant to reconcile the service to its scale by scaling up/down service instances. The reconcile can be triggered either by healthcheck update, or by service.activate called by an end user.

* sidekick services support

* volumesFrom service implementation

This PR modifies:

* The way we set propagate labels to LB service instance. We used to set label defined in the service, to the LB; now we do it on per host map basis as each instance within the same lb service can get different affinity constraint labels.


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/AffinityConstraintDefinition.java
```diff
@@ -7,8 +7,8 @@
         EQ("==", "{eq}"),
         NE("!=", "{ne}");
 
-        String envSymbol;
-        String labelSymbol;
+        public String envSymbol;
+        public String labelSymbol;
```
ibuildthecloud: Why's this?  Can we just put a getter if you need this value?

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
```diff
@@ -49,6 +60,30 @@ public Object create(String type, ApiRequest request, ResourceManager next) {
                     "name");
         }
 
+        // 2. if a part of sidekick, scale should match other sidekick participants scale
+        Map<String, Object> launchConfig = DataAccessor
```
ibuildthecloud: Do we really need to add validate this.  From a user perspective this will be annoying.  If the value must be X, can we just make it X?  For this scale.  We should try to keep these value in sync on scale up/scale down.  But the backend should assume they are.  So if they get out of sync because of some race condition, things shouldn't fail.  On the backed just take the max value.  (Maybe you do this already, haven't gotten that far yet in the code)

alena1108: Thats actually the thing I wanted to talk to you about. if we don't keep the scale same, its hard to validate what is the desired value (because what if the user scaled down)? and I cant update the scale inside of the process - I ran into way too many race conditions especially with idempotent checks on. Taking the max value would def make things easier.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -154,4 +103,52 @@ public void updateEnvironmentName(Environment env) {
...
+            /*
+             * Find all sidekick services. Just look at sidekick, not volumes-from or other things
+             */
+            List<? extends Service> services = sdService.listEnvironmentServices(initialService.getEnvironmentId());
+            for (Service svc : services) {
+                if (svc.getId().equals(initialService.getId())) {
+                    continue;
+                }
+
+                if (svc.getRemoved() == null) {
```
ibuildthecloud: Anytime we have `getRemoved() == null` check you have to consider will the code just get slower and slower over time as the user adds/deletes things.  Sometimes the nature of the code that there will only be a couple, but in this situation it will just keep growing.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/manager/ServiceResourceManager.java
```diff
@@ -0,0 +1,54 @@
...
+        return new String[0];
+    }
+
+    @Override
+    public Class<?>[] getTypeClasses() {
+        return new Class<?>[] { Service.class };
+    }
+
+    @Override
+    protected Object updateInternal(String type, String id, Object obj, ApiRequest request) {
```
alena1108: @ibuildthecloud if we decide to assume that the desired scale is the max scale, then we won't need this method - it was to update the scale for all the participants on a single service scaleup

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/AffinityConstraintDefinition.java
alena1108: Getter would do too.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -154,4 +103,52 @@ public void updateEnvironmentName(Environment env) {
...
+                    Map<String, String> svcLabels = sdService.getServiceLabels(svc);
+                    String svcSidekickLabel = svcLabels.get(
+                            ServiceDiscoveryConstants.LABEL_SERVICE_SIDEKICK) == null ? null
+                            : svcLabels.get(ServiceDiscoveryConstants.LABEL_SERVICE_SIDEKICK);
+                    if (svcSidekickLabel != null && svcSidekickLabel.equalsIgnoreCase(initialSvcSidekickLabel)) {
```
ibuildthecloud: nit: You can avoid the null check if you reverse the condition

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
alena1108: Agree. I will fix the dao method to query for non-removed things

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -154,4 +103,52 @@ public void updateEnvironmentName(Environment env) {
...
+
+    @Override
+    public List<Service> collectSidekickServices(Service initialService, Map<String, String> initialSvcLabels) {
+        List<Service> servicesToCollect = new ArrayList<>();
+        if (initialSvcLabels == null) {
+            initialSvcLabels = sdService.getServiceLabels(initialService);
+        }
+        String initialSvcSidekickLabel = initialSvcLabels.
+                get(ServiceDiscoveryConstants.LABEL_SERVICE_SIDEKICK) == null ? null : initialSvcLabels
+                .get(ServiceDiscoveryConstants.LABEL_SERVICE_SIDEKICK);
```
ibuildthecloud: is this essentially `A == null ? null : A`?

alena1108: jeez :) originally it was something else, then did the refactoring and it doesn't make sense any more

ibuildthecloud: Too many late nights :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+    ObjectProcessManager objectProcessManager;
+    @Inject
+    ObjectManager objectManager;
+    @Inject
+    ServiceExposeMapDao exposeMapDao;
+
+    @Override
+    public void activate(final Service service, final Map<String, Object> data) {
+        final List<Service> services = exposeMapDao.collectSidekickServices(service, null);
+        if (services.isEmpty()) {
```
ibuildthecloud: This can never be true, but you should maybe check if service == null, just return as the first thing in this function

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import javax.inject.Inject;
+
+import org.apache.commons.lang3.tuple.Pair;
+import jersey.repackaged.com.google.common.collect.Sets;
```
ibuildthecloud: Not the Sets you were looking for :)  #EclipseFail

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+            }
+        } catch (IdempotentRetryException ex) {
+            // if not caught, the process will keep on spinning forever
+            // figure out better solution
+        }
+
+    }
+
+    protected LockDefinition createLock(List<Service> services) {
+        return new ServicesSidekickLock(Sets.newHashSet(services));
```
ibuildthecloud: Just change the constructor to `Collection<Service>` and then you don't have to convert

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+                    scale - units.size());
+            availableNames.put(service, instanceNames);
+        }
+        
+        int i = 0;
+        while (units.size() < scale) {
+            String uuid = UUID.randomUUID().toString();
+            Map<Service, DeploymentUnitInstance> serviceToInstances = new HashMap<>();
+            for (Service service : services) {
+                Pair<String, Object> instanceNamePair = Pair.of(availableNames.get(service).get(i),
```
ibuildthecloud: In general I'm not a fan of `Pair` unless it adds a lot of value (like a simple composite key to a map).  Here I don't quite see why it's needed.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/LoadBalancerDeploymentUnitInstance.java
```diff
@@ -0,0 +1,113 @@
...
+    public void remove() {
+        objectProcessManager.scheduleProcessInstance(LoadBalancerConstants.PROCESS_LB_HOST_MAP_REMOVE,
+                objectManager.reload(hostMap),
+                null);
+    }
+
+    @Override
+    public DeploymentUnitInstance start(List<Integer> volumesFromInstancesIds) {
+        if (this.hostMap == null) {
+            LoadBalancer lb = objectManager.findOne(LoadBalancer.class, LOAD_BALANCER.SERVICE_ID, service.getId(),
```
ibuildthecloud: You have to be careful with findOne, if it finds two, it blows up with an exception.  Where as findAny just takes the first.  Sometime you have to decide if you want it to blow up or does something slightly wrong...  Not an easy decision a lot of the time.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -92,24 +55,10 @@ public void run() {
                         .and(SERVICE_EXPOSE_MAP.SERVICE_ID.eq(serviceId))
                         .and(SERVICE_EXPOSE_MAP.STATE.in(CommonStatesConstants.ACTIVATING,
                                 CommonStatesConstants.ACTIVE, CommonStatesConstants.REQUESTED)))
```
ibuildthecloud: In general you never need to consider requested.  It is a bug in the code if there is every anything in the DB with requested.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnitInstanceFactoryImpl.java
```diff
@@ -0,0 +1,115 @@
...
+         * If no 'io.rancher.deployment.unit' label is found on the container, generate a random UUID for the value
+         * and use that
+         */
+
+        List<DeploymentUnitInstance> deploymentUnitInstances = new ArrayList<>();
+        if (service.getKind().equalsIgnoreCase(KIND.SERVICE.name())) {
+            List<? extends Instance> serviceContainers = exposeMapDao.listServiceInstances(service.getId());
+            for (Instance serviceContainer : serviceContainers) {
+                Map<String, String> instanceLabels = DataAccessor.fields(serviceContainer)
+                        .withKey(InstanceConstants.FIELD_LABELS).as(Map.class);
```
ibuildthecloud: add `withDefault(Collections.emptyMap())` to avoid NPE

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnitInstanceFactoryImpl.java
```diff
@@ -0,0 +1,115 @@
...
+                String deploymentUnitUUID = instanceLabels.get(ServiceDiscoveryConstants.LABEL_SERVICE_DEPLOYMENT_UNIT);
```
ibuildthecloud: Where do you generate a random UUID if one is missing?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnitInstanceFactoryImpl.java
```diff
@@ -0,0 +1,115 @@
...
+                        .add(createDeploymentUnitInstance(deploymentUnitUUID, service, nameToInstanceMap));
+            }
+        } else if (service.getKind().equalsIgnoreCase(KIND.LOADBALANCERSERVICE.name())) {
+            LoadBalancer lb = objectManager.findOne(LoadBalancer.class, LOAD_BALANCER.SERVICE_ID, service.getId(),
+                    LOAD_BALANCER.REMOVED, null);
+
+            List<? extends LoadBalancerHostMap> hostMaps = mapDao.findNonRemoved(LoadBalancerHostMap.class,
+                    LoadBalancer.class, lb.getId());
+            for (LoadBalancerHostMap hostMap : hostMaps) {
+                // FIXME - read UUID from hostmap label
```
ibuildthecloud: does this need to be fixed? :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+        final List<Service> services = exposeMapDao.collectSidekickServices(service, null);
+        if (services.isEmpty()) {
+            return;
+        }
+
+        lockManager.lock(createLock(services), new LockCallbackNoReturn() {
+            @Override
+            public void doWithLockNoResult() {
+                final List<DeploymentUnit> units = collectDeploymentUnits(services);
+                final int requestedScale = DataAccessor.fieldInteger(services.get(0),
```
ibuildthecloud: Let's be a bit smarter about this.  Let's take the max scale of the services that are not side kicks.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+                        ServiceDiscoveryConstants.FIELD_SCALE);
+                // don't process if there is no need to reconcile
+                boolean needToReconcile = needToReconcile(services, units, requestedScale);
+
+                if (!needToReconcile) {
```
ibuildthecloud: Is this to avoid loops? between sidekicks?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+            addMissingUnits(services, units, requestedScale);
+        } else if (units.size() > requestedScale) {
+            removeExtraUnits(units, requestedScale);
+        }
+
+        return units;
+    }
+
+    private void removeExtraUnits(List<DeploymentUnit> units, int scale) {
+        // delete units
```
ibuildthecloud: Is the ordering here deterministic at all?  Just asking.  I assume in the future we will probably want to tweak this to be a bit smarter.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -0,0 +1,364 @@
...
+            DeploymentUnit unit = new DeploymentUnit(uuid, serviceToInstances);
+
+            units.add(unit);
+            i++;
+        }
+    }
+
+    private void fillUpMissingUnitInstances(List<Service> services, List<DeploymentUnit> units) {
+        // check if there are incomplete units
+        Map<Service, Integer> serviceToUnitInstanceCount = new HashMap<>();
```
ibuildthecloud: I've haven't quite understood this logic yet, but could this be encapsulated in the `DeploymentUnitInstance` itself, in the start unit.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: True, would fix

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: ok, will do

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: Ok, I can just pass it as 2 separate params.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/LoadBalancerDeploymentUnitInstance.java
alena1108: It should never find 2 in this case. Service LB gets created/removed along with the service. But I see your concern - I hit "findOne returned multiple records" error before for other use cases.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnitInstanceFactoryImpl.java
alena1108: ok

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnitInstanceFactoryImpl.java
alena1108: I always generate it per DeploymentUnit basis, not per instance basis. For deploymentUnit, its generated in DeploymentManagerImpl, addMissingUnits call

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnitInstanceFactoryImpl.java
alena1108: hehe, fixed already. I'm bad when it comes to cleanup after refactoring

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: @ibuildthecloud Darren, could you elaborate? Why "are Not sidekicks"? And why max if I always make sure to set the scale to for all sidekick services in ResourceManager, so we should never hit the point when services of the same sidekick have a diff scale

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: yes

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: No, order doesn't matter neither between the units, nor within the units. I agree that we might optimize it in the future.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: I do it outside - in the manager - only because I'm trying to preserve order for the instances names. If several units have to fill up the gaps for several services, for each service I grab the available names (and fill up the gaps in names if needed), then redistribute the names across units.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
alena1108: as per our hangouts discussion, at the moment service gets created and sidekick is specified, its scale will get set to the existing sidekick's services scale

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
alena1108: for serviceExposeMap, I don't kick off the activation process till the vm starts. Let me explain why. ServiceExposeMap activation triggers DNS and healthcheck updates, and for those config items the IP address has to be set on the instance already. So I couldn't think of better handling for it then:

* serviceExposeMap gets created along with the instanceCreate.
* Then instance.start triggers serviceExposeMap.activate
* serviceExposeMap.activate triggers DNS and Healthcheck update

I'm open for better solution suggestion :)

alena1108: Will review and remove

alena1108: Darren, when we validate if the service number of instances against the scale (to see if more instances need to be added/removed), we use this method. So as the map gets scheduled for creation as a part of service instance create, not executed, its possible that it might be in Requested state even though the service instance record is created (especially when idempotent checks are on)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
ibuildthecloud: It would be nicer encapsulation if the deploymentmanagerimpl could created deploymentunits and then deployment units would be responsible for creating unit instances.  Even for addMissingUnits, it would be cleaner if you just created the unit with the list of services it must implement and the unit then had the logic to create the unit isntances.  This means you probably need to pass in some factory objects to the unit since they can't be injected.

ibuildthecloud: Looking at the DeploymentUnit now... The class is a bit anemic.  I think it might be nice if the constructor just took a `List<Service>`, `DeploymentUnitInstanceFactory`.  But I do see how the instance names makes things hard.  What if you created a new class like

```java
interface DeploymentUnitInstanceNameGenerator {
  String getNextAvailableName()
}
```

Then in the deployment manager at the beginning you do the following logic

```java
  List<String> usedNames = collectedUsedInstanceNames(service);
  serviceToNameFactory.put(service, new NameGeneratorImpl(service, usedNames);
}
```

Then pass that generator to each deployment unit.  At the time the unit needs to create a unit instance it just calls the generator to get the next name.  The generator is create and destroy per call to DeploymentManager.activate(), so it can hold the transient state of which names have been "reserved".

The getNextAvailableName needs to synchronized and would do something like

```java
String testName = null;
do {
  testName = pickNextName();
} while (isNotTaken(testName));

markTaken(testName);

return testName;
```

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    protected DeploymentUnitInstance createInstance(Map<Service, DeploymentUnitInstance> serviceToInstance,
+            Service service, boolean waitForStart) {
+        List<Integer> volumesFromInstanceIds = new ArrayList<>();
+        List<Integer> volumesFromServiceIds = DataUtils.getFields(service).get(
+                ServiceDiscoveryConfigItem.VOLUMESFROMSERVICE
+                .getRancherName()) == null ? new ArrayList<Integer>()
```
ibuildthecloud: nit: I'd rather not have cattle refer to `Rancher` by name.  Rancher is the product name, where as cattle is the name of the orchestration engine.  Cattle is an internal name that will probably never change, it's just a component.  Who knows tomorrow for marketing reason we might rename the company to Whale Wrangler.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -0,0 +1,149 @@
...
+        for (DeploymentUnitInstance instance : serviceToInstance.values()) {
+            instance.waitForStart();
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    protected DeploymentUnitInstance createInstance(Map<Service, DeploymentUnitInstance> serviceToInstance,
+            Service service, boolean waitForStart) {
+        List<Integer> volumesFromInstanceIds = new ArrayList<>();
+        List<Integer> volumesFromServiceIds = DataUtils.getFields(service).get(
```
ibuildthecloud: I think if you use the `.withDefault()` method this line could be cleaned up.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Callable;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+public class RegularDeploymentUnitInstance extends DeploymentUnitInstance {
```
ibuildthecloud: nit: Maybe `DefaultDeploymentUnitInstance` would be a better name.  Who's to say what is regular or not.  It just feels this may seem unkind if you also create a `StarBellySneetchDeploymentUnitInstance` who will then feel as though he is not as important as the `RegularDeploymentUnitInstance` manager.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
ibuildthecloud: Sorry, you're right.  I was thinking about the from the API only.  In the API layer it is all one transaction, but you are absolutely right.  On the backend it isn't.  Good job, just testing you there :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
```diff
@@ -0,0 +1,173 @@
...
+    ServiceExposeMapDao exposeMapDao;
+
+    protected String instanceName;
+    protected ServiceExposeMap exposeMap;
+
+    public RegularDeploymentUnitInstance() {
+        super(null, null, null, null, null, null);
+    }
+
+    public RegularDeploymentUnitInstance(String uuid, Service service,
```
ibuildthecloud: To make this constructor less unwieldy maybe all of these service classes should be encapsulated in some DeploymentServiceContext object.  Similar to `io.github.ibuildthecloud.gdapi.validation.ValidationHandler.ValidationContext`.  Just a simple class with final members and no getters/setters.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    @Override
+    public boolean isStarted() {
+        return objectManager.reload(this.instance).getState().equalsIgnoreCase(InstanceConstants.STATE_RUNNING);
+    }
+
+    @Override
+    public boolean needsCleanup() {
+        // TODO Auto-generated method stub
```
ibuildthecloud: lol, your needsCleanup method needs some cleanup :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
```diff
@@ -0,0 +1,173 @@
...
+        return false;
+    }
+
+    protected Long getImage(String imageUuid, Integer registryCredentialId) {
+        Image image;
```
ibuildthecloud: @alena1108, @hibooboo2 recently made some changes to the registryCredentialIds.  I think this will cause some problems.  @hibooboo2 Where do you pick the registryCredentialId?  Which class is that?

code/iaas/service-discovery/server/src/main/resources/META-INF/cattle/system-services/spring-service-discovery-server-context.xml
```diff
@@ -45,7 +43,12 @@
        <bean class="io.cattle.platform.servicediscovery.service.impl.RancherImageToComposeFormatter" />
        <bean class="io.cattle.platform.servicediscovery.service.impl.RancherRestartToComposeFormatter" />
        <bean class="io.cattle.platform.servicediscovery.service.impl.RancherLoadBalancerConfigToComposeFormatter" />
-
+       <bean class="io.cattle.platform.servicediscovery.deployment.impl.DeploymentManagerImpl" />
+	   <bean class="io.cattle.platform.servicediscovery.deployment.impl.DeploymentUnitInstanceFactoryImpl" />
```
ibuildthecloud: whitespace, no tabs

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
alena1108: :) once healtchecks are in, and we can validate against them, this method will get more content :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
alena1108: Will fix.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
alena1108: haha :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
alena1108: Always appreciate tips like this, Darren. Will refactor

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/RegularDeploymentUnitInstance.java
alena1108: Would be nice if we moved getImage code from ApiFilter to some manager...this way I won't have to duplicate it, and can reference it directly.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: Thank you, let me try that

ibuildthecloud:
@alena1108 The build failed on `test_link_volumes` from `core/test_svc_discovery.py`

PR #434, opened 2015-05-14
Option to skip delegation on process cancel

cjellick:
This change adds the ability to say that you don't want a process to be delegated if it is canceled.

Here's why this is needed:
1. The "default process" implementation has the deactivate process setup to delegate to remove. 
2. If a scheduled process is cancelled, its delegate (if it has one) is automatically ran (even though the original process was cancelled).
3. With native docker container events and a short lived container, the create, start, and stop events come in to rancher rapidly. In this scenario, the intended behavior was for the stop action to simply be cancelled by the framework and effectively ignored, but since stop delegates to remove, remove is ran and the container jumps from created to removing (and then to removed).

It seems odd for deactivate (which is renamed to stop for instances) to delegate to remove. It also seems odd to immediately delegate to another process if a process is cancelled. But the fact of the matter is that a lot of things break and a lot of tests fail if we change either of those things.

So, this change allows for an override to the behavior without changing it for everything that relies on it. You can set the `skipDelegateOnCancel` option on a process's data and if the process is cancel, it will not delegate.

### DIFF
--- a/code/framework/engine/pom.xml
+++ b/code/framework/engine/pom.xml
@@ -8,6 +8,11 @@
         <relativePath>../../parent/pom.xml</relativePath>
     </parent>
     <dependencies>
+         <dependency>
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-framework-utils</artifactId>
+            <version>${project.version}</version>
+        </dependency>
         <dependency>
             <groupId>io.cattle</groupId>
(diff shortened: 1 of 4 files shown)


cjellick:
Throwing this up here to see how it does in drone. Still need to test more to see if it is impactful. Will put it out on a 10-acre and see if any validation tests blow up.

cjellick:
Closing. Real fix is in #474

PR #438, opened 2015-05-14
add rackspace config

wlan0:


wlan0:
@cjellick @will-chan

code/implementation/docker/machine/src/main/resources/schema/base/rackspaceConfig.json.d/rackspace.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        "sshUser": {
+            "type": "string",
+            "nullable": true
+        },
+        "sshPort": {
+            "type": "string",
+            "nullable": true
+        },
+        "dockerInstall": {
```
cjellick: It's less-than-ideal that we're using a string type where the value is going to be "true", but it looks like in the cli, they treat it as a string and want `--docker-install "true"`vs just `--docker-install` like we've seen with other truly boolean parameters.
@wlan0 have you tested this one specifically? Just asking because we got burnt by booleans once before.

cjellick:
Other than that one question, LGTM. So, once you confirm that, my LGTM is official.

code/implementation/docker/machine/src/main/resources/schema/base/rackspaceConfig.json.d/rackspace.json
wlan0: @cjellick I have tested that one specifically. The same question came up while writing this and I tested it.

will-chan:
LGTM.  The last one is weird but that's what the rackspace driver has.  It was a boolean at one point but for 0.2.0 DM, that's what it is.

PR #440, opened 2015-05-14
Can disable and enable actions based on !state using json.

hibooboo2:


hibooboo2:
```        "logs": {
            "input":"containerLogs",
            "output":"hostAccess",
            "attributes" : {
                "state": "!stopped"
            }
        },```

Would make logs only avaliable if state is !stopped.

ibuildthecloud:
There has got to be some tests for this.  If you don't have corresponding functionality that uses this PR, then just hold off and submit the this change when something uses it.

PR #442, opened 2015-05-15
Dont assert initial state on ha tests.

cjellick:
We should assert that contianers are running initially in test_ha.py
because the ping logic may have already gotten to them and stopped them.


alena1108:
LGTM

PR #448, opened 2015-05-17
Don't process docker create events

cjellick:
Wait until a start event is received to create a container. This will
let us always no the container's ip address when creating it, which we
use to determine its rancher network settings.


alena1108:
LGTM

sonchang:
LGTM

PR #452, opened 2015-05-18
Introduce new service for getting list of hosts for global service

sonchang:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+public interface AllocatorService {
+
+    public List<Long> getHostsForGlobalService(Long accountId, Map<String, String> labels);
+
+    @SuppressWarnings("rawtypes")
+    public List<Constraint> extractConstraintsFromEnv(Map env);
```
ibuildthecloud: Are we still supporting env variables?

sonchang: Yeah.  We're still supporting env variables.  i figure doesn't hurt

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/AffinityConstraintDefinition.java
```diff
@@ -7,8 +7,8 @@
         EQ("==", "{eq}"),
         NE("!=", "{ne}");
 
-        String envSymbol;
-        String labelSymbol;
+        public String envSymbol;
```
alena1108: As a part of my refactoring that got merged to master recently, I've introduced getters for retrieving these values (instead of "public" access modifier). Would it be possible for your code to use getters as well, so we are consistent.

alena1108:
LGTM

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/AffinityConstraintDefinition.java
sonchang: Yeah.  I can make this change

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorService.java
```diff
@@ -0,0 +1,18 @@
+package io.cattle.platform.allocator.service;
+
+import io.cattle.platform.allocator.constraint.Constraint;
+
+import java.util.List;
+import java.util.Map;
+
+public interface AllocatorService {
+
+    public List<Long> getHostsForGlobalService(Long accountId, Map<String, String> labels);
```
alena1108: Minor: we should probably rename this method as its not specific to GlobalService. It just returns all the hosts that satisfy the labels requirements.

sonchang: well, just the HostAffinity label requirement.  What about this name?
getHostsFilteredByHostAffinityConstraint()
In the future, I might refactor some more if the need arises

PR #454, opened 2015-05-18
SD:Propagate labels defined in launchConfig

alena1108:
Fixes the bug when labels defined in service launch config, weren't propagated to the service instance along with the labels generated by the service internally


cjellick:
LGTM

sonchang:
LGTM

PR #455, opened 2015-05-18
Service Discovery bug fixes

alena1108:
1) LBService: restart instance if stopped on activate
https://github.com/rancherio/rancher/issues/844

2) Check if service referenced via volumesFrom, still exists before resolving the instanceIds

https://github.com/rancherio/rancher/issues/849

3) Always wait for instance referenced via volumesFrom, to be fully started before creating a referencing instance. Otherwise the parameter might be ignored on the docker side.

https://github.com/rancherio/rancher/issues/846

And one more fix not corresponding to any bug - store used service generated instance names in map with key=serviceId instead of key=service. If we do the latter, the same DB object can be stored multiple times in the map, and it will grow fast.


cjellick:
LGTM

sonchang:
LGTM

sonchang:
after you've squashed, I can merge

PR #461, opened 2015-05-19
meaningless change

cjellick:


ibuildthecloud:
Don't lose hope.  We're here for you.  While you may think your work is meaningless, insignificant, a waste of time, overlooked, "what have you been doing for the last month?", or not appreciated, that is not the totally true, most of the time.  While you only seemed to have contributed one line of comments, it's not the size, but the importance of that comment that we appreciate.

cjellick:
:-)

PR #464, opened 2015-05-19
Make ContainerEventCreate survive null inspect

cjellick:
ContainerEventCreate relied on inspect being not null for no good reason. This change makes it possible for dockerInspect to be null without causing ContainerEventCreate to blow and keep getting retried forever.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
@@ -162,10 +162,7 @@ public class ContainerEventCreate extends AbstractDefaultProcessHandler {
     }
 
     Map<String, Object> getInspect(ContainerEvent event) {
-        Object obj = DataUtils.getFields(event).get(FIELD_DOCKER_INSPECT);
-        if (obj == null)
-            return null;
-        return CollectionUtils.toMap(obj);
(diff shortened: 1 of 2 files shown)


sonchang:
LGTM

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -185,9 +182,10 @@ void setHost(ContainerEvent event, Instance instance) {
     void setName(Map<String, Object> inspect, Map<String, Object> data, Instance instance) {
         String name = DataAccessor.fromMap(data).withKey(CONTAINER_EVENT_SYNC_NAME).as(String.class);
         if (StringUtils.isEmpty(name))
-            name = (String)inspect.get(INSPECT_NAME);
+            name = DataAccessor.fromMap(inspect).withKey(INSPECT_NAME).as(String.class);
 
-        name = name.replaceFirst("/", "");
+        if (name != null)
+            name = name.replaceFirst("/", "");
         instance.setName(name);
```
alena1108: Craig, any possible side affects if the name is null, and then we set it on the instance on line 189?

cjellick: @alena1108 no. Name can be null in rancher and if it is, the UI displays the ID instead

alena1108:
LGTM

PR #466, opened 2015-05-20
Handle unfound host for agent better in ping

cjellick:
LoadingCache doesn't allow null values. If you want to indicate that a
value could not be found for your key, you must raise (and handle) an exception.

The returned value can be null if a ping occurs between an agent responds to an instance ping before the host has been created.

### DIFF
--- a/code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
+++ b/code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
@@ -49,6 +49,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.netflix.config.DynamicLongProperty;
 
 public class PingInstancesMonitorImpl implements PingInstancesMonitor {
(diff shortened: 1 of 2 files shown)


cjellick:
https://github.com/rancherio/rancher/issues/852

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -344,17 +354,20 @@ protected ReportedInstances getInstances(Ping ping) {
...
-        if (agentIdAndHostUuid.left == null || StringUtils.isEmpty(agentIdAndHostUuid.right))
-            return null;
+        Long agentId = agentIdAndHostUuid.left;
+        String hostUuid = agentIdAndHostUuid.right;
+
+        Agent agent = objectManager.loadResource(Agent.class, agentId);
+        Host host = null;
+        Map<String, Host> hosts = null;
+        if (agent != null) {
+            hosts = agentDao.getHosts(agent.getId());
```
alena1108: We can optimize here and introduce new Dao method searching by agentId/hostUUID. So there would be no need for line 365.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -344,17 +354,20 @@ protected ReportedInstances getInstances(Ping ping) {
     }
 
     protected AgentAndHost loadAgentAndHostData(ImmutablePair<Long, String> agentIdAndHostUuid) {
```
alena1108: Minor: there is no real need to use Pair here, as you can just pass agentId/hostUUID as 2 separate params (and in this case you can avoid creating a new object + conversion back to long/string in loadAgentAndHostData).

I usually use pair only when its necessary (for return type for example).

ibuildthecloud:
You analysis perfectly describe why I'm seeing this more on the network mode branch.  On that branch the agent connect logic is different. Previously we would wait for a success ping that has resources to determine that the agent is active.  Now we do not wait.  This means it could be seconds after the agent is active before we add a host.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: @alena1108 It is a pair because of how the LoadingCache works. The pair is the key to its map:
https://github.com/rancherio/cattle/pull/466/files#diff-ba9537ff87cd70d9ee6498f7391d5ea1R88
and I must provide a method that will load a value based off of that key.

code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
cjellick: hostUUID in this case is actually "reportedHostUuid" and is stored in the json data field.

alena1108: ah, understood

alena1108:
LGTM

cjellick:
@ibuildthecloud rebased.

PR #467, opened 2015-05-20
Service discovery bug fixes

alena1108:
1) Set dataVolumes for LB service

https://github.com/rancherio/rancher/issues/861

2) Start instances referred by volumesFrom (if exist, but not running)

https://github.com/rancherio/rancher/issues/860

3) Fixed bug when LoadBalancer instance manager was creating the agent for removed hostMap. We should exclude those maps from the logic

4) In ServiceDiscovery code, schedule all processes through DeferredUtils so the event is published immediately. In this way the process will get picked faster for execution.

### DIFF
--- a/code/framework/object/src/main/java/io/cattle/platform/object/process/ObjectProcessManager.java
+++ b/code/framework/object/src/main/java/io/cattle/platform/object/process/ObjectProcessManager.java
@@ -25,4 +25,8 @@ public interface ObjectProcessManager {
 
     ExitReason executeProcess(String processName, Object resource, Map<String, Object> data);
 
+    void scheduleProcessInstanceAsync(String processName, Object resource, Map<String, Object> data);
+
+    void scheduleStandardProcessAsync(StandardProcess process, Object resource, Map<String, Object> data);
+
 }
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
@@ -8,7 +8,6 @@ import io.cattle.platform.core.model.Image;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.core.model.ServiceExposeMap;
-import io.cattle.platform.deferred.util.DeferredUtils;
 import io.cattle.platform.engine.process.impl.ProcessCancelException;
 import io.cattle.platform.object.process.StandardProcess;
 import io.cattle.platform.object.resource.ResourcePredicate;
@@ -20,7 +19,6 @@ import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.Callable;
 
 import org.apache.commons.lang3.tuple.Pair;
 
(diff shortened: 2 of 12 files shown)


code/framework/object/src/main/java/io/cattle/platform/object/process/ObjectProcessManager.java
```diff
@@ -25,4 +25,8 @@
 
     ExitReason executeProcess(String processName, Object resource, Map<String, Object> data);
 
+    void scheduleProcessInstanceAndPublishEvent(String processName, Object resource, Map<String, Object> data);
```
sonchang: I wonder should we rename these.  Maybe to scheduleNestedProcessInstance()?

alena1108: Sure, I'm fine with either name.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
```diff
@@ -94,19 +96,17 @@ public DeploymentUnitInstance start(List<Integer> volumesFromInstancesIds) {
...
+        context.objectProcessManager.scheduleStandardProcessAndPublishEvent(StandardProcess.CREATE, exposeMap,
+                null);
+
         if (InstanceConstants.STATE_STOPPED.equals(instance.getState())) {
             DeferredUtils.nest(new Callable<Object>() {
                 @Override
                 public Object call() throws Exception {
-                    context.objectProcessManager.scheduleProcessInstance(InstanceConstants.PROCESS_START, instance,
-                            null);
+                    context.objectProcessManager.scheduleProcessInstanceAndPublishEvent(
```
sonchang: Since this new call: scheduleProcessInstanceAndPublishEvent() already nests the process, do we need this outer one (line 105)?

alena1108: Good  catch :) will remove

code/framework/object/src/main/java/io/cattle/platform/object/process/ObjectProcessManager.java
cjellick: I don't know that either name is clear to me.
`scheduleProcessInstanceAsync` any better?

cjellick:
Reviewed. @alena1108 once you make the changes that @sonchang suggested, I'll LGTM.

code/framework/object/src/main/java/io/cattle/platform/object/process/ObjectProcessManager.java
alena1108: Its async in any case; it has similar functionality to  scheduleProcessInstance/scheduleStandardProcess with the only one difference - it gets scheduled within DeferredUtils context. And it means that the event will get generated and published right away, and the process engine will pick the process and execute it right after receiving the event. As opposed to the regular behavior, when periodic task within process engine scans for scheduled processes

alena1108: Thought about it more. Async in name makes sense, as the process gets scheduled in async manner by delegating scheduling to the Deferred Utils instead of scheduling it from the ObjectProcessManager. @sonchang what do you think about the name Craig suggested? I'm so bad when it comes to naming things :-/

sonchang: What about chainProcessInstance()?

alena1108: I guess we already use chainProcess term in cattle when we one process is followed by another...while in this case we schedule a single process instance

sonchang: I'm good with any name

cjellick:
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -299,7 +299,11 @@ private void populateExternalLinksForService(Service service, Map<String, Object
             if (volumesFrom == null) {
                 volumesFrom = new ArrayList<Integer>();
             }
-            volumesFrom.addAll(volumesFromInstanceIds);
+            for (Integer volumesFromInstanceId : volumesFromInstanceIds) {
+                if (!volumesFrom.contains(volumesFromInstanceId)) {
```
sonchang: Super minor: Not sure if it matters, but a Set would prevent duplicates.

alena1108: I know :) just wanted to avoid the back and forth conversion when retrieving/putting it to/from data field - there its stored in List format

sonchang:
LGTM

PR #471, opened 2015-05-21
fix label resource stuck in requested when idempotent checks is true

sonchang:
@cjellick


cjellick:
LGTM

ibuildthecloud:
LGTM

But unfortunately the tests are failing, I guess your just going to have to fix all of those :)  j/k

PR #475, opened 2015-05-22
fix allocation TF

sonchang:
@ibuildthecloud

### DIFF
--- a/tests/integration/cattletest/core/test_allocation.py
+++ b/tests/integration/cattletest/core/test_allocation.py
@@ -97,8 +97,12 @@ def test_allocation_failed_on_create(super_client, new_context):
     assert c.transitioning == 'error'
     assert c.transitioningMessage == 'Failed to find a placement'
 
-    assert c.allocationState == 'inactive'
+    assert c.allocationState == 'activating'
     assert c.volumes()[0].state == 'removed'
+
+    c = super_client.wait_success(super_client.reload(c.purge()))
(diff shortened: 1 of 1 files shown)


tests/integration/cattletest/core/test_allocation.py
```diff
@@ -97,9 +97,13 @@ def test_allocation_failed_on_create(super_client, new_context):
     assert c.transitioning == 'error'
     assert c.transitioningMessage == 'Failed to find a placement'
 
-    assert c.allocationState == 'inactive'
+    assert c.allocationState == 'activating'
     assert c.volumes()[0].state == 'removed'
 
+    c = super_client.wait_success(super_client.reload(c.purge()))
```
cjellick: Why do you need the reload and the wait_success?

sonchang: wait_success() is needed because there's a 'purging' transition state.  reload might not be necessary since I think wait_success() already does a reload anyways.

sonchang:
TF was actually caused by a change in InstanceDaoImpl's isOnSharedStorage().
This affected InstanceStop's shouldDeallocate() call.

tests/integration/cattletest/core/test_allocation.py
ibuildthecloud: The reload isn't needed.  I'm going to merge anyhow, not a big deal.

PR #477, opened 2015-05-22
Update agent image and python agent versions

cloudnautique:
These updates bring in the SSL updates for the container.


ibuildthecloud:
We should get master back to green today.  Would you mind waiting and rebasing?

PR #480, opened 2015-05-22
Schedule instance.remove on account/host purge

cjellick:
Using execute for instance.remove leads to idempotency failures because
of how we set the removed fields while the instance is still in
the "removing" state and then make a check against the removed field when
determing which instances to remove.


sonchang:
LGTM

alena1108:
LGTM

PR #500, opened 2015-05-26
Support network modes in native-docker

cjellick:
Adds NetworkMode transformation to inspect transformer api (used by compose).
Adds NetworkMode interpretation to native-docker container event
processing.


code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -40,12 +46,17 @@
...
 
 import com.netflix.config.DynamicBooleanProperty;
 
 @Named
 public class ContainerEventCreate extends AbstractDefaultProcessHandler {
 
+    public static final String AGENT_ID = "agentId";
+    private static final String INSTANCE_INSPECT_EVENT_NAME = "compute.instance.inspect";
+    private static final String INSTANCE_INSPECT_DATA_NAME = "instanceInspect";
     private static final DynamicBooleanProperty MANAGE_NONRANCHER_CONTAINERS = ArchaiusUtil.getBoolean("manage.nonrancher.containers");
```
alena1108: I got input from Darren once to try not to use "rancher" name in our internal naming; but use cattle instead..as rancher is a product name, and can get changed

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -161,8 +177,38 @@ void scheduleInstance(ContainerEvent event, Instance instance, Map<String, Objec
...
+    private Map<String, Object> getInspect(ContainerEvent event, Map<String, Object> data) {
+        Object inspectObj = DataUtils.getFields(event).get(FIELD_DOCKER_INSPECT);
+        if (inspectObj == null) {
+            EventVO<Object> inspectEvent = newInspectEvent(null, event.getExternalId());
+            inspectObj = callAgentForInspect(data, inspectEvent);
+        }
+        return CollectionUtils.toMap(inspectObj);
+    }
+
+   private Object callAgentForInspect(Map<String, Object> data, EventVO<Object> inspectEvent) {
```
alena1108: we should probably use type of Event interface instead of EventVO implementation. Shouldn't cause the problem switching because agent.callSync expects the interface.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -190,27 +236,108 @@ void setName(Map<String, Object> inspect, Map<String, Object> data, Instance ins
...
 
     void setNetwork(Map<String, Object> inspect, Map<String, Object> data, Instance instance) {
-        Network network = null;
-        Network rancherNetwork = networkDao.getNetworkForObject(instance, KIND_HOSTONLY);
-        String ip = getDockerIp(inspect);
-        if (StringUtils.isNotEmpty(ip) && isIpInNetwork(ip, rancherNetwork)) {
-            DataAccessor.fields(instance).withKey(FIELD_REQUESTED_IP_ADDRESS).set(ip);
-            network = rancherNetwork;
+        String networkMode = checkNoneNetwork(inspect);
+        if (setNetworkModeIfNotNull(instance, networkMode))
```
alena1108: I can see that the setNetworkModeIfNotNull returns true only when networkMode is not null. Do you think it would be more clear if we call this method only once, and only after networkMode is resolved to not null in the caller stack?

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -190,27 +236,108 @@ void setName(Map<String, Object> inspect, Map<String, Object> data, Instance ins
...
+        if (networkMode != null) {
+            DataAccessor.fields(instance).withKey(FIELD_NETWORK_MODE).set(networkMode);
+            return true;
+        }
+        return false;
+    }
+
+    private String checkNoneNetwork(Map<String, Object> inspect) {
+        Object netDisabledObj = CollectionUtils.getNestedValue(inspect, "Config", "NetworkDisabled");
+        if (netDisabledObj != null && netDisabledObj instanceof Boolean && (Boolean)netDisabledObj) {
```
alena1108: Nit: we can use "Boolean.TRUE.equals(Object)" structure here, and then you don't need extra not null/instanceof checks.

code/implementation/simulator/agent-connection/src/main/java/io/cattle/platform/agent/connection/simulator/impl/SimulatorInstanceInspectProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+import io.cattle.platform.util.type.CollectionUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class SimulatorInstanceInspectProcessor implements AgentSimulatorEventProcessor {
+
+    @Override
+    public Event handle(AgentConnectionSimulator simulator, Event event) throws Exception {
+        if (!"compute.instance.inspect".equals(event.getName())) {
```
alena1108: We should replace the hardcoded value with INSTANCE_INSPECT_EVENT_NAM here

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: Nice. Thanks.

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
cjellick: Yes, will change.

code/implementation/simulator/agent-connection/src/main/java/io/cattle/platform/agent/connection/simulator/impl/SimulatorInstanceInspectProcessor.java
cjellick: The reason I didn't was because it would introduce a new inter-project dependency. Will revisit to see if there is a cleaner way.

alena1108: As an idea: we can define this var in some project used by both projects then.

sonchang:
LGTM

code/iaas/logic/src/main/java/io/cattle/platform/process/containerevent/ContainerEventCreate.java
```diff
@@ -130,8 +146,14 @@ public HandlerResult doWithLock() {
...
                             objectProcessManager.scheduleStandardProcess(StandardProcess.REMOVE, instance, data);
                         } catch (ProcessCancelException e) {
-                            data.put(REMOVE_OPTION, true);
-                            objectProcessManager.scheduleProcessInstance(PROCESS_STOP, instance, data);
+                            if (STATE_STOPPING.equals(state)) {
+                                // handle docker forced stop and remove
+                                instance = resourceMonitor.waitForNotTransitioning(instance);
+                                objectProcessManager.scheduleStandardProcess(StandardProcess.REMOVE, instance, data);
+                            } else {
+                                data.put(REMOVE_OPTION, true);
```
alena1108: Not required for this PR, but just a hint: we can chain processes using ProcessUtils.chainInData method

alena1108:
LGTM

PR #501, opened 2015-05-26
Exponentially backoff processing events

sonchang:
@ibuildthecloud

### DIFF
--- a/code/framework/engine/src/main/java/io/cattle/platform/engine/eventing/impl/ProcessEventListenerImpl.java
+++ b/code/framework/engine/src/main/java/io/cattle/platform/engine/eventing/impl/ProcessEventListenerImpl.java
@@ -1,5 +1,6 @@
 package io.cattle.platform.engine.eventing.impl;
 
+import io.cattle.platform.archaius.util.ArchaiusUtil;
 import io.cattle.platform.async.utils.TimeoutException;
 import io.cattle.platform.engine.eventing.ProcessEventListener;
 import io.cattle.platform.engine.manager.ProcessManager;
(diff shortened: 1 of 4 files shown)


code/framework/engine/src/main/java/io/cattle/platform/engine/eventing/impl/ProcessEventListenerImpl.java
```diff
@@ -81,6 +91,25 @@ public void processExecute(Event event) {
...
+        ProcessRecord record = processRecordDao.getRecord(processId);
+        if (record == null || record.getEndTime() != null) {
+            // is there any additional processing that happens or can we
+            // even short circuit it here and be done with this event?
+            // for now, assume additional processing occurs to get rid
+            // of event
+            return false;
+        }
+        int numPreviousAttempts = processRecordDao.getNumPreviousExecutions(processId);
+        if (numPreviousAttempts < 2) {
```
sonchang: I wanted to give some leeway for the first retry in case of slight differences in clock times

code/framework/engine/src/main/java/io/cattle/platform/engine/eventing/impl/ProcessEventListenerImpl.java
```diff
@@ -81,6 +91,25 @@ public void processExecute(Event event) {
...
+            return false;
+        }
+
+        long now = System.currentTimeMillis();
+        long lastExecution = processRecordDao.getLastExecutionTimestamp(processId);
+        return now < lastExecution + Math.pow(2,  numPreviousAttempts) * 15000;
```
sonchang: We can tweak this too if we want.  I guess I should've extract this into an env variable.

ibuildthecloud:
@sonchang Should we go with this approach in the short term and then look at more scalable approach in the future?  If you feel good with that, then I'll review this in depth, right now I haven't looked at the actual backoff calculation logic.

sonchang:
@ibuildthecloud I feel good with that.  With Will's latest request, the earliest I'll probably look at this again is a week from now

PR #504, opened 2015-05-26
Projects refactor to allow basic auth supporting /projects/${id}/${endpoints}

hibooboo2:
As long as the person has an access_token cached in rancher.


code/iaas/accounts/src/main/resources/META-INF/cattle/accounts/spring-accounts-context.xml
```diff
@@ -46,6 +46,8 @@
                         <bean class="io.cattle.platform.spring.resource.SpringUrlListFactory" >
                             <property name="resources">
                                 <list>
+                                    <value>classpath:schema/token/token-auth.json</value>
+                                    <value>classpath*:schema/token/token-auth.json.d/**/*.json</value>
```
ibuildthecloud: What's the motivation for doing this?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/ExternalIdHandler.java
```diff
@@ -8,4 +11,6 @@
     
     ExternalId untransform(ExternalId externalId);
 
+    Set<ExternalId> getExternalIds(Account account); //Should never return null. If no externalIds then return and Empty set.
```
ibuildthecloud: Put that comment in a javadoc comment

code/iaas/accounts/src/main/resources/META-INF/cattle/accounts/spring-accounts-context.xml
hibooboo2: There was an error in the order of auth handlers and admin needed to be able to read the token as a result. But i fixed it. This was not supposed to still be here.

PR #514, opened 2015-05-28
Dont recreate removed containers

cjellick:
Addresses https://github.com/rancherio/rancher/issues/933

This fix will cause any events that come in once the container is removed in rancher to be ignored.

If the scenario does arise where a container that is removed in rancher is running on the host, there is logic in PingInstancesMonitorImpl to handle it. that logic will force the container stop on the host.


cjellick:
Ping @ibuildthecloud @alena1108 @sonchang

ibuildthecloud:
LGTM

PR #517, opened 2015-05-28
Expose healthState to an end user

alena1108:
And added supported values for the field to the API schema


alena1108:
Closing the PR as the field is going to move to Instance object

PR #520, opened 2015-05-28
Set networkMode on service instances

alena1108:
Otherwise networkId is not being set on the service instances


sonchang:
LGTM

cjellick:
LGTM

PR #523, opened 2015-05-29
process labels from host.create

sonchang:


ibuildthecloud:
whoa, something bad happened with that build

PR #524, opened 2015-05-29
 add azureConfig

wlan0:
@cjellick @will-chan


cjellick:
LGTM, but we need to coordinate this closely with the go-rancher and go-machine-service PRs and ensure they all get merged close together. It creates problems for @vincent99 and the UI if the API is available but the backend isnt. 

On the last one, we had the API in a release without the backend. I dont want to deal with that again.

So, won't merge this until the other two are good to go. Also @wlan0, @ibuildthecloud has a big go-rancher PR going. Can you coordinate with him on when/how azure should get merged into it.

vincent99:
@cjellick Technically it's only a problem for me if there's a UI for `$driver`.  I check for the schema to determine if the API supports that driver and hide but button if it's not there.  In this case I haven't built Azure yet so until that changes you _could_ merge in the API before backend (not suggesting it, just sayin')

will-chan:
azure is going to require more work anyways.  According to Sid, inbound ports are blocked so we will need to provide additional support much like EC2 to get the UX working nicely.

wlan0:
@will-chan @vincent99 Azure by default blocks all incoming and outgoing traffic to all ports. Docker-Machine opens up 22 for ssh and 2379 for docker. 

We need to open up 500, 4500 UDP for IPSec and 9345, 9346 TCP for host stats and 8080 TCP for the web UI

will-chan:
LGTM

cjellick:
LGTM. @wlan0 Please rebase and I'll get it merged as soon as the build is done.

PR #527, opened 2015-05-29
Renamed service.dataVolumesFromService param

alena1108:
to dataVolumesFromServiceIds

@ibuildthecloud as per your request. @vincent99 - UI would need an update to work with the new param name


ibuildthecloud:
As you pointed out container has dataVolumesFrom (not ids), so I'm fine with leaving this parameter as you named it.

alena1108:
@ibuildthecloud ok, closing it then

PR #528, opened 2015-05-29
Respect volumesFrom/requestedHostId for lb instance

alena1108:
Such an embarrassing regression

https://github.com/rancherio/rancher/issues/982


cjellick:
LGTM

PR #536, opened 2015-05-31
update global service

sonchang:


sonchang:
@ibuildthecloud @alena1108

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -636,4 +636,19 @@ protected Long getImage(String imageUuid, Integer registryCredentialId) {
 
         return image == null ? null : image.getId();
     }
+
+    // TODO: It'd be nice to have a service_label_map rather than inspect all the labels within the field
+    @Override
+    public List<Service> getActiveGlobalServices() {
```
alena1108: We should probably consider services in Activating state as well.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -246,6 +246,17 @@ public void doWithLockNoResult() {
         });
     }
 
+    @Override
+    public void activateGlobalServicesForHost(long hostId) {
+        List<Service> services = sdSvc.getActiveGlobalServices();
+        for (Service service: services) {
```
alena1108: I think we should think about calling activate() in parallel here, otherwise things can get slow + if for example service1 fails to activate due to lock wait timeout, service2 activation wouldn't be attempted in this run.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -636,4 +636,19 @@ protected Long getImage(String imageUuid, Integer registryCredentialId) {
...
+        List<Service> activeGlobalServices = new ArrayList<Service>();
+        List<Service> activeServices = objectManager.find(Service.class, SERVICE.REMOVED, null, SERVICE.STATE, CommonStatesConstants.ACTIVE);
```
ibuildthecloud: This should be scoped to an account

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: Just spoke with @ibuildthecloud about it - the problem is we currently don't have an ability to schedule non-blocking deploymentManager.serviceAcivate. Darren's healtcheck monitoring would also need this functionality for scheduling reconcile on heatlh state updte. We should think about what would be the best solution in this case.

sonchang:
@alena1108  @ibuildthecloud  Included Activating state and scoped to accountId.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
sonchang: I'll try to think about this more and address this in a subsequent PR

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -194,4 +196,14 @@ public ServiceExposeMap getServiceIpExposeMap(Service service, String ipAddress)
...
     }
 
+    @Override
+    public List<? extends Service> getGlobalServices(long accountId) {
+        return create()
+                .select(SERVICE.fields())
+                .from(SERVICE)
+                .where(SERVICE.ACCOUNT_ID.eq(accountId))
+                .and(SERVICE.REMOVED.isNull())
+                .and(SERVICE.STATE.in(CommonStatesConstants.ACTIVE, CommonStatesConstants.ACTIVATING))
```
alena1108: Sorry, should have included it earlier - we should also include updating-active state here.

alena1108:
Besides that generic problem with not being able to call DeploymentManager.activate in non-blocking manner, LGTM

ibuildthecloud:
L(meh)TM.  The solutions isn't great, but there isn't a better one right now.  So I'm fine with it.

PR #539, opened 2015-06-01
fix retrieving global services; fix merging of labels

sonchang:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorServiceImpl.java
```diff
@@ -87,13 +87,19 @@ public void mergeLabels(Map<String, String> srcMap, Map<String, String> destMap)
             return;
         }
         for (Map.Entry<String, String> entry : srcMap.entrySet()) {
-            String destValue = destMap.get(entry.getKey());
-            if (StringUtils.isEmpty(destValue)) {
-                destMap.put(entry.getKey(), entry.getValue());
-            } else if (StringUtils.isEmpty(entry.getValue())) {
-                continue;
+            if (entry.getKey().startsWith("io.rancher.scheduler.affinity")) {
+                // merge labels
```
sonchang: only merge the affinity labels; otherwise, just overwrite

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorServiceImpl.java
```diff
@@ -87,13 +87,19 @@ public void mergeLabels(Map<String, String> srcMap, Map<String, String> destMap)
...
+                String destValue = destMap.get(entry.getKey());
+                if (StringUtils.isEmpty(destValue)) {
+                    destMap.put(entry.getKey(), entry.getValue());
+                } else if (StringUtils.isEmpty(entry.getValue())) {
+                    continue;
+                } else if (!destValue.contains(entry.getValue())) {
```
sonchang: This is to prevent the same value from constantly appending.  So we don't have something like:
foo=bar,foo=bar,foo=bar,foo=bar,foo=bar

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/dao/impl/ServiceExposeMapDaoImpl.java
```diff
@@ -197,7 +197,7 @@ public ServiceExposeMap getServiceIpExposeMap(Service service, String ipAddress)
     }
 
     @Override
-    public List<? extends Service> getGlobalServices(long accountId) {
+    public List<? extends Service> getActiveServices(long accountId) {
```
sonchang: This is a bit embarrassing.  I had an idea over the weekend, but just half-way executed and then got distracted.  Anyways, this call wasn't retrieving global services (I was originally tempted to just do a string regex in the DB field to do the filtering there); but, that's too hacky.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/ServiceDiscoveryService.java
```diff
@@ -39,5 +39,4 @@
 
     List<Integer> getServiceInstanceUsedOrderIds(Service service);
```
sonchang: Also, sadly, forgot to remove this after having DeploymentManagerImpl retrieve global services from expMapDao instead of sdSvc

alena1108: Why do you have to remove this particular method?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorServiceImpl.java
```diff
@@ -87,13 +87,19 @@ public void mergeLabels(Map<String, String> srcMap, Map<String, String> destMap)
             return;
         }
         for (Map.Entry<String, String> entry : srcMap.entrySet()) {
-            String destValue = destMap.get(entry.getKey());
-            if (StringUtils.isEmpty(destValue)) {
-                destMap.put(entry.getKey(), entry.getValue());
-            } else if (StringUtils.isEmpty(entry.getValue())) {
-                continue;
+            if (entry.getKey().startsWith("io.rancher.scheduler.affinity")) {
```
alena1108: Should we introduce some variable for the label name, instead of hardcoding? In case we decide to change it in the future

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/ServiceDiscoveryService.java
sonchang: This function hasn't been removed.  The one removed was: getActiveGlobalServices()

alena1108:
LGTM

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorServiceImpl.java
sonchang: Yeah.  I'll do that later

PR #542, opened 2015-06-01
Named service links suppport

alena1108:
API changes:

1) addServiceLink. New parameter "name" (optional) got added
2) setServiceLinks. "serviceIds" array is replaced with "serviceLinks" map, where key is the link name, and value is the reference to the service. Unfortunately couldn't make link name optional (by making Service a key in the map) as our API framework supports only Strings for a key

@vincent99 @ibuildthecloud please review


vincent99:
Yes, that does indeed appear to be Java.

We'll need to synchronize merging this and the UI change to support it (which I haven't written yet), unless you want to make it backwards compatible for one release by having the input take `serviceIds` (for old behavior) *or* `serviceLinks` (for new behavior).

alena1108:
I would rather wait for the UI to be ready; maintaining backwards compatibility would be a bit of overhead before official beta/GA...unless you object otherwise @vincent99 @ibuildthecloud

ibuildthecloud:
@alena1108 can you rebase?

vincent99:
UI PR, we should merge together: https://github.com/rancherio/ui/pull/189

PR #545, opened 2015-06-02
fix bug with deploying containers to specific hosts via service

sonchang:
@alena1108


alena1108:
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -194,7 +194,6 @@ public boolean isComplete() {
     protected Map<String, String> getLabels(DeploymentUnitInstance instance) {
         Service service = instance.getService();
         Map<String, String> labels = new HashMap<>();
-        labels.putAll(context.sdService.getServiceLabels(service));
```
sonchang: I thought I took this out over the weekend already per alena's comment.   Not sure how it slipped back in.  But the service labels get merged in a different part of the code.

PR #551, opened 2015-06-02
Add parsing for sha256 tags

hibooboo2:


hibooboo2:
This addresses rancherio/rancher#977

code/implementation/docker/storage/src/main/java/io/cattle/platform/docker/client/DockerImage.java
```diff
@@ -33,63 +33,66 @@ public DockerImage(String id, String repository, String namespace, String tag, S
...
+        }
+        return fromArray(fields);
+    }
+
+    private static String[] repoAndTag(String uuid){
+        String[] fields = new String[4];
+        String[] split;
+        if (uuid.contains("@")){
+            split = uuid.split("@");
+            fields[2] = split[0];
```
sonchang: Since any value could be passed in via the API (I think), it's possible to run into an indexOutOfBounds here.

hibooboo2: If you read above i parse before calling this function.

hibooboo2: Actually you are right need to make this test the split length.

sonchang:
LGTM

alena1108:
LGTM

PR #554, opened 2015-06-03
Wait on serviceExposeMap to be removed during reconcile cleanup

alena1108:
for units in a "bad" state, before creating a new units. As the "used"
instances names generator might get wrong data based on the
serviceExposeMap that was scheduled for the removal, but wasn't removed
yet.

Should potentially fix one of the TF observed on master

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
@@ -15,7 +15,6 @@ import io.cattle.platform.servicediscovery.deployment.InstanceUnit;
 import io.cattle.platform.servicediscovery.deployment.impl.DeploymentManagerImpl.DeploymentServiceContext;
 
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.lang3.tuple.Pair;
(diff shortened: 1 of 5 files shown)


alena1108:
@ibuildthecloud @sonchang the unrelated test failed - test_docker.py, test_container_fields

ibuildthecloud:
@alena1108 Instead of waiting wouldn't just checking for !=REMOVING state work also?  I'd rather not wait if we don't really have to.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
```diff
@@ -60,14 +56,6 @@ public void remove() {
                                 InstanceConstants.PROCESS_STOP, InstanceConstants.PROCESS_REMOVE));
             }
         }
-
-        // 2) remove the mapping
-        List<? extends ServiceExposeMap> maps = context.objectManager.mappedChildren(
```
sonchang: In the abstract class, we're just scheduling for removal one ServiceExposeMap (the class member exposeMap).  But here, we were removing a List of them (all of the ones associated with this instance).

alena1108:
@ibuildthecloud I don't think excluding Removing would work. Because we:

1) schedule "bad" serviceExpose maps for removal
2) generate the usedNames map based on the current serviceExpose map entries 

if at the moment 2) executes, service.exposeMap.remove process might not be started yet, and the map state wound't be "removing"

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
alena1108: @sonchang  There can be only one map for instance->Service. I was just retrieving it not the most efficient way before. This map is always stored as a class member, and gets removed in the abstract class.

PR #558, opened 2015-06-03
collocate for container networking mode

sonchang:


ibuildthecloud:
LGTM.  For some reason there is no build and I couldn't find a failed drone webhook.

ibuildthecloud:
I cool with merging this if we can just get a green build.

PR #559, opened 2015-06-03
fix service containers deploying to same host

sonchang:
@alena1108

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
@@ -321,8 +321,9 @@ public class ServiceDiscoveryServiceImpl implements ServiceDiscoveryService {
                     ((List<Object>) dataObj).addAll((List<Object>) launchConfigItems.get(key));
                 }
             }
-
-            launchConfigItems.put(key, dataObj);
+            if (dataObj != null) {
+                launchConfigItems.put(key, dataObj);
+            }
         }
 
         // 3. add extra parameters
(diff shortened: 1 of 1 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -380,11 +382,8 @@ private String getInstanceName(Instance instance) {
         return data;
     }
 
-    @SuppressWarnings("unchecked")
     private Map<String, Object> getLaunchConfigDataAsMap(Service service) {
-        return (Map<String, Object>) DataAccessor.fields(service)
-                .withKey(ServiceDiscoveryConstants.FIELD_LAUNCH_CONFIG).withDefault(Collections.EMPTY_MAP)
-                .as(Map.class);
+        return getServiceDataAsMap(service);
```
alena1108: those are 2 diff methods. GetServiceData returns all services data including params defined on the service - scale, lb config, launch config - while launch config is meant to return only params specific to instance launch config.

alena1108:
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -321,8 +321,9 @@ private void populateExternalLinksForService(Service service, Map<String, Object
                     ((List<Object>) dataObj).addAll((List<Object>) launchConfigItems.get(key));
                 }
             }
-
-            launchConfigItems.put(key, dataObj);
+            if (dataObj != null) {
+                launchConfigItems.put(key, dataObj);
```
sonchang: service config coming from UI was populated requestedHostId to be null.  This overrode the requestedHostId we figure out from the globalService causing all the containers to go to the same host

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -382,9 +383,21 @@ private String getInstanceName(Instance instance) {
...
     @SuppressWarnings("unchecked")
     private Map<String, Object> getLaunchConfigDataAsMap(Service service) {
-        return (Map<String, Object>) DataAccessor.fields(service)
+        Map<String, Object> launchConfig = (Map<String, Object>) DataAccessor.fields(service)
                 .withKey(ServiceDiscoveryConstants.FIELD_LAUNCH_CONFIG).withDefault(Collections.EMPTY_MAP)
                 .as(Map.class);
+        Map<String, Object> data = new HashMap<>();
+        data.putAll(launchConfig);
+
+        Object labels = data.get(ServiceDiscoveryConfigItem.LABELS.getCattleName());
```
sonchang: This deep copy of labels is because when we're processing the labels for each deployment unit, we were inadvertently modifying the labels in the service.  So with each subsequent container getting deployed by the service for that particular service, it was getting labels from the previous container.  Hence, causing all the containers to get deployed to the same host.

cjellick:
Is it possible (and worth) writing tests for the two associated bugs? I see the one issue already got sent back from QA and wondering if a few tests would prevent that from happening again.

sonchang:
There are actually several existing tests.  I'll have to modify them to more accurately simulate what the UI is doing.

cjellick:
@sonchang Can you clarify? Does that mean you will or will not add/modify tests in this PR?

sonchang:
@cjellick The test for the global bug will have to wait until a subsequent PR since it looks like it'll be a little bit more difficult to simulate what the UI is doing.  I'll also have to postpone modifying the other test regarding the other bug as well.  (Sorry for the confusion.  Was hoping it might be something easy I can put in immediately).

cjellick:
@sonchang np. The change looks fine to me. Just wanted to throw it out there and after your response I wasnt sure if I should wait for tests to finish reviewing.

cjellick:
LGTM

PR #561, opened 2015-06-04
fix global service where no host labels are involved

sonchang:
@cjellick @alena1108

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
@@ -10,6 +10,7 @@ import io.cattle.platform.configitem.version.ConfigItemStatusManager;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.core.model.Service;
+import io.cattle.platform.deferred.util.DeferredUtils;
 import io.cattle.platform.engine.idempotent.IdempotentRetryException;
 import io.cattle.platform.eventing.EventService;
 import io.cattle.platform.eventing.model.EventVO;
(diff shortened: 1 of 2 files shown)


sonchang:
Basically, fixes scenario where we have a global service that's currently active (no host affinity constraints).  Then we add a new host.  We expect the service to be deployed to that new host.  Except the trigger happens in HostActivatePostHandler at which point there is no host entry in the DB yet.  I'm using DeferredUtils so that we complete the host activation thread first, then do the service activate.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -268,12 +269,26 @@ public void doWithLockNoResult() {
 
     @Override
     public void activateGlobalServicesForHost(long accountId, long hostId) {
+
```
alena1108: By deferring the service.activate call, are we 100% sure that the action won't happen before the host is fully active? If some delay happens in host activation process for example

sonchang: Yup.  Just tested it out with breakpoints set in:
- AgentResourcesMonitorImpl.java where we create the host DB entry
- and the activate(service) call in the DeferredServiceActivate class

activate(service) doesn't get executed until after we've created the host DB entry

alena1108:
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -268,12 +269,26 @@ public void doWithLockNoResult() {
...
+            }
+        }
+
         List<? extends Service> services = expMapDao.getActiveServices(accountId);
         for (Service service: services) {
             Map<String, String> serviceLabels = sdSvc.getServiceLabels(service);
             if (serviceLabels.containsKey(ServiceDiscoveryConstants.LABEL_SERVICE_GLOBAL) &&
                     allocatorSvc.hostSatisfiesHostAffinity(hostId, serviceLabels)) {
-                activate(service);
+                DeferredUtils.defer(new DeferredServiceActivate(service));
```
ibuildthecloud: Instead of this pattern, I think you want to io.cattle.platform.servicediscovery.deployment.DeploymentManager#reconcileServicesFor() which will schedule a DeploymentManager.activate() in the background in a reliable fashion.  for this to work though you need to change io.cattle.platform.servicediscovery.service.ServiceDiscoveryService#getServicesFor() also to handle objects of instance type `Service`

ibuildthecloud: As a side note, global in general should use this approach to by calling DeploymentManager.reconcileServicesFor(account).  This method was added by me to handle health check but I had planned that global could use it too.

ibuildthecloud:
LGTM

PR #572, opened 2015-06-05
collocate for container networking mode

sonchang:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/NetworkContainerConstraintProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void appendConstraints(AllocationAttempt attempt, AllocationLog log,
+            List<Constraint> constraints) {
+        Instance instance = attempt.getInstance();
+
+        if (instance != null) {
+            String networkMode = DataAccessor.fields(instance).withKey(DockerInstanceConstants.FIELD_NETWORK_MODE).as(String.class);
+            if (DockerNetworkConstants.NETWORK_MODE_CONTAINER.equals(networkMode)) {
+                Integer containerId = instance.getNetworkContainerId().intValue();
```
ibuildthecloud: In the current test we incorrectly set container network mode w/o setting the networkContainerId.  This test is fixed in https://github.com/rancherio/cattle/pull/578/files.  But this bad test does cause a NPE here.  I would say that if networkContainerId is null, we just don't apply the CollcationConstraint.

sonchang: Added null check and squashed

PR #585, opened 2015-06-08
Only use githubClient functions if github is configured.

hibooboo2:
This stops us from having to worry about token invalidity. When auth is off.

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubExternalIdHandler.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubExternalIdHandler.java
@@ -70,6 +70,9 @@ public class GithubExternalIdHandler implements ExternalIdHandler {
 
     @Override
     public Set<ExternalId> getExternalIds(Account account) {
+        if (!githubClient.githubConfigured()){
+            return new HashSet<>();
+        }
         ApiRequest request = ApiContext.getContext().getApiRequest();
         githubUtils.findAndSetJWT();
         String jwt = githubUtils.getJWT();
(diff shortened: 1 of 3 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/github/GithubExternalIdHandler.java
```diff
@@ -77,7 +77,15 @@ public ExternalId untransform(ExternalId externalId) {
         if(StringUtils.isBlank(jwt) && !StringUtils.isBlank(accessToken)){
             try {
                 jwt =  ProjectConstants.AUTH_TYPE + githubTokenHandler.getGithubToken(accessToken).getJwt();
-            } catch (IOException e) {
+            } catch (ClientVisibleException e) {
```
ibuildthecloud: I don't think this actually deletes the object from the database.  But it doesn't seem to be the right approach to modify the account during a call like this.  It can really cause things to get out of sync.  What's the driving motivation for this change?

hibooboo2: The driving motivation for this change is so that if the access token becomes invalid we remove the cached one and the user will have to get a new one. Otherwise the error gets masked by another one and it is not apparent that the token is invalid/out of date. More specifically if you have an access Token and auth switches from GHE to github or vice versa everything blows up.

hibooboo2: No longer deletes the access token. Just doesn't give you a jwt. Which as a result doesn't give you access to any other accounts. eg: Environments.

PR #587, opened 2015-06-08
Allow defining multiple states using json for a resource action.

hibooboo2:


ibuildthecloud:
I don't think this is the right approach.  Purge is the final state for some resources but then removed is for others.  So the big will apply to objects that go to removed also.  This only impacts custom actions.  Actions based on processes will never been seen in the purged.

I'd rather we just configure the custom actions to only show in proper states.

cjellick:
Was my suggestion. Sorry for leading you astray @hibooboo2

code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
```diff
@@ -458,7 +458,7 @@ protected boolean isValidAction(Object obj, Action action) {
             return false;
         }
 
-        if (!StringUtils.isBlank(state) && !state.equals(io.cattle.platform.object.util.ObjectUtils.getState(obj))) {
+        if (!StringUtils.isBlank(state) && !state.contains(io.cattle.platform.object.util.ObjectUtils.getState(obj))) {
```
ibuildthecloud: This is a bit hacky.  State is an array, but you're doing toString() on an array and then `contains()`.  The problem here is that other types have `state` as a string.  So depending on the type `state` is a string or array.  How about you add a new field called `states` that is always an array.

resources/content/schema/base/container.json
```diff
@@ -227,7 +227,10 @@
     "resourceActions": {
         "logs": {
             "input":"containerLogs",
-            "output":"hostAccess"
+            "output":"hostAccess",
+            "attributes": {
+                "state": [ "running", "stopped" ]
```
ibuildthecloud: stopping should be valid also

code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
```diff
@@ -48,6 +48,8 @@
 import java.util.Map;
 
 import javax.inject.Inject;
+import javax.print.DocFlavor;
+import javax.swing.*;
```
ibuildthecloud: No swing!

tests/integration/cattletest/core/test_container.py
```diff
@@ -724,3 +724,16 @@ def test_container_network_modes(context, super_client):
...
+
+def test_container_resource_actions_json_state(context):
+    c = context.create_container(startOnCreate=True)
+    c.stop()
+    c.logs()
+    c = context.client.wait_success(c)
+    c.logs()
+    context.client.delete(c)
+    c = context.client.wait_success(c)
+    with pytest.raises(AttributeError) as e:
```
ibuildthecloud: `'logs' not in c` is a bit prettier

code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
hibooboo2: No idea how swing got imported.

ibuildthecloud: and no DocFlavor? #intellijFail

PR #588, opened 2015-06-08
Healtcheck fixes

alena1108:
* added start_count to instance
* added new state=initializing to the instance
* don't reconcile services in inactive state on healtcheck updates
* fixed excluding the host container runs on, from the healtcheck monitors list for the instance (unless its the only host in the system)

@ibuildthecloud


alena1108:
https://github.com/rancherio/rancher/issues/1055
https://github.com/rancherio/rancher/issues/1054
https://github.com/rancherio/rancher/issues/1013

PR #590, opened 2015-06-09
Reconcile LB targets on lb service activate

alena1108:
on other events, always check the lb service state, and don't propagate the targets if the state is Inactive. Otherwise targets updates kick off the lb agent start within the inactive service

https://github.com/rancherio/rancher/issues/902

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/LoadBalancerTargetDaoImpl.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/LoadBalancerTargetDaoImpl.java
@@ -83,7 +83,7 @@ public class LoadBalancerTargetDaoImpl extends AbstractJooqDao implements LoadBa
     }
 
     @Override
-    public List<? extends Instance> getLoadBalancerActiveInstanceTargets(long lbId) {
+    public List<? extends Instance> getLoadBalancerActiveTargetInstances(long lbId) {
         return create()
                 .select(INSTANCE.fields())
                 .from(INSTANCE)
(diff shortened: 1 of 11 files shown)


code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/LoadBalancerTargetDaoImpl.java
```diff
@@ -111,4 +111,18 @@ public LoadBalancerTarget getLbIpAddressTarget(long lbId, String ipAddress) {
                         .and(LOAD_BALANCER_TARGET.LOAD_BALANCER_ID.eq(lbId)))
                 .fetchInto(LoadBalancerTargetRecord.class);
     }
+
+    @Override
+    public List<? extends LoadBalancerTarget> getLoadBalancerActiveInstanceTargets(long lbId) {
```
sonchang: Super minor, but the name is very similar to the one above: 'getLoadBalancerActiveTargetInstances'.  I guess since this function name ends with *Targets, it's not too bad.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePostListener.java
```diff
@@ -0,0 +1,120 @@
...
+        List<Long> targetInstanceToAdd = new ArrayList<>();
+        targetInstanceToAdd.addAll(consumedInstancesIds);
+        targetInstanceToAdd.removeAll(lbTargetIds);
+
+        for (Long toAdd : targetInstanceToAdd) {
+            lbService.addTargetToLoadBalancer(lb, toAdd);
+        }
+    }
+
+    protected List<Long> getConsumedInstancesIds(Service service) {
```
sonchang: I have some performance concerns here since I'm thinking we can probably replace it with an SQL query in one of the Dao classes.  But I don't mind tackling this later.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePostListener.java
```diff
@@ -0,0 +1,120 @@
...
+            return null;
+        }
+        
+        LoadBalancer lb = objectManager.findOne(LoadBalancer.class, LOAD_BALANCER.SERVICE_ID,
+                service.getId(), LOAD_BALANCER.REMOVED, null);
+        
+        List<Long> consumedInstancesIds = getConsumedInstancesIds(service);
+        List<Long> lbTargetIds = getLoadBalancerTargetInstancesIds(lb);
+
+        addMissingTargets(lb, consumedInstancesIds, lbTargetIds);
```
sonchang: General comment: Probably good to move this logic into the LoadBalancerService

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceDiscoveryLoadBalancerTargetAddPostListener.java
```diff
@@ -89,70 +92,88 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+        List<Pair<ServiceExposeMap, Service>> exposeMapToLBService = new ArrayList<>();
         if (process.getName().equalsIgnoreCase(InstanceConstants.PROCESS_START)) {
-            Instance instance = (Instance) state.getResource();
-            List<ServiceExposeMap> maps = objectManager.mappedChildren(
-                    objectManager.loadResource(Instance.class, instance.getId()),
-                    ServiceExposeMap.class);
-            if (maps.isEmpty()) {
-                // handle only instances that are the part of the service
-                return exposeMapToLBService;
+            getLBServiceForInstance(state, exposeMapToLBService);
```
sonchang: I think it might be more clear if we pass the casted state.getResource() into these functions so the method signatures for these make more sense.

sonchang:
LGTM

PR #591, opened 2015-06-09
Reconcile service from a host delete

sonchang:


ibuildthecloud:
@sonchang The idea of DeploymentManager.reconcileServicesFor() was that most process handlers could essentially do `deploymentManager.reconcileServicesFor(state.getResource())`.   We would then centralize the logic in DeploymentManager/ServiceDiscoveryService to then decide if resource X has changed what services are impacted.

DeploymentManager.reconcileServicesFor(...) internally will call DeploymentManager.activate() for each impacted Service.  The service reconcile logic should then be the one to decide to delete an instance on a removed host.  So somewhere in `io.cattle.platform.servicediscovery.deployment.impl.DeploymentManagerImpl#activateDeploymentUnits` it should decide that the deployment unit is on a removed host and should be deleted.

sonchang:
@ibuildthecloud I'll make those adjustments.  I remember you mentioned you were looking into removing the instances for a deleted host, so I was sort of waiting to see how those changes would integrate with these.  Sounds like it would be part of the service reconciliation.

ibuildthecloud:
I will hopefully finish the container remove late tonight.  On host remove all containers will be scheduled to be removed.  Which means when host remove is done there will still be some processing going on.

sonchang:
I wonder do we need to worry about order of operation (removes completing before creation of new units)?  I was going to start centralizing the logic into DeploymentManagerImpl, but if you have changes, I'm wondering should I hold off to avoid conflicts and stepping on each other.  I still have another bug to fix and the TF to investigate.

alena1108:
@ibuildthecloud @sonchang 

>So somewhere in >io.cattle.platform.servicediscovery.deployment.impl.DeploymentManagerImpl#activateDeploymentUnits >it should decide that the deployment unit is on a removed host and should be deleted.

There are 2 methods in DeploymentUnit determining if the unit is functional:

* isError - these units are removed prior to adding new units
* isHealthy - these units are left around till the new healthy units are created; and only then removed (in the future, the removal will be postponed to give user time to debug them)

Depending on the order of operation, one of the above can be picked for signaling the DeploymentManager that the unit is bad, and remove should be called for it.

sonchang:
@ibuildthecloud @alena1108 can you take a look?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/ServiceDeploymentPlanner.java
```diff
@@ -28,7 +28,7 @@ public ServiceDeploymentPlanner(List<Service> services, List<DeploymentUnit> uni
 
         if (units != null) {
             for (DeploymentUnit unit : units) {
-                if (unit.isUnhealthy()) {
+                if (unit.isUnhealthy() || !unit.isHostActive()) {
```
alena1108: Son, is there a reason why we segregate "isHostActive" to its separate method, instead of making it a part of isUnhealthy? Because its really what it means - unit is unhealthy when the healtcheck failed for it, or when the host where unit runs on, is unhealthy. I think it would make code cleaner if we don't expose TMI to the deployment manager.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -92,6 +95,34 @@ public boolean isError() {
         return false;
     }
 
+    public boolean isHostActive() {
+        for (DeploymentUnitInstance deployUnitInstance : getDeploymentUnitInstances()) {
+            if (deployUnitInstance instanceof ExternalDeploymentUnitInstance) {
```
alena1108: It should be if !(deploymentUnitInstance instanceof InstanceUnit). Otherwise the code will break when the new implementation of DeploymentUnitInstance that is not instance of InstanceUnit, gets added.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/ServiceDeploymentPlanner.java
sonchang: I think I'm good going either way.  The reason I did this was because it looked like "unit unhealthy" is based on the health_state on the instance which I'm assuming is based on the health check specified by the user.  So I didn't want to mix the two in case we want to handle host deletion in a special way.

alena1108: I see your point. What if we leave isHostActive in deploymentUnit, but make it private, and call from within unit.isUnhealthy. Then it would be easy to separate it in the future if there is a need for it.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
sonchang: I felt the common case should have an instance and ExternalDeploymentUnitInstance is more the exception.  And that's why I had it throw a RuntimeException in its implementation.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/ServiceDeploymentPlanner.java
sonchang: I can make this change.  My original thought was DeploymentUnit should just supply the info and DeploymentPlanner can take whatever actions it wants using that info.  But in this function, I guess it's just grouping into healthy and unhealthy anyways.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
alena1108: The point is valid. But we shouldn't assume that the instance is the common case. Like today DNS service doesn't have corresponding UnitInstance, but if we decide to add it in the future - lets say some extra functionality required for this service besides configuring DNS - this implementation won't have an instance, and will break things.

sonchang: I'll make the change

alena1108:
LGTM

PR #595, opened 2015-06-09
Allow remove to be called from any state.

hibooboo2:


sonchang:
Is there any additional logic in handlers that we need to worry about?

hibooboo2:
userPreferences has no handlers.

sonchang:
LGTM

alena1108:
We should really add a test for every new state transition we introduce.

hibooboo2:
Hmm...

PR #598, opened 2015-06-10
it's possible to have some rare timing issues. also, deactivate servi…

sonchang:
…ce (esp. global)

at end of test


sonchang:
@ibuildthecloud

sonchang:
@alena1108 @ibuildthecloud 
Should fix the intermittent update_labels TF we get in drone.  The deactivation of global service wasn't necessary for this fix, but I figure it's good to not have those lying around since I'm not sure it plays nicely when we do the rest of the test teardown and I think could cause hangups downstream.

alena1108:
LGTM

cjellick:
LGTM

PR #604, opened 2015-06-11
Don't collect purged instances for services

alena1108:
And remove service-expose map as a pre-handler for instance.purge, not
post handler

Supposed to fix one of the TFs observed on Drone


alena1108:
Also supposed to fix one of the failures observed on RC1 demo environment

PR #605, opened 2015-06-11
Add machine binaries to Dockerfile

cloudnautique:
As we develop services that are packaged with Rancher, we need to
have a way for them to be used during our development process.
Having them a part of cattle dockerfile will allow integration
testing along with build-master environments to have easier access
to development releases.

ping @ibuildthecloud /cc @cjellick @wlan0


ibuildthecloud:
@cloudnautique Does this mean we have to update both rancherio/cattle and rancherio/server Dockerfiles?  Or is rancherio/server somehow reading this?

cloudnautique:
@cjellick and I talked more about a different approach that will help get the binaries into the build master environments. It avoids duplicate settings like this.

PR #610, opened 2015-06-11
include updating-active state for when a host is updating labels

sonchang:
This fixes a problem where the labels for a host would get updated.  The hostLabelMap changes triggers a reconciliation of the global service.  However, the currently updating host would not be considered since it's in updating-active state.


sonchang:
@alena1108 @cjellick @ibuildthecloud

alena1108:
LGTM

cjellick:
Returning hosts that are in a transitioning state (updating active) wont have any negative side effects in the bigger picture?

sonchang:
Well, right now this call is just used to gather a list of potential hosts to deploy the global service to.  I was a little bit concerned about whether we might run into problems deploying a container to a host that is in 'updating-active' state.  But I tested this in the debugger via breakpoints and it looks like that part isn't an issue.

cjellick:
Ok, then LGTM

PR #612, opened 2015-06-11
Deactivate/acivate LB along with LB service

alena1108:
When service is activated, activate the LB. When deactivated -
deactivate. If LB is left in Active state when LB service is inactive,
LB update caused by target instance.stop/instance.start would trigger LB
instance restart within the inactive service

https://github.com/rancherio/rancher/issues/902

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
@@ -1,0 +1,48 @@
+package io.cattle.platform.servicediscovery.process;
+
+import static io.cattle.platform.core.model.tables.LoadBalancerTable.LOAD_BALANCER;
+import io.cattle.platform.core.constants.CommonStatesConstants;
+import io.cattle.platform.core.model.LoadBalancer;
+import io.cattle.platform.core.model.Service;
+import io.cattle.platform.engine.handler.HandlerResult;
+import io.cattle.platform.engine.handler.ProcessPreListener;
+import io.cattle.platform.engine.process.ProcessInstance;
+import io.cattle.platform.engine.process.ProcessState;
+import io.cattle.platform.object.process.StandardProcess;
+import io.cattle.platform.process.common.handler.AbstractObjectProcessLogic;
+import io.cattle.platform.servicediscovery.api.constants.ServiceDiscoveryConstants;
+import io.cattle.platform.util.type.Priority;
+
+import javax.inject.Named;
+
+@Named
+public class LoadBalancerServiceActivatePreListener extends AbstractObjectProcessLogic implements ProcessPreListener,
+        Priority {
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { ServiceDiscoveryConstants.PROCESS_SERVICE_ACTIVATE };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        Service service = (Service) state.getResource();
+
+        if (!service.getKind().equalsIgnoreCase(ServiceDiscoveryConstants.KIND.LOADBALANCERSERVICE.name())) {
+            return null;
+        }
+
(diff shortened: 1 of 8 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
```diff
@@ -0,0 +1,50 @@
...
+        Service service = (Service) state.getResource();
+
+        if (!service.getKind().equalsIgnoreCase(ServiceDiscoveryConstants.KIND.LOADBALANCERSERVICE.name())) {
+            return null;
+        }
+
+        LoadBalancer lb = objectManager.findOne(LoadBalancer.class, LOAD_BALANCER.SERVICE_ID, service.getId(),
+                LOAD_BALANCER.REMOVED, null);
+        if (lb != null
+                && (lb.getState().equalsIgnoreCase(CommonStatesConstants.INACTIVE) || lb.getState().equalsIgnoreCase(
```
sonchang: If it's already in activating state, do we want to activate again?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
```diff
@@ -0,0 +1,50 @@
...
+                        CommonStatesConstants.ACTIVATING))) {
+            objectProcessManager.executeStandardProcess(StandardProcess.ACTIVATE, lb, null);
```
sonchang: Any reason why we're executing immediately rather than scheduling for activation?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceDeactivatePreListener.java
```diff
@@ -0,0 +1,51 @@
...
+
+        if (!service.getKind().equalsIgnoreCase(ServiceDiscoveryConstants.KIND.LOADBALANCERSERVICE.name())) {
+            return null;
+        }
+
+        LoadBalancer lb = objectManager.findOne(LoadBalancer.class, LOAD_BALANCER.SERVICE_ID, service.getId(),
+                LOAD_BALANCER.REMOVED, null);
+
+        List<String> activeStates = Arrays.asList(CommonStatesConstants.ACTIVATING, CommonStatesConstants.ACTIVE,
+                CommonStatesConstants.UPDATING_ACTIVE, CommonStatesConstants.DEACTIVATING);
```
sonchang: Same questions here.  If we're in deactivating already, why deactivate again?  Also, any particular need to execute immediately?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
alena1108: Its done this way to handle idempotent retries for "execute". Otherwise the first failed attempt will change the state to Activating, but won't proceed further, so the LB will get stuck forever in Activating state.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
alena1108: Actually on a second thought, scheduling might work. Then I won't have to include "Activating" state to the list of valid start states. Let me try that.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceActivatePreListener.java
alena1108: Activating is removed from the list of valid states after i've switched from "execute" to "schedule"

sonchang:
LGTM

hibooboo2:
LGTM

PR #614, opened 2015-06-12
Reconcile services on instance.stop; serviceReconcile triggers refactoring

alena1108:
And refactoring for serviceReconcile triggers - centralized them into 2 classes (for Post and Handlers), moved all trigger related logic out of DeploymentManager


alena1108:
@ibuildthecloud @sonchang @cjellick

cjellick:
Review.
While I don't have enough context for a big picture review, fundamentally,
LGTM

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -143,10 +142,6 @@ private void activateServices(final Service initialService, final List<Service>
          */
         try {
             for (Service service : services) {
-                if (service.getId().equals(initialService.getId())) {
```
sonchang: I haven't looked at the rest of the changes yet, but was curious about this change.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServicesReconcilePostTrigger.java
```diff
@@ -0,0 +1,52 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import com.netflix.config.DynamicStringListProperty;
+
+@Named
+public class ServicesReconcilePostTrigger extends AbstractObjectProcessLogic implements ProcessPostListener {
+    private static final DynamicStringListProperty PROCESS_NAMES = ArchaiusUtil
+            .getList("service.reconcile.posttrigger.processes");
```
sonchang: Since I don't foresee this value changing often (for example, between prod vs debug or for other types of environments), I'm not sure if there's that much value extracting it out like this.  Also, funny enough, I've actually grown to like referencing the *Constants classes we've created.  I've noticed it's actually a pretty easy way to find all the handlers for that particular event.  For example, in eclipse, I could just look for all references to HostConstants.PROCESS_CREATE.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServicesReconcilePostTrigger.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        Collection<? extends Service> services = null;
+        for (ServiceLookup lookup : serviceLookups) {
+            services = lookup.getServices(state.getResource());
+            if (services != null && !services.isEmpty()) {
+                deploymentManager.reconcileServices(services);
+                break;
```
sonchang: Hmm.  I feel like we should reconcile the union of the services returned by all the ServiceLookups.  For example, both GlobalHostActivateServiceLookup and HostServiceLookup handles the Host resource, but will return a different set of Services.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
alena1108: We just re-enabled something that was enabled before, and temporarily disabled :) Basically when reconcile is called for the service from heatlcheck or global host update, the activateService should trigger state transition activation for the service we are operating on. Otherwise the reconcile will be happening, and service state won't reflect it.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServicesReconcilePostTrigger.java
alena1108: The new way makes it easier to add more triggers. We follow this format in config.item update logic as well as load balancer update. So @ibuildthecloud suggested we use the same logic for reconcile.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServicesReconcilePostTrigger.java
alena1108: You are right. Will change that.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServicesReconcilePostTrigger.java
sonchang: Yeah.  I remember seeing it for the config updates too.
I like having the ServicesReconcilePostTrigger class.  But I think I prefer having them all explicitly listed in the getProcessNames() function as ugly as it may seem.  I guess I still don't see much value making it configurable via an environment variable and I think having them explicitly listed in getProcessNames() makes it easier for someone to debug the system.

alena1108: The only value I see in the current approach - its easier to remove it from the config file if something goes wrong, and restart the server. I'm fine with either way.

alena1108: Lets put it as a separate commit as a part of this PR, so we can roll back to the configurable way if we find a need in it.

sonchang: ok

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
sonchang: thx!

sonchang:
LGTM

PR #616, opened 2015-06-12
No assert, no problem

cjellick:
Or "The Story of the Never Failing Tests"


alena1108:
LGTM

sonchang:
Looks good to me, although, I'm wondering whether we should just have one 'type' set and then pass the appropriate client and add/remove sets.  I guess I just see the duplication in the admin_client check.

PR #619, opened 2015-06-12
Add read on create only support

hibooboo2:
This makes it so that you can add "o" as a permission on a resourceField.
When "o" is present that field is only returned when the resource is created
for that schema. All other requests exclude the field.


alena1108:
build fails due to pom changes; can you take a look?

hibooboo2:
Fixed was waiting on a pr to gdapi

PR #625, opened 2015-06-14
Cleanup services with account.purge

alena1108:
and include services in Activating/Updating-Active when searching for out of sync config items.

@ibuildthecloud

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/account/AccountPurge.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/account/AccountPurge.java
@@ -4,6 +4,7 @@ import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.model.Account;
 import io.cattle.platform.core.model.Agent;
 import io.cattle.platform.core.model.Credential;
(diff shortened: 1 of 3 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/account/AccountPurge.java
```diff
@@ -49,6 +50,13 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
             deactivateThenRemove(agent, state.getData());
         }
 
+        for (Environment env : getObjectManager().children(account, Environment.class)) {
+            if (env.getRemoved() != null) {
+                continue;
+            }
+            objectProcessManager.scheduleStandardProcessAsync(StandardProcess.REMOVE, env, null);
```
ibuildthecloud: Can you do remove().  All other processes are not async

alena1108: Containers' cleanup is async; so I've made services to be async as well as they are the closest to container in concept. I believe Craig did change sync to async execution for containers cleanup due to some idempotent issues (commit 7a5ed73)

PR #630, opened 2015-06-15
Service discovery bug fixes

alena1108:
1) Exclude Restarting instances from healtcheck

https://github.com/rancherio/rancher/issues/1180

2) Launch services in order from environment.activateServices

3)  And added missing state transition - allow service.remove from deactivating state

4)  Skip activation for removed links during instance.start

5) Wait for nonTransition state on service instances before performing any action

https://github.com/rancherio/rancher/issues/1254

6)  Don't explicitly clean env in tests as they are removed as a part of account.purge

7) Export links as array of strings (not map) to compose

https://github.com/rancherio/rancher/issues/1120


ibuildthecloud:
How does this fix work.  Instances are only in restarting for a very very short time.

alena1108:
@ibuildthecloud yes, but during this short period of time they shouldn't be included in the config (the config refresh can be triggered by some other instance...so there is a slight chance that Restarting instances might get in there)

alena1108:
@ibuildthecloud Darren, could you pls review when you get a chance? This PR has the bug fix for instance.start (to exclude remove links when handle start)

PR #631, opened 2015-06-15
add documentation for machine fields

wlan0:
@ibuildthecloud @cjellick


will-chan:
Shouldn't we be able to add some tests to validate this?

wlan0:
This is machine-generated @will-chan. I'm not sure what kind of tests we could add.

cjellick:
Maybe @will-chan is suggesting a test that reads the schema and verifies the description is set?
Since this documentation feature is already part of the framework, I don't think that's necessary.
LGTM.

cjellick:
@wlan0 tests failed because of these changes. Please take a look.

PR #636, opened 2015-06-17
Compose: don't export null values

alena1108:
1) Don't include null values to docker/rancher-compose.yml
2) Remove requestedHostId/cpuSet from rancher-compose (used to keep there
as docker-compose doesn't support these params, but never read them). We can re-enable cpuSet, but to put it to the docker-compose.yml once docker-compose starts supporting it (https://github.com/docker/compose/issues/754)


sonchang:
LGTM. Wondering whether we should do a StringUtils.isEmpty() check or is null sufficient

alena1108:
@sonchang I'll keep it in mind as a future improvement.

PR #637, opened 2015-06-17
Schedule lb processes async

alena1108:
To make things picked up for execution faster.


code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
```diff
@@ -204,7 +204,7 @@ protected void start(final Instance agentInstance) {
             DeferredUtils.nest(new Callable<Object>() {
                 @Override
                 public Object call() throws Exception {
-                    processManager.scheduleProcessInstance(InstanceConstants.PROCESS_START, agentInstance, null);
+                    processManager.scheduleProcessInstanceAsync(InstanceConstants.PROCESS_START, agentInstance, null);
```
ibuildthecloud: @alena1108 Isn't this redundant?  You're already in a nest call.  Maybe you should just remove the outer nest so the code is just a one-liner.

alena1108: ah, yes, it is

PR #642, opened 2015-06-18
LB fixes

alena1108:
1) allow only Active hosts being added to the load balancer
https://github.com/rancherio/rancher/issues/628

2) lb service instance: removing state qualify instance for removal

https://github.com/rancherio/rancher/issues/1288


ibuildthecloud:
Tests?

alena1108:
@ibuildthecloud will add

PR #646, opened 2015-06-19
Exclude instances with removed serviceexposemap ref

alena1108:
When do service reconcile


alena1108:
@sonchang @ibuildthecloud

sonchang:
LGTM

PR #648, opened 2015-06-19
Don't start unhealthy units

alena1108:
@ibuildthecloud @sonchang


alena1108:
unrelated test (test_container_restart) failed

PR #651, opened 2015-06-24
No need to wait for serviceMap create on instance.start

alena1108:
https://github.com/rancherio/rancher/issues/1311


alena1108:
The "wait" was put originally as a part of initial healtcheck implementation (where health_state was defined on per exposeMap basis, so we had to ensure that the host map existed before instance registers for the healtcheck). Today healtcheck is defined on the instance itself, and we don't need anything extra for the instance to be registered.

alena1108:
@cjellick @sonchang could you please review?

sonchang:
LGTM

cjellick:
LGTM

PR #652, opened 2015-06-24
update container label affinity via env var

sonchang:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/ContainerLabelAffinityConstraint.java
```diff
@@ -10,7 +10,7 @@
 import com.google.common.collect.Multimap;
 
 public class ContainerLabelAffinityConstraint implements Constraint {
-    public static final String ENV_HEADER_AFFINITY_CONTAINER_LABEL = "affinity:container_label:";
+    public static final String ENV_HEADER_AFFINITY_CONTAINER_LABEL = "affinity:";
```
alena1108: Is it backward compatible?

sonchang: No.  Although, I really wonder how many people use this since we encourage using labels instead of environment variables.  @ibuildthecloud What do you think?  Basically, we introduced our syntax before swarm.  This PR is to keep parity with swarm.

sonchang:
@cjellick @ibuildthecloud Can you please review?

cjellick:
Functionally, the change is LGTM, but I think the backwards compatibility bit should get a thumbs up from @ibuildthecloud or @will-chan.

In the end, I agree with you @sonchang: since we encourage labels (and the UI uses labels), its better to break backwards compat and have one "right" way of supporting env vars for affinity constraints that is consistent with swarm as opposed to supporting one right way and one legacy way that is inconsistent with swarm.

PR #653, opened 2015-07-01
LB hostname routing support + ServiceLinks/LoadBalancerTarget API changes + lb service healthcheck changes

alena1108:
This PR introduces support for LB hostname routing. As a part of the feature, API changes to service links and LB target were done.

Update: as a part of this PR, Load Balancer healtcheck configuration are done too. Before, we used to read the healthcheck config from the LB service config; now its read on per consumed service basis first. If no healthcheck is defined for consumed service, we always try to read it from the Load Balancer service config - to support backwards compatibility + standalone IP based balancing.


alena1108:
API changes done as a part of the PR:

1) add/remove/setServiceLinks APIs no longer accept serviceIds. Instead they accept ServiceLink param (for regular service) and LoadBalancerServiceLink (for LB service).

ServiceLink fields:

* serviceId 
* name (DNS name for named links)

LoadBalancerServiceLink extends from ServiceLink, and has 1 extra param:

* ports (array[string]). Each value has a format of "<port>[:<hostname><path>]". Example: 80:foo.com/bar. Meaning that this service would accept lb traffic to the exposed port 80, and only traffic coming from foo.com domain and /bar path

2) add/remove/setLoadBalancerTarget also got "ports" param to support hostName routing for the standalone LB case

code/iaas/lb-logic/src/main/java/io/cattle/iaas/lb/service/impl/LoadBalancerServiceImpl.java
```diff
@@ -94,43 +96,33 @@ public void removeListenerFromConfig(LoadBalancerConfig config, long listenerId)
...
-    public void addTargetToLoadBalancer(LoadBalancer lb, long instanceId) {
-        LoadBalancerTarget target = lbTargetDao.getLbInstanceTarget(lb.getId(), instanceId);
-        if (target == null) {
-            target = objectManager.create(LoadBalancerTarget.class, LOAD_BALANCER_TARGET.INSTANCE_ID, instanceId,
-                    LOAD_BALANCER_TARGET.LOAD_BALANCER_ID, lb.getId(), LOAD_BALANCER_TARGET.IP_ADDRESS, null,
-                    LOAD_BALANCER_TARGET.ACCOUNT_ID, lb.getAccountId());
-        }
-        objectProcessManager.scheduleProcessInstance(LoadBalancerConstants.PROCESS_LB_TARGET_MAP_CREATE,
-                target, null);
+    public void addTargetToLoadBalancer(LoadBalancer lb, LoadBalancerTargetInput targetInput) {
```
ibuildthecloud: Since you are refactoring some of this code now, is it possible to split out the methods called in the API tier from the backend tier?  So LoadBalancerServiceImpl and LoadBalancerApiServiceImpl or something similar.  Given all the troubles with DeferredUtils it would be good to cleanly separate the code.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/api/action/AddServiceLinkActionHandler.java
```diff
@@ -16,14 +15,12 @@
...
-    @Inject
-    ServiceConsumeMapDao consumeMapDao;
 
     @Inject
-    ObjectManager objectManager;
+    JsonMapper jsonMapper;
 
     @Inject
-    ObjectProcessManager objectProcessManager;
+    ServiceDiscoveryService sdService;
```
ibuildthecloud: Along the same lines as the comment about LoadBalancerServiceImpl.  It scares me to see API level code invoking backend classes.  I like to see a clear separation.

alena1108: Sure, Darren, will fix.

alena1108:
@vincent99  Besides the new APIs support, following needs to be done in the UI:

* Disable standalone LB ^__^
* Disable target port on the Load Balancer service; leave only listener port. Target port now defined on per service link basis
* Disable healtcheck on the Load Balancer. Now healtcheck is being read from the service that is registered to the load balancer.

will-chan:
#1042

PR #655, opened 2015-07-03
Fixed the bug in sidekick service reconcile

alena1108:
Recreate the entire unit if one of the instance is missing

https://github.com/rancherio/rancher/issues/1363

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
@@ -86,6 +86,9 @@ public class DeploymentUnit {
     }
 
     public boolean isError() {
+        if (!this.isComplete()) {
+            return true;
+        }
         /*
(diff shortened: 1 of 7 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -84,6 +84,9 @@ private void createMissingUnitInstances(Map<Long, DeploymentUnitInstanceIdGenera
     }
 
     public boolean isError() {
+        if (!this.isComplete()) {
```
ibuildthecloud: Why is an incomplete unit treated as an error?

alena1108: Currently we don't keep track of the relationship within the deploymentUnit, so if any part of it is missing, the entire unit has to be recreated.  For example, when the sidekick instance is gone, the living parent is no longer valid as it references non-existing container.

PR #659, opened 2015-07-07
Update DNS when instance unassigned from service

alena1108:
https://github.com/rancherio/rancher/issues/1368


alena1108:
When instance is removed from the service, DNS record has to be updated. Before the fix, dns update wasn't triggered.

sonchang:
LGTM

cjellick:
LGTM

Issue #658, opened 2015-07-07
Generic proxy isn't able to whitelist a host when specifying a protocol


vincentbernat:
Hey!

While the code in https://github.com/rancherio/cattle/blob/master/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/request/handler/GenericWhitelistedProxy.java seems to be written to handle the ability to proxy to HTTPS, it doesn't seem to work as expected:

```session
$ curl 'http://localhost:8080/proxy/ec2.ap-southeast-2.amazonaws.com/'
[works]
$ curl 'http://localhost:8080/proxy/http://ec2.ap-southeast-2.amazonaws.com/'
{"id":"8a77ae82-b03e-4144-b507-1b35b919c393","type":"error","links":{},"actions":{},"status":403,"code":"Forbidden","message":"Forbidden","detail":null}
$ curl 'http://localhost:8080/proxy/https://ec2.ap-southeast-2.amazonaws.com/'
{"id":"8a77ae82-b03e-4144-b507-1b35b919c393","type":"error","links":{},"actions":{},"status":403,"code":"Forbidden","message":"Forbidden","detail":null}
```

Any hint?

vincentbernat:
I am unable to reproduce this by running cattle locally on my machine. Only the provided docker image seems to have the problem. However, the generic proxy code has never been modified. Is it possible to get the commit the Docker image was generated from?

vincentbernat:
From `/v1/settings`, I see that the version in the Docker image is 0.58.0. I have tried with 0.58.0 and it works fine too. However, there seems to be some differences with my setup and the one in Docker. For example, here is what I get when requesting a non-whitelisted URL:

```
< HTTP/1.1 403 Forbidden
< X-API-Schemas: http://localhost:8080/v1/schemas
< X-API-ACCOUNT-ID: 1a1
< Content-Type: application/json; charset=utf-8
< Content-Length: 152
< Server: Jetty(8.1.11.v20130520)
```

In Docker, I get:

```
< HTTP/1.1 403 Forbidden
< Content-Length: 152
< Content-Type: application/json; charset=utf-8
< Server: Jetty(8.1.11.v20130520)
< X-Api-Account-Id: 1a1
< X-Api-Schemas: http://localhost:8080/v1/schemas
< Date: Wed, 08 Jul 2015 12:10:05 GMT
```

So, there is some middleware mangling the output headers. I suppose there could be one mangling the incoming URL. Still searching...

[...]

PR #663, opened 2015-07-08
Log first database connection error

vincentbernat:
The exception may be helpful to debug a database error. Log it on first
attempt, then don't log it anymore.

I was trying to use the h2 backend but the generic message was unhelpful.


ibuildthecloud:
This looks good.  Just a heads up H2 has issues.  It mostly works but we've found some of our SQL queries break against H2.

vincentbernat:
Thanks! Dully noted for H2. I am just using it to figure issue #658.

PR #665, opened 2015-07-08
preserve backwards compatibility with 'project'

sonchang:


sonchang:
@ibuildthecloud @alena1108 @cjellick Can you take a look?

alena1108:
LGTM

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
```diff
@@ -55,8 +55,12 @@
     public static final String LINK_COMPOSE_CONFIG = "composeConfig";
 
     public static final String LABEL_SERVICE_DEPLOYMENT_UNIT = "io.rancher.service.deployment.unit";
-    public static final String LABEL_PROJECT_NAME = "io.rancher.stack.name";
-    public static final String LABEL_PROJECT_SERVICE_NAME = "io.rancher.stack_service.name";
+    public static final String LABEL_STACK_NAME = "io.rancher.stack.name";
+    public static final String LABEL_STACK_SERVICE_NAME = "io.rancher.stack_service.name";
+    // LEGACY: preserving project_name
+    public static final String LABEL_PROJECT_NAME = "io.rancher.project.name";
+    public static final String LABEL_PROJECT_SERVICE_NAME = "io.rancher.project_service.name";
```
cjellick: Commenting on pre-existing code and also a nit, but, it seems odd to mix dots and underscores here. To me, `io.rancher.project.service.name` seems more natural.

I won't hold up a merge bc of this, just wanted to voice my opinion.

cjellick:
Other than my one comment, LGTM. @sonchang , I'll wait for a reply form you to see if you care to make that change or not.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
sonchang: It's because the value looks like ${project_name}/${service_name}.
I think with the dot instead of the underscore, I would think it's just the service name alone (that happens to be scoped to the project we're currently in)

cjellick: Ok. Sounds good.

PR #666, opened 2015-07-08
Antiaffinity fix for sidekick

alena1108:
Wait for unit to be allocated before starting a new unit. This fix is needed for the case when anti-affinity rule is defined on the service having sidekicks. If we continue deploying units in parallel, its possible for the deployment units to clash - when Primary instance of Unit1 lands on the same host with Secondary instance of unit2. In this case neither primary instance of Unit2, nor secondary instance of Unit1 going to deploy as they have 2 contradicting affinity rules now:

* deploy on the same host with other instance of the same deployment unit
* deploy on a diff host from other instances of the same launch config.

@sonchang is going to put a fix for service-anti-affinity-to-itself translation to include launch config name, so instances of the same deployment unit won't conflict with each other like they do today - as they share the same service name anti affinity rule.


alena1108:
@ibuildthecloud @sonchang please review

sonchang:
LGTM

PR #667, opened 2015-07-09
proxy: use HTTPS by default when no protocol specified

vincentbernat:
When no protocol was specified, the protocol used for the incoming
request was used for the outgoing request. So, if a proxy request comes
over HTTP, the outgoing request is going over HTTP too. We use HTTPS by
default instead. Amazon EC2 endpoints are all HTTPS capable.

See #658.


vincent99:
Looks like the `protocol` variable is unused now and can be removed

vincentbernat:
I've updated the commit.

PR #668, opened 2015-07-09
Bump py-agent to v.0.41.0

cjellick:


cjellick:
This is just a version bump PR. Im merging so I can get it off my plate.

PR #669, opened 2015-07-09
Dont process containerEvent if host is removed.

cjellick:
Addresses https://github.com/rancherio/rancher/issues/1445.
If a host is removed and we try to process a container event, that
event will hang forever and a huge queue will build up for it in
processInstances. To avoid this, immediately exit if host is removed.


cjellick:
Couldn't think of a good way to add tests for this as it requires a scenario where an containerEvent is created and the host is removed before that event is processed.

Manually tested using break points to pause containerEvent processing and removed the host while it was paused.

Our load test environment has roughly 25 thousands such containerEvents, so it will be a good test bed for this fix.

alena1108:
LGTM

sonchang:
LGTM

PR #672, opened 2015-07-09
modify service name to include launchConfig as well

sonchang:
@alena1108 @ibuildthecloud @cjellick 
The background to this change started from https://github.com/rancherio/rancher/issues/1419
Basically, @sangeethah spotted a bug where we weren't able to support even the basic case of a "spread" algorithm via the anti-affinity rule with a sidekick.  Since the anti-affinity rule label gets merged between the 2 launchConfigs and applied to the deployment of the primary and secondary instances, it automatically conflicted with trying to collocate them on the same host.
Adding the launchConfig value into the ${service_name} macro allows for a more granular anti-affinity rule.


alena1108:
LGTM

PR #677, opened 2015-07-10
Postgres compatibility tweaks.

dx9:


ibuildthecloud:
This is really cool.  Some logistical issues though.  We currently support upgrades.  The way liquibase works is that keeps the checksums of changesets and if the checksum changes then changeset X != X.  So basically what happens is it won't upgrade any more because it sees change 42 had hash ABC and now it has hash XYZ, so something changed and it does know what.

To make this work for upgrades we have to put the old hash in the xml in the validChecksum parameter.  An example is here https://github.com/rancher/cattle/blob/master/resources/content/db/core-015.xml.  To get the value you basically have to run the old XML and then look at the DATABASECHANGELOG table to get the checksum.  I think you have to put both the old and new hash.

Finally, any chance you can change https://github.com/rancher/cattle/blob/master/scripts/ci to test Postgres.  It's great you've fixed these issues but inevitably we will add more issue if we don't test this in CI.

Again, thanks, this is really cool.  We really want to support Postgres but just been a little constrained on time.

dx9:
Yep and yep.

dx9:
The saga continues in #742

PR #680, opened 2015-07-10
proxy: add a comment about specifying a protocol with generic proxy

vincentbernat:
Looking at the code, one may think that it's a sensible thing to do, but
it doesn't work because of rancher/rancher#1485. Add a comment with a
reference to the bug. See also rancher/cattle#658.


vincentbernat:
Feel free to just close this PR if you think that the comment is too invading to be helpful.

PR #683, opened 2015-07-10
Internal load balancer support

alena1108:
API changes:

1) loadBalancerListener.add

* sourcePort is optional now
* new parameter - privatePort - was added.

Either one or another parameter is required; both can be specified as well. 
If sourcePort only is specified, private port is defaulted to it. 

2) loadBalancerService.add

* new parameter is added to launchConfig - "expose". Every port defined in expose,  will get translated to the listener having NULL public port, and port=port from expose set, and Format for expose is: [<port>/[protocol]]. If no protocol is specified, http listener will get created.
* if both ports and expose is specified, for every port defined in the expose, private listener will get created; and for each port - public listener.

### DIFF
--- a/resources/content/db/changelog.xml
+++ b/resources/content/db/changelog.xml
@@ -58,4 +58,5 @@
     <include file="db/core-051.xml"/>
     <include file="db/core-052.xml"/>
     <include file="db/core-053.xml"/>
+    <include file="db/lbport.xml"/>
 </databaseChangeLog>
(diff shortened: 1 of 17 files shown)


alena1108:
@ibuildthecloud Darren, please review. As a part of the feature, I had to enable "expose" parameter on the launchConfig (container) object. The parameter is read by ServiceDiscovery code only, to create loadBalancerListener, and not propagated to the backend for containers. Let me know if we should add backend implementation for "expose" as a part of this PR, or separate.

alena1108:
https://github.com/rancher/rancher/issues/1433

ibuildthecloud:
I'd hold off on expose in the backend

resources/content/db/changelog.xml
```diff
@@ -58,4 +58,5 @@
     <include file="db/core-051.xml"/>
     <include file="db/core-052.xml"/>
     <include file="db/core-053.xml"/>
+    <include file="db/lbport.xml"/>
```
ibuildthecloud: rename

alena1108:
Closing this PR as the feature is going to be merged as a part of https://github.com/rancher/cattle/pull/704

PR #685, opened 2015-07-12
Update haproxy.cfg.ftl

willseward:
Updated the haproxy config to use an exact match for host headers to avoid premature substring matches


willseward:
https://github.com/rancher/rancher/issues/1503

alena1108:
LGTM, good catch, @willseward

sonchang:
LGTM

PR #686, opened 2015-07-13
add support for aggregated stats

wlan0:
@cjellick


wlan0:
The error seems unrelated in the test build - ```test_add_lb_w_host_and_target``` failed due to a timeout

wlan0:
https://github.com/rancher/rancher/issues/1265

cjellick:
@wlan0 Please add tests

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+    ObjectManager objectManager;
+
+    @Override
+    public String[] getTypes() {
+        return new String[] { InstanceConstants.TYPE_CONTAINER, HostConstants.TYPE };
+    }
+
+    @Override
+    public boolean handles(String type, String id, String link, ApiRequest request) {
+        return StatsConstants.CONTAINER_STATS.equals(link);
```
cjellick: Can you change this to equalsIgnoreCase?

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsLinkHandler.java
```diff
@@ -23,56 +28,79 @@
...
     @Override
     public String[] getTypes() {
-        return new String[] { InstanceConstants.TYPE_CONTAINER, HostConstants.TYPE };
+        return new String[] { HostConstants.TYPE, ProjectConstants.TYPE };
     }
 
     @Override
     public boolean handles(String type, String id, String link, ApiRequest request) {
-        return HostStatsConstants.LINK_STATS.equals(link);
+        return StatsConstants.HOST_STATS.equals(link) && ("project".equals(type) || HostConstants.TYPE.equals(type));
```
cjellick: ignore case

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ServiceContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final String SERVICE_PATH = "/service/";
+
+    @Override
+    public String[] getTypes() {
+        return new String[] { "service" };
+    }
+
+    @Override
+    public boolean handles(String type, String id, String link, ApiRequest request) {
+        return StatsConstants.CONTAINER_STATS.equals(link);
```
cjellick: ignore case

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/StatsLinkHandler.java
```diff
@@ -0,0 +1,96 @@
...
+    ObjectManager objectManager;
+
+    @Override
+    public String[] getTypes() {
+        return new String[] { InstanceConstants.TYPE_CONTAINER, HostConstants.TYPE };
+    }
+
+    @Override
+    public boolean handles(String type, String id, String link, ApiRequest request) {
+        return StatsConstants.LINK_STATS.equals(link);
```
cjellick: ignore case

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/utils/StatsConstants.java
```diff
@@ -0,0 +1,9 @@
+package io.cattle.platform.host.stats.utils;
+
+public class StatsConstants {
+
+    public static final String LINK_STATS = "stats";
```
cjellick: @wlan0 Can you change these to camelCase so that the keys in the link map match the case of all the other links?
Meaning:
`'links': { 'containerStats': 'http...'}` instead of `containerstats`

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+            host = DockerUtils.getHostFromContainer(objectManager, instance);
+        } else if (obj instanceof Host) {
+            host = (Host) obj;
+        }
+
+        if (host == null) {
+            return null;
+        }
+
+        if (!"active".equals(host.getState())) {
```
cjellick: Can you use the constant in CommonStateConstants?

wlan0:
@cjellick Changinig it to camelCase here doesn't change it to camelCase in the links.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+            return null;
+        }
+
+        Map<String, Object> payload = new HashMap<String, Object>();
+        Map<String, Object> containerIdsMap = new HashMap<String, Object>();
+
+        List<Instance> instances = objectManager.mappedChildren(host, Instance.class);
+        for (Instance i : instances) {
+            if (!"running".equals(i.getState())) {
```
cjellick: InstanceConstants

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/utils/StatsConstants.java
wlan0: @cjellick Changing it to camelCase here doesn't change it to camelCase in the links.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+        if (!"active".equals(host.getState())) {
+            return null;
+        }
+
+        Map<String, Object> payload = new HashMap<String, Object>();
+        Map<String, Object> containerIdsMap = new HashMap<String, Object>();
+
+        List<Instance> instances = objectManager.mappedChildren(host, Instance.class);
+        for (Instance i : instances) {
```
cjellick: This handler is used for the containerStats link on ***container*** too, right?
The token you are building always has ***all*** containers on the associated host, even if this link is in the context of a a single container.
Can you change this so that it doesn't return all containers when the link is coming from an instance? And can you add a test to explicit confirm this?

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+            if (!"running".equals(i.getState())) {
+                continue;
+            }
+            containerIdsMap.put(DockerUtils.getDockerIdentifier(i), ApiContext.getContext().getIdFormatter().formatId(objectManager.getType(i), i.getId()));
+        }
+        payload.put("containerIdsMap", containerIdsMap);
```
cjellick: containerIdsMap seems redundant. Could you rename this to simply `containers` or `containerIds`

cjellick:
Per slack convo, include stopped containers on host's container stats.

cjellick:
We also determined it would be cool if host stats could dynamically add newly created containers. This is significantly more effort. It would mean _not_ enumerating containers in the host's jwt and instead making host-api monitor container creation.

@wlan0 Can you just look into this, determine its feasibility and the level effort? ...Don't do this until you've addressed all the other PR changes.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+
+        HostApiAccess apiAccess = hostApiService.getAccess(host.getId(), payload);
+        if (apiAccess == null) {
+            return null;
+        }
+
+        StringBuilder url = new StringBuilder(HostApiUtils.HOST_API_PROXY_SCHEME.get());
+        url.append("://").append(apiAccess.getHostAndPort());
+        url.append(HOST_STATS_PATH.get());
```
cjellick: We do this URL building in 6 different places. Maybe its worth a utility method.

cjellick:
@wlan0 You return a 404 on the containerstats link for a container (ie http://localhost:8080/v1/projects/1a5/containers/1i1/containerstats). Probably shouldn't do that. 
The current behavior of the stats link (ie http://localhost:8080/v1/projects/1a5/containers/1i1/stats) returns a token even if the container is stopped. Does it make sense to continue to do that bheavior?

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsLinkHandler.java
```diff
@@ -23,56 +28,79 @@
...
+        List<StatsAccess> serviceStatsQuery = new ArrayList<>();
+        for (Host host : hosts) {
+            if (!"active".equals(host.getState())) {
+                continue;
+            }
+            Map<String, Object> payload = new HashMap<>();
+            payload.put("resourceId", ApiContext.getContext().getIdFormatter().formatId(objectManager.getType(host), host.getId()));
+            HostApiAccess apiAccess = hostApiService.getAccess(host.getId(), payload);
+            if (apiAccess == null) {
+                return null;
```
cjellick: It seems incorrect to `return null` here. `continue` seems more appropriate.

wlan0: can you elaborate?

cjellick:
The url field on a service's containerstats resources is null until the service is started. That is probably not right.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
wlan0: Agreed.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsLinkHandler.java
cjellick: I was thinking if just 1 host out of 10 gets a null apiAccess, we could just skip it and return the other 9.

wlan0: hmmm, that makes sense. I'll change it.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/StatsLinkHandler.java
wlan0: I just moved the previous HostStatsLinkHandler to StatsLinkHandler. This is old code and we will remove this as soon as vince creates UI for this feature.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/StatsLinkHandler.java
```diff
@@ -0,0 +1,93 @@
...
+import io.github.ibuildthecloud.gdapi.request.ApiRequest;
+
+import java.io.IOException;
+import java.util.Collections;
+
+import javax.inject.Inject;
+
+import com.netflix.config.DynamicStringProperty;
+
+public class StatsLinkHandler implements LinkHandler {
```
cjellick: Can you please add a `@Deprecated` annotation stating that this will be removed once the UI switches to using the new stats links?

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ServiceContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        return StatsConstants.CONTAINER_STATS.equalsIgnoreCase(link);
+    }
+
+    @Override
+    public Object link(String name, Object obj, ApiRequest request) throws IOException {
+        Service service = (Service) obj;
+
+        List<Instance> serviceInstances = objectManager.mappedChildren(service, Instance.class);
+
+        if (service == null) {
```
cjellick: Should check for null service before getting the mapped children.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsOutputFilter.java
```diff
@@ -16,15 +20,38 @@
...
         if (original instanceof Host && DockerHostConstants.KIND_DOCKER.equals(((Host) original).getKind())) {
             add = true;
+            hostStats = true;
+            containerStats = true;
         } else if (original instanceof Instance && InstanceConstants.KIND_CONTAINER.equals(((Instance) original).getKind())) {
+            containerStats = true;
             add = true;
         }
 
+        if (original instanceof Account && AccountConstants.PROJECT_KIND.equals(((Account) original).getKind())) {
```
cjellick: Could you make these one big if/else along with the above if/else. They're mutually exclusive, right?

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsOutputFilter.java
```diff
@@ -1,11 +1,15 @@
 package io.cattle.platform.host.stats.api;
```
cjellick: The class really shouldn't be called `HostStatsOutputFilter` since it deals with all types of stats. Would `StatsOutputFilter` be more appropriate.

wlan0: Yep. Agreed.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsLinkHandler.java
```diff
@@ -1,75 +1,99 @@
...
+            project = true;
         }
 
-        String[] pathSegments = null;
-        if (instance != null) {
-            pathSegments = new String[] { HOST_STATS_PATH.get(), DockerUtils.getDockerIdentifier(instance) };
-        } else {
-            pathSegments = new String[] { HOST_STATS_PATH.get() };
+        if (hosts.size() == 0) {
+            return null;
```
cjellick: Returning null causes the API to return a 404. We shouldn't do that.

wlan0: I did this to keep consistent with the existing /v1/stats API.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,116 @@
...
+        } else if (obj instanceof Host) {
+            host = (Host) obj;
+        }
+
+        if (host == null) {
+            return null;
+        }
+
+        if (!CommonStatesConstants.ACTIVE.equals(host.getState())) {
+            return null;
```
cjellick: Results in a 404. Shouldn't do that.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,116 @@
...
+
+        if (obj instanceof Instance) {
+            instance = (Instance) obj;
+            host = DockerUtils.getHostFromContainer(objectManager, instance);
+        } else if (obj instanceof Host) {
+            host = (Host) obj;
+        }
+
+        if (host == null) {
+            return null;
```
cjellick: I've pointed out other places where this causes a 404, but I'm ok with it in this case because I can't imagine a valid scenario where the host would be null here.

We could also possibly just raise an `IllegalStateException` instead.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/StatsLinkHandler.java
cjellick: Ok, fine to leave as is.

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
wlan0: host could be null only in invalid scenarios. 

possible scenarios include : 
1. cannot not authorize the user to view stats for that host
2. a host cannot be found for a requested container. 

So, `IllegealStateException` sounds right

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
wlan0: I'll change the 404s to Empty `Object()` instead

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+
+public class ContainerStatsLinkHandler implements LinkHandler {
+
+    HostApiService hostApiService;
```
alena1108: Minor: we can use Inject here instead of setting these vars with set()

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsOutputFilter.java
```diff
@@ -16,15 +20,38 @@
...
+        }
+
         if (add) {
-            converted.getLinks().put(HostStatsConstants.LINK_STATS, ApiContext.getUrlBuilder().resourceLink(converted, HostStatsConstants.LINK_STATS));
+            converted.getLinks().put(StatsConstants.LINK_STATS, ApiContext.getUrlBuilder().resourceLink(converted, StatsConstants.LINK_STATS));
+        }
+        if (hostStats) {
+            converted.getLinks().put(StatsConstants.HOST_STATS, ApiContext.getUrlBuilder().resourceLink(converted, StatsConstants.HOST_STATS));
+        }
+        if (containerStats) {
```
alena1108: Minor: if these conditions are mutually exclusive, it would probably make sense to do elseif so the next condition is not checked if the previous one is evaluated to true.

wlan0: These two booleans are independent and can co-exist

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ServiceContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+            return null;
+        }
+
+        String metaUrl = null;
+        StatsAccess meta = new StatsAccess();
+        List<StatsAccess> serviceStatsQuery = new ArrayList<>();
+        for (Instance instance : serviceInstances) {
```
alena1108: What is the reason for including Stopped state to the list of valid states for the stats? If there is any stats available for such containers, we should also include other states - rather than Removed one - to the list

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ServiceContainerStatsLinkHandler.java
```diff
@@ -0,0 +1,117 @@
...
+            HostApiAccess apiAccess = hostApiService.getAccess(request, host.getId(), payload, new String[] { StatsConstants.CONTAINER_STATS_PATH.get() });
+            if (apiAccess == null) {
+                continue;
+            }
+
+            StringBuilder url = new StringBuilder();
+            
+            url.append(apiAccess.getUrl());
+
+            metaUrl = url.toString() + SERVICE_PATH;
```
alena1108: Nit: as url is already a builder, you can just use append to concat SERVICE_PATH to it, and then call toString later

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/HostStatsLinkHandler.java
```diff
@@ -1,75 +1,99 @@
...
+            HostApiAccess apiAccess = hostApiService.getAccess(request, host.getId(), payload, new String[] { StatsConstants.HOST_STATS_PATH.get() });
+            if (apiAccess == null) {
+                continue;
+            }
+
+            String url = apiAccess.getUrl();
+
+            metaUrl = url + PROJECT_PATH;
+
+            StatsAccess statsAccess = new StatsAccess();
```
alena1108: Do you think it would make sense to create a constructor in StatsAccess having token/url as input params? As this code seems to repeat several times:

 statsAccess.setToken(apiAccess.getAuthenticationToken());
statsAccess.setUrl(url.toString());

code/implementation/host-stats/src/main/java/io/cattle/platform/host/stats/api/ServiceContainerStatsLinkHandler.java
wlan0: I don't want to add the SERVICE_PATH to the builder. That's why I do this.

cjellick:
LGTM

alena1108:
LGTM

PR #688, opened 2015-07-14
For primary config, just use stack_name/service_name

sonchang:
@alena1108 @ibuildthecloud Can you take a quick look.  This removes that nasty "io.rancher.service.primary.launch.config" part from the ${service_name} macro for the primary config.  secondary config will still have it.  This also allows for backwards compatibility prior to adding the launchConfig name to the service name (for the anti-affinity blocker bug).


alena1108:
LGTM

sonchang:
Regarding the TF, I'm still looking at it but wasn't able to replicate it locally.  Plus, it doesn't look related to my changes.

ibuildthecloud:
I don't get warm fuzzies about that TF, but I'll merge

PR #689, opened 2015-07-14
Generate uuid for the backend using domain/hostname

alena1108:
https://github.com/rancher/rancher/issues/1514


alena1108:
@sonchang @ibuildthecloud

sonchang:
LGTM.  Although, curious since I don't have much background.  Is it possible to have multiple LBs for different ports?

alena1108:
I'm open for suggestions on better way to generate unique integer hash based on 2 strings. The replaced hashcode generated the same hashes for combinations like:

service2.html/www.abc2.com and service1.html/www.abc3.com generated the same hashcodes

alena1108:
@sonchang here we generate the unique backend on the load balancer based on the hostName/path.

sonchang:
Thx!  LGTM

PR #690, opened 2015-07-14
make check for container name case insensitive too

sonchang:


sonchang:
@alena1108 @cjellick @ibuildthecloud 
Making check for container names case insensitive as well (in addition to host labels and container labels)

cjellick:
LGTM

alena1108:
LGTM

PR #692, opened 2015-07-14
Add map to extensionManager.

hibooboo2:
This allows you to use interfaces to create a map of beans in the extensionmanager instead of a list.

This allows constant time access to beans needed of a certain type, instead of n by iterating over a list.

### DIFF
--- a/code/framework/extension/src/main/java/io/cattle/platform/extension/ExtensionManager.java
+++ b/code/framework/extension/src/main/java/io/cattle/platform/extension/ExtensionManager.java
@@ -1,6 +1,7 @@
 package io.cattle.platform.extension;
 
 import java.util.List;
+import java.util.Map;
 
 public interface ExtensionManager {
 
@@ -9,6 +10,8 @@ public interface ExtensionManager {
     <T> T first(String key, Class<T> type);
 
     List<?> list(String key);
+
+    Map<String, Object> map(String key);
 
     <T> List<T> getExtensionList(Class<T> type);
 
--- a/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
+++ b/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.regex.Pattern;
 
@@ -32,11 +33,13 @@ public class ExtensionManagerImpl implements ExtensionManager, InitializationTas
 
     Map<String, List<Object>> byKeyRegistry = new HashMap<String, List<Object>>();
     Map<String, ExtensionList<Object>> extensionLists = new HashMap<String, ExtensionList<Object>>();
+    Map<String, ExtensionMap<String, Object>> extensionMaps = new HashMap<>();
     Map<String, List<Object>> byName = new HashMap<String, List<Object>>();
     Map<Object, String> objectToName = new HashMap<Object, String>();
     Map<String, Class<?>> keyToType = new HashMap<String, Class<?>>();
     Map<Pattern, List<String>> wildcards = new HashMap<Pattern, List<String>>();
     boolean started = false;
+
 
     @SuppressWarnings("unchecked")
     @Override
@@ -52,12 +55,27 @@ public class ExtensionManagerImpl implements ExtensionManager, InitializationTas
     @SuppressWarnings("unchecked")
     @Override
     public <T> T first(String key, Class<T> type) {
-        return (T) Proxy.newProxyInstance(type.getClassLoader(), new Class<?>[] { type }, new FirstInstanceInvocationHandler(getExtensionListInternal(key)));
+        return (T) Proxy.newProxyInstance(type.getClassLoader(), new Class<?>[]{type}, new FirstInstanceInvocationHandler(getExtensionListInternal(key)));
     }
 
     @Override
     public List<?> list(String key) {
         return getExtensionListInternal(key);
+    }
+
+    @Override
+    public Map<String, Object> map(String key) {
+        return getExtensionMapInternal(key);
+    }
+
+    protected synchronized ExtensionMap<String, Object> getExtensionMapInternal(String key) {
+        ExtensionMap<String, Object> map = extensionMaps.get(key);
+        if (map == null) {
+            Map<String, Object> inner = started ? getMap(key) : Collections.<String, Object>emptyMap();
+            map = new ExtensionMap<>(this, key, inner);
+            extensionMaps.put(key, map);
+        }
+        return map;
     }
 
     @Override
@@ -134,10 +152,22 @@ public class ExtensionManagerImpl implements ExtensionManager, InitializationTas
                 ExtensionList<?> extensionList = getExtensionListInternal(key);
                 extensionList.inner.clear();
                 extensionList.inner.addAll(getList(key));
+                ExtensionMap<String, Object> extensionMap = getExtensionMapInternal(key);
+                extensionMap.inner.clear();
+                extensionMap.inner.putAll(getMap(key));
             }
 
             started = true;
         }
+    }
+
+    protected synchronized Map<String, Object> getMap(String key) {
+        List<?> list = getList(key);
+        Map<String, Object> map = new ConcurrentHashMap<>();
+        for (Object item : list) {
+            map.put(NamedUtils.getName(item), item);
+        }
+        return map;
     }
 
     public void reset() {
--- a/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
+++ b/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
@@ -1,0 +1,84 @@
+package io.cattle.platform.extension.impl;
+
+import io.cattle.platform.extension.ExtensionManager;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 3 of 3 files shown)


code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
```diff
@@ -61,6 +51,21 @@
...
 
     @Override
+    public Map<String, Object> map(String key) {
+        return getExtensionMapInternal(key);
+    }
+
+    protected ExtensionMap<String, Object> getExtensionMapInternal(String key) {
+        ExtensionMap<String, Object> map = extensionMaps.get(key);
+        if (map == null) {
+            Map<String, Object> inner = started ? getExtensionMapInternal(key) : Collections.<String, Object>emptyMap();
```
sonchang: the call to getExtensionMapInternal(key) looks wrong to me here.  did you mean to call getMap(key)?

hibooboo2: See https://github.com/hibooboo2/cattle/blob/add_map/code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java#L83 to see why i did this. But that is the right call. This is a recursive function.

sonchang:
LGTM

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
```diff
@@ -4,23 +4,11 @@
 import io.cattle.platform.extension.ExtensionImplementation;
 import io.cattle.platform.extension.ExtensionManager;
 import io.cattle.platform.extension.ExtensionPoint;
-import io.cattle.platform.util.type.CollectionUtils;
-import io.cattle.platform.util.type.InitializationTask;
-import io.cattle.platform.util.type.NamedUtils;
-import io.cattle.platform.util.type.PriorityUtils;
-import io.cattle.platform.util.type.ScopeUtils;
+import io.cattle.platform.util.type.*;
```
cjellick: Nit: This is the only place in our entire code base where we would use the wildcard on non-static imports. Also, eclipse's import organizer would just re-expand this. Can you not do this.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
```diff
@@ -134,12 +139,24 @@ public synchronized void start() {
...
+                ExtensionMap<String, Object> extensionMap = getExtensionMapInternal(key);
+                extensionMap.inner.clear();
+                extensionMap.inner.putAll(getMap(key));
             }
 
             started = true;
         }
     }
 
+    protected synchronized Map<String, Object> getMap(String key) {
```
cjellick: Are you sure that this needs to be synchronized?

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
```diff
@@ -61,6 +51,21 @@
     }
 
     @Override
+    public Map<String, Object> map(String key) {
+        return getExtensionMapInternal(key);
+    }
+
+    protected ExtensionMap<String, Object> getExtensionMapInternal(String key) {
```
cjellick: Are you sure that this _doesn't_ need to be synchronized? Appears to by modifying state.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
```diff
@@ -0,0 +1,81 @@
...
+
+    Map<K, V> map;
+    ConcurrentHashMap<K, V> inner;
+    String key;
+    ExtensionManager extensionManager;
+
+    public ExtensionMap(ExtensionManager extensionManager, String key, Map<K, V> map) {
+        this.extensionManager = extensionManager;
+        this.key = key;
+        this.inner = map == null ? new ConcurrentHashMap<K, V>() : new ConcurrentHashMap<>(map);
```
cjellick: Why does inner need to be an instance level variable? Seems like it could be local to the constructor?

cjellick: Oh, I see, you call it directly. Hmmm, not sure how I feel about directly accessing a variable like that...not idiomatic java.

cjellick: So, disregarding inner for the moment, why do the other variables (extensionManager & key) )need to be instance level

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
```diff
@@ -0,0 +1,81 @@
...
+        this.map = Collections.unmodifiableMap(inner);
```
cjellick: Can you explain why we need an unmodifiable map here? Seems overly complex, especially since you hold onto a modifiable version of the map (inner) and modify it when needed.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
hibooboo2: Yes i can. Was auto done by intellij when i reinstalled.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
hibooboo2: I made it synchronized as i was following the same implemntation @ibuildthecloud used for io.cattle.platform.extension.impl.ExtensionManagerImpl#getList

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
hibooboo2: It does. You are correct. Thank you.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
hibooboo2: See io.cattle.platform.extension.impl.ExtensionList#ExtensionList As to why this is needed. Basically so that the injected lists using the extensionmanager.list method returns a list or map in the case of io.cattle.platform.extension.impl.ExtensionMap#ExtensionMap So that the classes that use `@inject` cannot modify the list.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
hibooboo2: io.cattle.platform.extension.impl.ExtensionMap#ExtensionMap Uses the same paradigm that io.cattle.platform.extension.impl.ExtensionList#ExtensionList uses.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
cjellick: Hm. TBH, I cant see why getList is synchronized either. @sonchang do you see a reason why getList() would be synchronized and/or why this method would need to be?

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
cjellick: Ok, makes sense.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionMap.java
cjellick: Alright, I don't love it, but since it's already established as a pattern here, I'm ok with it.

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
```diff
@@ -61,6 +67,21 @@
...
 
     @Override
+    public Map<String, Object> map(String key) {
+        return getExtensionMapInternal(key);
+    }
+
+    protected synchronized ExtensionMap<String, Object> getExtensionMapInternal(String key) {
+        ExtensionMap<String, Object> map = extensionMaps.get(key);
+        if (map == null) {
+            Map<String, Object> inner = started ? getMap(key) : Collections.<String, Object>emptyMap();
```
ibuildthecloud: The inner map does not need to be a synchronized collection.  In fact it shouldn't be.

code/framework/extension/src/main/java/io/cattle/platform/extension/ExtensionManager.java
```diff
@@ -10,6 +11,8 @@
 
     List<?> list(String key);
 
+    Map<String, Object> map(String key);
```
ibuildthecloud: Shouldn't this be `<T> Map<String,List<T>> map(String key)`?

ibuildthecloud: Oh wait, I think misunderstood what this code is doing.  What about `<T> Map<String,T> map(String key)`

code/framework/extension/src/main/java/io/cattle/platform/extension/impl/ExtensionManagerImpl.java
ibuildthecloud: Wait, I'm totally wrong.

ibuildthecloud:
LGTM contingent on green build.

PR #700, opened 2015-07-18
Read linkName from serviceName when null/empty/blank

alena1108:
https://github.com/rancher/rancher/issues/1564


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceLinkActivateHandler.java
```diff
@@ -38,7 +39,7 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
 
         ServiceConsumeMap map = objectManager.loadResource(ServiceConsumeMap.class, link.getServiceConsumeMapId());
         String serviceName = map.getName();
-        if (serviceName == null) {
+        if (StringUtils.isEmpty(serviceName)) {
```
ibuildthecloud: I think you want StringUtils.isBlank().  isEmpty doesn't test for "   ".

PR #701, opened 2015-07-20
remove protocol from api.host setting

cjellick:
The api.host setting can begin with a protocol like
`https://mydomain.com`. The host-api token generation code was not
respecting that setting.

Addresses rancher/rancher#1580

### DIFF
--- a/code/implementation/host-api/src/main/java/io/cattle/platform/host/api/HostApiProxyTokenManager.java
+++ b/code/implementation/host-api/src/main/java/io/cattle/platform/host/api/HostApiProxyTokenManager.java
@@ -86,6 +86,7 @@ public class HostApiProxyTokenManager extends AbstractNoOpResourceManager {
         }
 
         String apiProxyScheme = HostApiUtils.HOST_API_PROXY_SCHEME.get();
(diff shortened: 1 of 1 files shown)


code/implementation/host-api/src/main/java/io/cattle/platform/host/api/HostApiProxyTokenManager.java
```diff
@@ -86,6 +86,7 @@ protected Object createInternal(String type, ApiRequest request) {
         }
 
         String apiProxyScheme = HostApiUtils.HOST_API_PROXY_SCHEME.get();
+        hostAddress = hostAddress.replace("http://", "").replace("https://", "");
```
sonchang: Looks fine to me.  I'm guessing doing a replaceFirst() is probably not necessary

sonchang:
LGTM

ibuildthecloud:
not 100% this is correct

cjellick:
After talking with @ibuildthecloud, we determined that this is not the best fix. This fix is dependent upon the api.host setting accepting a protocol. We've determined that we ***do not*** want to support that. That setting should ***only*** be the host name (and port, if necessary).

I'm going to work on a more comprehensive fix.

PR #705, opened 2015-07-21
allow users to just specify any service name within stack for affinity

sonchang:


sonchang:
This is to address lately issue in https://github.com/rancher/rancher/issues/1531

sonchang:
Adding more context:

before, i was being too restrictive and only pre-pending the stack name if the service mentioned is the same service as itself.  This solves the use case like anti-affinity to itself.  But if we try to do anti-affinity to another service within the same stack, it won’t prepend.

but I also don’t want to just blindly prepend the stack name to whatever string is there.  that’s why there’s the check to make sure that at least it’s still a service within the same stack

alena1108:
LGTM

sonchang:
@cjellick @ibuildthecloud Can you please review?

cjellick:
LGTM

sonchang:
Darn.  This drone build passed just a couple minutes ago with only a flake8 failure which I fixed

PR #707, opened 2015-07-22
Allow null input for hostname/externalIps

alena1108:
In external service. As these 2 params are mutually exclusive now.

W/o this fix, externalService.update will always fail in the UI.


cjellick:
LGTM

sonchang:
LGTM

PR #712, opened 2015-07-22
Fixed the bug for the 2 LB listeners case

alena1108:
When no source port is defined on the port spec

https://github.com/rancher/rancher/issues/1619


cjellick:
LGTM

alena1108:
@cjellick @sonchang don't merge it yet, one thing is regressed, fixing it now

sonchang:
LGTM

PR #713, opened 2015-07-22
Program hostname routing rules in order

alena1108:
hostname + path first; then single hostname or path; then default route

https://github.com/rancher/rancher/issues/1621

Also paths should are sorted by length in descending order, so ACLs are configured on haproxy as:

/foo/bar/other
/foo/bar
/foo


vincent99:
Shouldn't paths be sorted by descending length, so that if you have `/path/` and `/path/subpath` the more specific one matches first?

Also you're using `path_beg` so a request for '/pathPlusSomeOtherStuff/blah/blah' would match `/path`.  Ensuring that there is always a trailing slash seems too restrictive in the API (what if I want `/path.html`) so maybe just note this in the documentation @deniseschannon?  If the user is matching a directory they probably want add a trailing slash, i.e. set it to `/path/`.

alena1108:
@vincent99 will try to fix by sorting paths in descending order on the java side.

cjellick:
LGTM

sonchang:
LGTM

PR #714, opened 2015-07-23
Service discovery fixes

alena1108:
1) Update the lb target instead of removing on ports change. So there is no downtime on haproxy
2) Allow linking services cross stacks
https://github.com/rancher/rancher/issues/1577
3) Generate load balancer labels in exported config
https://github.com/rancher/rancher/issues/1495
4) Put "hostname" for external service to rancher-compose - as for external service, its not a docker property like it is for the normal service where docker containers get started using this param

### DIFF
--- a/code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
+++ b/code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
@@ -126,7 +126,8 @@
 
     <!-- Load Balancer Target -->
     <process:process name="loadbalancertarget.create" resourceType="loadBalancerTarget" start="requested" transitioning="activating" done="active" />
-    <process:process name="loadbalancertarget.remove" resourceType="loadBalancerTarget" start="active,activating" transitioning="removing" done="removed" />
+    <process:process name="loadbalancertarget.remove" resourceType="loadBalancerTarget" start="active,activating,updating-active" transitioning="removing" done="removed" />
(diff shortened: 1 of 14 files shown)


sonchang:
LGTM.  How is the new LB label used?

alena1108:
@sonchang its used by the rancher-compose only.

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -126,7 +126,8 @@
 
     <!-- Load Balancer Target -->
     <process:process name="loadbalancertarget.create" resourceType="loadBalancerTarget" start="requested" transitioning="activating" done="active" />
-    <process:process name="loadbalancertarget.remove" resourceType="loadBalancerTarget" start="active,activating" transitioning="removing" done="removed" />
+    <process:process name="loadbalancertarget.remove" resourceType="loadBalancerTarget" start="active,activating,updating-active" transitioning="removing" done="removed" />
+    <process:process name="loadbalancertarget.update" resourceType="loadBalancerTarget" start="active,activating" transitioning="updating-active" done="active" />
```
cjellick: @alena1108 are you sure you want to be able to call update while the resource is in the "activating" state?
That is not standard. The standard way only allows updating from the resting states of active and inactive.

alena1108: Yea, as this resource never goes through regular activate/deactivate cycle. It will be in activating state in a split of a second during the initial link creation, and I want to make sure that if 2 updates are called quickly one by one for the same service, they don't error out.

cjellick:
LGTM

PR #715, opened 2015-07-23
fix bug with previously deleted stacks

sonchang:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorServiceImpl.java
```diff
@@ -144,11 +142,11 @@ public void normalizeLabels(Map<String, String> systemLabels, Map<String, String
         }
     }
 
-    private Set<String> getServiceNamesInStack(String stackName) {
+    // TODO: Fix repeated DB call even if DB's cache no longer hits the disk
+    private Set<String> getServiceNamesInStack(long environmentId) {
         Set<String> servicesInEnv = new HashSet<String>();
 
-        Environment stack = objectManager.findOne(Environment.class, ENVIRONMENT.NAME, stackName);
```
sonchang: This was not taking into account previously deleted stacks.  So multiple entries were being returned even though we specified 'findOne'.  Also, wasn't taking into account accountId.

sonchang:
fixes #1632.  @alena1108 @cjellick @ibuildthecloud  Can you review and merge?

alena1108:
LGTM

PR #716, opened 2015-07-24
Export cross stack links in external_links

alena1108:
In format of environment/service.name:linkName


alena1108:
Unrelated TF happening in standalone lb case, filed a ticket to keep track of it:

https://github.com/rancher/rancher/issues/1637

PR #718, opened 2015-07-24
Fixed the bug when export lb service with no hostname routing rules

alena1108:
Tested by adding 2 links: 1 with hostname routing rules ("ports" param) and another without - in the integration test

https://github.com/rancher/rancher/issues/1645

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/service/impl/ServiceDiscoveryApiServiceImpl.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/service/impl/ServiceDiscoveryApiServiceImpl.java
@@ -168,7 +168,9 @@ public class ServiceDiscoveryApiServiceImpl implements ServiceDiscoveryApiServic
             for (String port : ports) {
                 bldr.append(port).append(",");
             }
-            labels.put(labelName, bldr.toString().substring(0, bldr.length() - 1));
+            if (bldr.length() > 0) {
+                labels.put(labelName, bldr.toString().substring(0, bldr.length() - 1));
+            }
         }
 
         if (!labels.isEmpty()) {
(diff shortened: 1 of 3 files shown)


code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/service/impl/ServiceDiscoveryApiServiceImpl.java
```diff
@@ -168,7 +168,9 @@ protected void populateLoadBalancerServiceLabels(Service service,
             for (String port : ports) {
                 bldr.append(port).append(",");
             }
-            labels.put(labelName, bldr.toString().substring(0, bldr.length() - 1));
+            if (bldr.length() > 0) {
```
sonchang: darn.  sorry i missed this earlier!

tests/integration/cattletest/core/test_svc_discovery_lb.py
```diff
@@ -814,6 +815,54 @@ def test_private_lb(client, context):
...
+    lb_service = lb_service.addservicelink(serviceLink=service_link1)
+    compose_config = env.exportconfig()
+    assert compose_config is not None
+    document = yaml.load(compose_config.dockerComposeConfig)
+    assert len(document[lb_service.name]['links']) == 2
+    assert len(document[lb_service.name]['labels']) == 1
+    labels = {"io.rancher.loadbalancer.target.web": "a.com:90"}
+    links = ["web:web", "web1:web1"]
+    assert document[lb_service.name]['labels'] == labels
+    assert document[lb_service.name]['links'] == links
```
sonchang: are we always guaranteed the ordering of the links?

alena1108: @sonchang  hm, good point. I preserve the order of the links returned by the DB, and it usually returns it in ascending order based on id. So should be fine.

sonchang:
LGTM

PR #721, opened 2015-07-27
Add support for Host and x-forwarded-{proto,port}

cjellick:
When building the request URL, respect the Host and
x-forwarded-{proto,port} headers

### DIFF
--- a/code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
+++ b/code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
@@ -7,10 +7,8 @@ import io.github.ibuildthecloud.gdapi.util.RequestUtils;
 import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
 
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -33,7 +31,11 @@ public class DefaultApiRequestParser implements ApiRequestParser {
 
     public static final String DEFAULT_OVERRIDE_URL_HEADER = "X-API-request-url";
     public static final String DEFAULT_OVERRIDE_CLIENT_IP_HEADER = "X-API-client-ip";
-    public static final String FORWARDED_HEADER = "X-Forwarded-For";
+    public static final String FORWARDED_FOR_HEADER = "X-Forwarded-For";
+    public static final String FORWARDED_HOST_HEADER = "X-Forwarded-Host";
+    public static final String FORWARDED_PROTO_HEADER = "X-Forwarded-Proto";
+    public static final String FORWARDED_PORT_HEADER = "X-Forwarded-Port";
+    public static final String HOST_HEADER = "Host";
 
     public static final String HTML = "html";
     public static final String JSON = "json";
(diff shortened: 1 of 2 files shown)


code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
```diff
@@ -160,22 +162,76 @@ protected String parseClientIp(ApiRequest apiRequest, HttpServletRequest request
...
+        clientIp = getOverrideHeader(request, FORWARDED_FOR_HEADER, clientIp, false);
 
         return clientIp;
     }
 
+    /**
+     * Constructs the request URL based off of standard headers in the request, falling back to the HttpServletRequest.getRequestURL() 
+     * if the headers aren't available. Here is the ordered list of how we'll attempt to construct the URL: 
+     *  - x-api-request-url
+     *  - x-forwarded-proto://x-forwarded-host:x-forwarded-port/HttpServletRequest.getRequestURI()
```
sonchang: So I guess the x-forward-proto is required if they want to use any of the other x-forwarded-* headers?

cjellick: In my completely made-up requirements, yes.
Possible alternative would be to do something like:
 "if the host header is set and proto header isnt there and host header is not in requestUrl, then use the host from the host header and the protocol parse out of the requestUrl".

My inclination is that that is heavy handed and that users have to set x-forwarded-proto in their proxy, but I may be wrong here.

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
```diff
@@ -160,22 +162,76 @@ protected String parseClientIp(ApiRequest apiRequest, HttpServletRequest request
...
         }
 
-        requestUrl = getOverrideHeader(request, overrideUrlHeader, requestUrl);
+        if (host == null) {
+            return null;
+        }
+
+        String port = getOverrideHeader(request, FORWARDED_PORT_HEADER, null, false);
+        if (StringUtils.equals(port, "443") || StringUtils.equals(port, "80")) {
+            port = null; // Don't include default ports in url
```
sonchang: I wonder should we only strip it if the protocol matches "https" for 443 and "http" for 80?

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
sonchang: no problem.  i didn't know whether there were any particular rules we were supposed to follow or not.  if not, then i'm fine

sonchang:
LGTM

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
```diff
@@ -160,22 +162,76 @@ protected String parseClientIp(ApiRequest apiRequest, HttpServletRequest request
...
+            return null;
+        }
+
+        String host = getOverrideHeader(request, FORWARDED_HOST_HEADER, null, false);
+        if (host == null) {
+            host = getOverrideHeader(request, HOST_HEADER, null, false);
         }
 
-        requestUrl = getOverrideHeader(request, overrideUrlHeader, requestUrl);
+        if (host == null) {
```
alena1108: Nit: should we use StringUtils.isEmpty() here to eliminate both host being null, and host being empty?

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
cjellick: I considered that as well. I decided against it because it woudl be a rare/weird case like you said and since this is ran on every api request, I was trying to minimize checks as much as possible (not that 1 more string comparison would have broken the bank).

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/request/parser/DefaultApiRequestParser.java
cjellick: Yeah, I went back and forth on that. Decided against it because it seemed very unlikely it would be empty and I was trying to be as minimalistic as possible in my checks since this is ran on every api request.

Thank it is better to just be safe and make it a StringUtils.isEmpty() check?

alena1108: its ok, I agree its highly unlikely

alena1108:
LGTM

PR #724, opened 2015-07-30
Fix websocket-proxy URL creation for HA mode

cjellick:
Base the protocol off of the official responseUrlBase, not the url of
the incoming request.


ibuildthecloud:
LGTM

PR #725, opened 2015-07-30
Encryption of fields in rancher.

hibooboo2:


code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/EncryptionUtils.java
```diff
@@ -0,0 +1,52 @@
...
+            return new JSONObject();
+        }
+    }
+
+    public static void saveKeyToFile(String key, String value) throws IOException {
+        JSONObject jsonObject =  getKeyStore();
+        FileWriter fileWriter = new FileWriter(fileName);
+        jsonObject.put(key, value);
+        jsonObject.writeJSONString(fileWriter);
+        fileWriter.flush();
```
alena1108: you think we should put .flush and .close into "finally" block?

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
```diff
@@ -0,0 +1,72 @@
...
+    private Key key;
+    private SecureRandom rn;
+
+    public Aes128Encrypter() throws Exception {
+        rn = new SecureRandom();
+
+        encrypter = Cipher.getInstance("AES/CBC/PKCS5PADDING", "SunJCE");
+        try {
+            encryptionKey = EncryptionUtils.getKeyFromFile(NAME + "Key");
+        } catch (NullPointerException e) {
```
alena1108: Catching NPE here implies that  you know the details of EncryptionUtils.java implementation. Can we instead throw some checked exception by EncryptionUtils.java method:

```
public static String getKeyFromFile(String propertyName) {
      return getKeyStore().get(propertyName).toString(); // check if key is null here, and throw an appropriate exception 
    }
```

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/EncryptionUtils.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+public class EncryptionUtils {
+    private static String fileName = "keys.json";
+
+    public static boolean isEqual(byte[] a, byte[] b) {
+        if (a.length != b.length) {
```
alena1108: You think it would be worth putting not null checks for a/b to avoid NPEs? As there is a chance "transform" attribute is put on the field allowing null values.

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
```diff
@@ -0,0 +1,70 @@
...
+        Object response = request.getResponseObject();
+        if (response instanceof WrappedResource) {
+
+            Schema schema = request.getSchemaFactory().getSchema(request.getType());
+            request.setResponseObject(decrypt((Resource) response, schema));
+        } else if (response instanceof CollectionImpl){
+            SchemaFactory schemaFactory = request.getSchemaFactory();
+            for (Resource resource:((CollectionImpl) response).getData()){
+                decrypt(resource, schemaFactory.getSchema(resource.getType()));
+                //ToDo Store to responseObject. Check that the mods apply.
```
alena1108: is the ToDo supposed to be done in this PR?

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
hibooboo2: Would renaming the function to getOrCreateKeyFromFile and making the function generate and save a key if it can't get one be a good option as well? basically taking the logic here and moving it into encryption utils.

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
```diff
@@ -0,0 +1,70 @@
...
+            if (!fields.containsKey(fieldName) || object.get(fieldName) == null){
+                continue;
+            }
+            Field field = fields.get(fieldName);
+            if (StringUtils.isNotBlank(field.getTransform())){
+                String decrypted;
+                try {
+                    decrypted = encryptionHandler.untransform((String) object.get(fieldName));
+                } catch (UnsupportedOperationException | IllegalArgumentException e) {
+                        decrypted = "";
```
alena1108: Any reason why we've chosen to return empty string when decryption failure happens?

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
hibooboo2: great Catch i need to do this before we merge this. Glad i put it there. Thanks

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
alena1108: Even if you move it there, I don't think catching NPE would be a good idea. It would be cleaner if some exception indicating that the key doesn't exist, is thrown by getKeyFromFile method. Or simply make it return null when the key can't be found. And just validate if encryptionKey is null. If its null, create a new one.

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/EncryptionUtils.java
alena1108: On a second thought, if the filed-to-encrypt/decrypt is null, we shouldn't even try to call encryption methods on it. Basically upon the very first retrieval of the field, check if its not null, then if it is, return. If not null, check if encryption is enabled for it, and proceed with the logic.

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/handler/impl/TransformerHandlerImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+        if (!getTransformers().containsKey(name)){
+            throw new IllegalArgumentException(name);
+        }
+        return name + TransformerHandler.ENCRYPTER_NAME_DELM + getTransformers().get(name).transform(value);
+    }
+
+    @Override
+    protected String untransformInternal(String transformerName, String value) {
+        if (!getTransformers().containsKey(transformerName)){
```
alena1108: Nit: we can extract lines 31:33 to its own method. As this logic is being re-used on lines 39:41

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    public String decrypt(String value) {
+        try {
+            String[] split = value.split(":", 2);
+            decrypter.init(Cipher.DECRYPT_MODE, key, new IvParameterSpec(split[0].getBytes("UTF-8")));
+            return new String(decrypter.doFinal(Hex.decodeHex(split[1].toCharArray())), "UTF-8");
+        } catch (Exception e){
+            throw new IllegalArgumentException();
```
alena1108: Can we provide more context in this exception? Some message indicating on what exactly failed

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
hibooboo2: Ok I can do that.

alena1108:
We should add a couple more tests: 

1) test filed with NULL value set with "transform" attribute
2) test filed with empty value set with "transform" attribute

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
hibooboo2: Firstly should always be able to decrypt. 
Secondly if it is an unsupportedoperationException it is a hashed value so it should be blank returned. 
Thirdly if it is an IllegalArgumentException then we don't know how to decrypt it but we should not be exposing the info in the db, however we should also not be erroring from trying to decrypt something.

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/handler/impl/TransformerHandlerImpl.java
hibooboo2: Sure

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
hibooboo2: Actually I am logging the stacktrace for illegals so I believe this is the correct behavior.

hibooboo2:
I am assuming encrypt decrypt of blank values should return blank values correct?'

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
alena1108: >> Firstly should always be able to decrypt. -

Would the above be true for the upgrade scenario when field was stored as unencrypted, and then "transform" attribute would be set for it? Just want to make sure that these 2 exceptions resulting in empty return, won't happen in this situation.

alena1108:
@hibooboo2 
>> I am assuming encrypt decrypt of blank values should return blank values correct?'

yes

code/framework/encryption/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk16</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
+            <version>1.6</version>
```
ibuildthecloud: remove version

code/framework/encryption/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+        </dependency>
+        <dependency>
```
ibuildthecloud: I'd remove this

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/Encrypter.java
```diff
@@ -0,0 +1,31 @@
...
+public abstract class Encrypter implements Transformer {
+
+    @Override
+    public boolean compare(String plainText, String encrypted) {
+        return EncryptionUtils.isEqual(decrypt(encrypted), plainText);
+    }
+
+    @Override
+    public String transform(String value){
+        if (StringUtils.isBlank(value)){
```
ibuildthecloud: hmmm

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/EncryptionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+            return null;
+        }
+        JSONObject keyStore = getKeyStore();
+        if (keyStore == null) {
+            return null;
+        }
+        return keyStore.get(propertyName) == null ? null : keyStore.get(propertyName).toString();
+    }
+
+    private static JSONObject getKeyStore() {
```
ibuildthecloud: Use jackson unmarshall to a Map<String,String>

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/EncryptionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+        try (FileReader reader = new FileReader(fileName)) {
+            JSONParser parser = new JSONParser();
+            return (JSONObject) parser.parse(reader);
+        } catch (ParseException | IOException e) {
+            return null; //Should we actually error her for parseexception and not io?
+        }
+    }
+
+    public static void saveKeyToFile(String key, String value) throws IOException {
```
ibuildthecloud: make this method synchronized

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/EncryptionUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
+import org.apache.commons.lang3.StringUtils;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
+
+public class EncryptionUtils {
+    private static String fileName = "keys.json";
```
ibuildthecloud: Properties file would be better

ibuildthecloud: Make this configurable from `Archaius`

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
```diff
@@ -0,0 +1,67 @@
...
+    private String encryptionKey;
+
+    private Cipher encrypter;
+    private Cipher decrypter;
+    private static final String NAME = "AES128";
+    private Key key;
+    private SecureRandom rn;
+
+    public Aes128Encrypter() throws Exception {
+        rn = new SecureRandom();
```
ibuildthecloud: Do initializtion in `@PostConstruct init()`

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
```diff
@@ -0,0 +1,67 @@
...
+
+        encrypter = Cipher.getInstance("AES/CBC/PKCS5PADDING", "SunJCE");
```
ibuildthecloud: encrypter/decrypter not thread safe

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Sha256Hasher.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.UnsupportedEncodingException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+
+public class Sha256Hasher extends Hasher {
+
+    private static final String NAME = "SHA256";
+
+    private MessageDigest md;
```
ibuildthecloud: not thread safe

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Sha256Hasher.java
```diff
@@ -0,0 +1,59 @@
...
+        try {
+            md.update(salt.getBytes("UTF-8"));
+            String hashed = bytesToHex(md.digest(data.getBytes("UTF-8")));
+            return salt + ":" + hashed;
+        } catch (UnsupportedEncodingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+    private String bytesToHex(byte[] bytes) {
+        StringBuffer result = new StringBuffer();
```
ibuildthecloud: Use stringbuilder

code/framework/encryption/src/main/resources/META-INF/cattle/defaults.properties
```diff
@@ -0,0 +1,3 @@
+#Why isn't this being read?
```
ibuildthecloud: Because

ibuildthecloud: Needs to be put in a module directory like `META-INF/cattle/encryption/`

code/framework/encryption/src/main/resources/META-INF/cattle/system-services/spring-cattle-framework-encryption-context.xml
```diff
@@ -0,0 +1,26 @@
...
+       xmlns:tx="http://www.springframework.org/schema/tx"
+       xmlns:extension="http://cattle.io/schemas/spring/extension"
+       xsi:schemaLocation="http://www.springframework.org/schema/aop http://www.springframework.org/schema/aop/spring-aop-3.0.xsd
+		http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
+		http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util-3.2.xsd
+		http://www.springframework.org/schema/tx http://www.springframework.org/schema/tx/spring-tx.xsd
+        http://cattle.io/schemas/spring/extension http://cattle.io/schemas/spring/extension-1.0.xsd
+		http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-3.0.xsd">
+
+       <bean class="io.cattle.platform.framework.encryption.impl.NoOpTransformer" />
```
ibuildthecloud: Lets just not register this bean

ibuildthecloud:
In general avoid str.split() in this code just because it's performance sensitive.  Use StringUtils.split(a,b,2)

code/meta-parent/pom.xml
```diff
@@ -377,6 +377,11 @@
                 <version>3.1</version>
             </dependency>
             <dependency>
+                <groupId>io.github.ibuildthecloud</groupId>
+                <artifactId>gdapi-java-server</artifactId>
+                <version>0.11.4-SNAPSHOT</version>
```
ibuildthecloud: remove

resources/content/cattle-global.properties
```diff
@@ -70,3 +70,6 @@ db.maxtotal=1000
 db.cattle.mysql.validationquery=SELECT 1
 db.cattle.hsqldb.validationquery=SELECT 1 FROM INFORMATION_SCHEMA.SYSTEM_USERS
 
+# Settings for encryption implementation to use. Currently these two are the only implementations.
+api.encryption.encrypter=AES128
```
ibuildthecloud: remove

ibuildthecloud: once you get the other file working

resources/content/schema/base/apiKey.json
```diff
@@ -0,0 +1,10 @@
+{
+  "resourceFields": {
+    "secretValue": {
+      "transform": "HASH"
+    },
+    "name": {
```
ibuildthecloud: remove

ibuildthecloud:
Move Transform abstraction to gdapi.  Encryption stays in encryption package/mvn module.

Add TransformHandler to ApiContext as property.  Create api handler that runs in the beginning that sets the TransformHandler on ApiContext (transformation api handler would be a good place).

TransformerHandler => TransformationService

Get rid of DetransformationHandler and put logic in WrappedResource.init().

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/impl/Aes128Encrypter.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final String NAME = "AES128";
+    private Key key;
+    private SecureRandom rn;
+
+    @PostConstruct
+    public void init() {
+        try {
+            rn = new SecureRandom();
+
+            encryptionKey = encryptionUtils.getKeyFromFile(NAME + "Key");
```
ibuildthecloud: i don't see that this is hex decoded.  The encryptionUtils should probably get and take a `byte[]` and not have the caller do that.

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/request/handler/DeTransformationHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class DeTransformationHandler extends AbstractApiRequestHandler {
+
+    @Override
+    public void handle(ApiRequest request) throws IOException {
+        //TODO: remove this whole class after verifying that it works from wrapped resource.
```
ibuildthecloud: Is it time to do this?

hibooboo2: So i did but i guess i put it on the wrong pr.
It is on top of local auth pr. Ill fix that.

PR #727, opened 2015-07-30
Dns support for containers running on host networking 

alena1108:
https://github.com/rancher/rancher/issues/1669


alena1108:
Works in conjunction with https://github.com/rancher/host-api/pull/42

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceCreate.java
```diff
@@ -153,13 +163,37 @@ protected Volume getRoot(Instance instance, List<Volume> volumes) {
...
         List<Long> subnetIds = DataUtils.getFieldList(instance.getData(), InstanceConstants.FIELD_SUBNET_IDS, Long.class);
         List<Long> vnetIds = DataUtils.getFieldList(instance.getData(), InstanceConstants.FIELD_VNET_IDS, Long.class);
-
         return createNicsFromIds(instance, nics, data, networkIds, subnetIds, vnetIds);
     }
 
+    protected List<Long> populateNetworks(Instance instance) {
+        List<Long> networkIds = DataUtils.getFieldList(instance.getData(), InstanceConstants.FIELD_NETWORK_IDS,
+                Long.class);
+        Label dnsLabel = labelsDao.getLabelForInstance(instance.getId(), SystemLabels.LABEL_USE_RANCHER_DNS);
```
ibuildthecloud: Why do you refer to the database and not the `Labels` field on the container?  During create I would almost assume the labels table wouldn't be populated.  Regardless, the Labels field on the container should be the source of truth and the other db tables are just a normalized form to make some queries better.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceCreate.java
```diff
@@ -153,13 +163,37 @@ protected Volume getRoot(Instance instance, List<Volume> volumes) {
...
+        if (networkIds == null || dnsLabel == null || !dnsLabel.getValue().equals("true")) {
+            return networkIds;
+        }
+        Network hostNetwork = null;
+        Network managedNetwork = null;
+        for (Long networkId : networkIds) {
+            Network network = objectManager.loadResource(Network.class, networkId);
+            if (network.getKind().equalsIgnoreCase(DockerNetworkConstants.KIND_DOCKER_HOST)) {
```
ibuildthecloud: When is this network created?

ibuildthecloud: oops, nvm.  I remember

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceCreate.java
alena1108: Labels do get populated during the create, in the same InstanceCreate handler. But sure, I can read it from the field instead.

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/NetworkDaoImpl.java
```diff
@@ -1,30 +1,38 @@
 package io.cattle.platform.core.dao.impl;
 
 
-import static io.cattle.platform.core.model.tables.NetworkServiceProviderInstanceMapTable.*;
```
ibuildthecloud: boo... set your Eclipse settings to only require 1 static import to allow *.  (You don't need to change this now)

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/NetworkDaoImpl.java
```diff
@@ -123,4 +131,30 @@ public Network getNetworkForObject(Object object, String networkKind) {
         }
         return systemNetworks.get(0);
     }
+
+    @Override
+    public IpAddress getHostIPForInstanceWHostNetwork(Instance instance) {
```
ibuildthecloud: nit: W... I'll never know what W stands for.  Can you just put `With` (I assume you meant that)

resources/content/config-content/host-iptables/apply.sh
```diff
@@ -22,4 +22,9 @@ done
 
 add_route_table 300
 
+
+if ! ip route show | grep 169.254.169.250; then
```
ibuildthecloud: This won't fly because we are hard coding docker0.  Let me look, I think there is a better place for this.

resources/content/config-content/hosts/content-home/etc/cattle/dns/answers.json.ftl
```diff
@@ -18,6 +18,10 @@
                     </#if>
                 </#list>
         ],
+        	<#else>
+        "recurse": [
+            PARENT_DNS
```
ibuildthecloud: What's this coolness for?

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/NetworkDaoImpl.java
alena1108: will fix

resources/content/config-content/hosts/content-home/etc/cattle/dns/answers.json.ftl
alena1108: Remember we spoke how host container itself will get represented as "default" source in answers.json file? So for it in order to resolve external ips there should be an external DNS ref, therefore this line

PR #728, opened 2015-08-01
Upgrade to jetty 9

cjellick:
Upgrade to jetty 9 for improved websocket stability.

Dont merge yet. I want to run some tests in GCE first.

### DIFF
--- a/code/framework/api-pub-sub-jetty/src/main/java/io/cattle/platform/api/pubsub/subscribe/jetty/JettyWebSocketSubcriptionHandler.java
+++ b/code/framework/api-pub-sub-jetty/src/main/java/io/cattle/platform/api/pubsub/subscribe/jetty/JettyWebSocketSubcriptionHandler.java
@@ -15,9 +15,10 @@ import java.util.concurrent.ExecutorService;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-import org.eclipse.jetty.websocket.WebSocket;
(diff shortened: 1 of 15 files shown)


code/framework/api-pub-sub-jetty/src/main/java/io/cattle/platform/api/pubsub/subscribe/jetty/JettyWebSocketSubcriptionHandler.java
```diff
@@ -37,20 +38,17 @@ protected MessageWriter getMessageWriter(ApiRequest apiRequest) throws IOExcepti
...
             }
-
-            @Override
-            public boolean checkOrigin(HttpServletRequest request, String origin) {
-                return true;
-            }
         });
 
-        if (factory.acceptWebSocket(req, resp)) {
+        if ("websocket".equalsIgnoreCase(req.getHeader("Upgrade")) && factory.acceptWebSocket(req, resp)) {
```
alena1108: Is there an easy way to add a test for this use case? for "upgrade" request header

cjellick: Effectively, it is already tested. The websocket specification requires that that header be present when a client requests a websocket connection. We already have tests that establish websocket connections.

alena1108:
LGTM

PR #730, opened 2015-08-03
Sidekick fixes

alena1108:
1) Restart service's stopped containers only when RestartPolicy is explicitly set to "No" on the service.

https://github.com/rancher/rancher/issues/1596

@vincent99 the UI by default passes restartPolicy=No. If its not possible not to send restartPolicy unless user choses to use it, we might change the default to be Always, so the default behavior is "always restart"

2) Don't destroy service's container, unless its:

* unhealthy 
* in error state
* references sidekick container that is being dead

Before the fix, we've re-created all deployment unit's containers if at least one of them was dead. After the fix, we are more deliberate about what containers have to be recreated based on the dependencies (networkFrom/dataVolumesFrom within the deployment unit)

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/DeploymentUnitInstance.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/DeploymentUnitInstance.java
@@ -1,10 +1,13 @@
 package io.cattle.platform.servicediscovery.deployment;
 
+import io.cattle.platform.core.addon.RestartPolicy;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.core.model.ServiceExposeMap;
+import io.cattle.platform.docker.constants.DockerInstanceConstants;
 import io.cattle.platform.object.process.StandardProcess;
 import io.cattle.platform.object.resource.ResourcePredicate;
+import io.cattle.platform.servicediscovery.api.util.ServiceDiscoveryUtil;
 import io.cattle.platform.servicediscovery.deployment.impl.DeploymentManagerImpl.DeploymentServiceContext;
 
 import java.util.Map;
@@ -21,6 +24,8 @@ public abstract class DeploymentUnitInstance {
     protected ServiceExposeMap exposeMap;
     protected String launchConfigName;
     protected Service service;
+    protected boolean autoRestart = true;
+    protected boolean firstStart = true;
 
     public abstract boolean isError();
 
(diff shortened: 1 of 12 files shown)


alena1108:
Don't merge till we confirm with @vincent99 the default restart policy - or lack thereof - set in the UI for the service.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/DeploymentUnitInstance.java
```diff
@@ -42,21 +47,60 @@ public boolean evaluate(ServiceExposeMap obj) {
...
     protected DeploymentUnitInstance(DeploymentServiceContext context, String uuid, Service service,
             String launchConfigName) {
         this.context = context;
         this.uuid = uuid;
         this.launchConfigName = launchConfigName;
         this.service = service;
+        Object policy = ServiceDiscoveryUtil.getLaunchConfigObject(service, launchConfigName,
+                DockerInstanceConstants.FIELD_RESTART_POLICY);
+
+        if (policy != null && ((Map<String, String>) policy).get("name").equalsIgnoreCase(RestartPolicy.RESTART_NEVER)) {
```
sonchang: Might want to switch it around to:
RestartPolicy.RESTART_NEVER.equalsIgnoreCase(((Map<String, String>) policy).get("name")) just in case ((Map<String, String>) policy).get("name") is null

alena1108: Sure, thanks

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentManagerImpl.java
```diff
@@ -175,6 +179,11 @@ protected void activateDeploymentUnits(ServiceDeploymentPlanner planner) {
         deleteBadUnits(planner);
 
         /*
+         * Clenaup incomplete units
```
sonchang: super tiny typos

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -195,75 +221,51 @@ protected DeploymentUnitInstance createInstance(String launchConfigName, Service
...
+            }
         }
-        
-        if (volumesFromLaunchConfigs != null) {
-            for (String volumesFromLaunchConfig : (List<String>) volumesFromLaunchConfigs) {
+
+        Object sidekicksLaunchConfigObj = ServiceDiscoveryUtil.getLaunchConfigObject(service, launchConfigName,
+                sidekickType.launchConfigType);
+        if (sidekicksLaunchConfigObj != null) {
+            List<String> sidekicksLaunchConfigIds = new ArrayList<>();
```
sonchang: Minor: The name 'sidekicksLaunchConfigIds' was a little confusing since the elements were compared against service.getName().  And I usually don't think of service name as an ID.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DeploymentUnit.java
```diff
@@ -24,6 +24,24 @@
 import java.util.UUID;
 
 public class DeploymentUnit {
+    
+    public static class SidekickType {
```
sonchang: Minor: Usually I go with an enum but this works too

sonchang:
LGTM

vincent99:
The default is "no".. I can make it "always" (or anything else), but I don't think a "null" option is good idea, what would it mean to the user?  It would have to be a 5th option that I show and explain to the user (or once you've clicked of the existing options there's no way to go back to null).

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -2229,6 +2235,46 @@ def test_validate_image(client, context):
     assert e.value.error.fieldName == 'imageUuid'
 
 
+def test_validate_restart_policy(client, context):
+    env = client.create_environment(name=random_str())
+    env = client.wait_success(env)
+    assert env.state == "active"
+
+    restart_policy = {"maximumRetryCount": 0, "name": "no"}
```
sonchang: name=no seems a little weird

alena1108: @sonchang  thats docker thing. Restart policy name can be one of the values below:

--restart="no"             Restart policy (no, on-failure[:max-retry], always)

and for on-failure you can also specify max-retry count param

sonchang: What about the key being "name".  Feels like it should've been "value" maybe

alena1108:
@vincent99 I didn't mean we should present "none" it as an option, I rather thought that as with our other params, we just don't send the param unless user specifically sets it.

With this fix, we just want to utilize restartPolicy parameter to signal Service manager when the user doesn't care if the service's instances are always in Running state after their initial creation; but only when user sets this param explicitly... 

So not sure what should be the default in the UI, perhaps "always" to be aligned with the default service reconcile behavior - always bring up services instances that were stopped.

@ibuildthecloud need your opinion on that

tests/integration/cattletest/core/test_svc_discovery.py
alena1108: thats again, docker thing. Here is their format:

"RestartPolicy": {
            "Name": "no",
            "MaximumRetryCount": 0
        },

sonchang: Thx!

vincent99:
But "user sets explicitly" doesn't work for something that's an enumeration/radio buttons.. there's 4 radio buttons and one of them has to be checked.  (Technically I can have none of them checked and that can mean null, but once you've click one there's no way to go back to none checked).

The things that aren't sent unless the user sets are nullable, arbitrary string fields (and it's actually a side-effect, special-cased because there are some fields in the API where null is valid but an empty string isn't.  HTML inputs don't have a null value)

alena1108:
@vincent99 all understood :) just thinking what would be the less confusing approach here as we switching behavior from auto restarting services instances on all possible restart policy conditions + lack thereof, to the current behavior when we don't restart only when restartPolicy=no is set (current UI default)

PR #731, opened 2015-08-04
Set envVar instance.name for service's instance

alena1108:
https://github.com/rancher/rancher/issues/1724


ibuildthecloud:
Why are we touching environment variables?

alena1108:
@ibuildthecloud there is a use case (@will-chan has more details on that) when service container needs to know its own name; and it should be able to figure it out from within the container (so just setting up the label won't be a solution). Metadata service would be able to solve it properly, so I believe this is the temporary workaround.

PR #732, opened 2015-08-05
Service Discovery bug fixes

alena1108:
1) Host name override support 

https://github.com/rancher/rancher/issues/1724

2) Small fix for hostname routing - only set backend as default when it has no host/domain set

https://github.com/rancher/rancher/issues/1722


alena1108:
@ibuildthecloud

PR #734, opened 2015-08-06
Include port number in haproxy config for hdr(host) acl.

kaos:
In order for HA proxy to route traffic based on the Host header, when the client includes the port number in the Host header, the acl must also include the port number in order for it to match.

See http://serverfault.com/questions/502443/ignore-port-numbers-in-haproxy-host-header-matches for a related issue that describes the use case this solves.

### DIFF
--- a/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
+++ b/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
@@ -39,6 +39,7 @@ frontend ${listener.uuid}_frontend
         <#if (listener.sourceProtocol == "http" || listener.sourceProtocol == "https") && (backend.portSpec.domain != "default" || backend.portSpec.path != "default")>
         <#if backend.portSpec.domain != "default">
         acl ${backend.uuid}_host hdr(host) -i ${backend.portSpec.domain}
+        acl ${backend.uuid}_host hdr(host) -i ${backend.portSpec.domain}:${sourcePort}
     	</#if>
     	<#if backend.portSpec.path != "default">
         acl ${backend.uuid}_path path_beg -i ${backend.portSpec.path}


resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
```diff
@@ -39,6 +39,7 @@ frontend ${listener.uuid}_frontend
         <#if (listener.sourceProtocol == "http" || listener.sourceProtocol == "https") && (backend.portSpec.domain != "default" || backend.portSpec.path != "default")>
         <#if backend.portSpec.domain != "default">
         acl ${backend.uuid}_host hdr(host) -i ${backend.portSpec.domain}
+        acl ${backend.uuid}_host hdr(host) -i ${backend.portSpec.domain}:${sourcePort}
```
alena1108: @kaos thanks for pointing out the use case. We need to name this ACL rule differently though because we already have the ACL rule above using the same name (${backend.uuid}_host hdr(host)) + include it to the line evaluating it:

use_backend ${listener.uuid}_${backend.uuid}_backend if <#if backend.portSpec.domain != "default">${backend.uuid}_host</#if> <#if backend.portSpec.path != "default">${backend.uuid}_path</#if>

alena1108:
@kaos nvm my last comment, there seems to be 3 valid ways of creating OR condition in haproxy ACLs:

1) yours
2)  Putting 2 conditions into the same ACL separated by space: acl ${backend.uuid}_host hdr(host) -i ${backend.portSpec.domain} ${backend.portSpec.domain}:${sourcePort}
3) creating 2 ACLs with diff name, and mention them with OR in "use_backend" condition.

Tested all, works fine.

LGTM

cjellick:
LGTM

PR #736, opened 2015-08-07
Validate instance state after ensuring it exists

alena1108:
Fixing NPE Sangeetha sees in the log. Attempt to restart instance only when it exists

https://github.com/rancher/rancher/issues/1755


cjellick:
LGTM

PR #739, opened 2015-08-10
Metadata support

alena1108:
This PR also requires metadata service packaging

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/MetaDataInfoDao.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/MetaDataInfoDao.java
@@ -1,0 +1,13 @@
+package io.cattle.platform.configitem.context.dao;
+
+import io.cattle.platform.configitem.context.data.ContainerMetaData;
+
+import java.util.List;
+
+public interface MetaDataInfoDao {
+
+    List<ContainerMetaData> getServicesContainersData(long accountId);
+
+    List<ContainerMetaData> getStandaloneContainersData(long accountId);
+
+}
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
@@ -1,0 +1,143 @@
+package io.cattle.platform.configitem.context.dao.impl;
+
+import static io.cattle.platform.core.model.tables.HostIpAddressMapTable.HOST_IP_ADDRESS_MAP;
+import static io.cattle.platform.core.model.tables.HostTable.HOST;
+import static io.cattle.platform.core.model.tables.InstanceHostMapTable.INSTANCE_HOST_MAP;
+import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
+import static io.cattle.platform.core.model.tables.IpAddressNicMapTable.IP_ADDRESS_NIC_MAP;
+import static io.cattle.platform.core.model.tables.IpAddressTable.IP_ADDRESS;
+import static io.cattle.platform.core.model.tables.NicTable.NIC;
+import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.SERVICE_EXPOSE_MAP;
+import io.cattle.platform.configitem.context.dao.MetaDataInfoDao;
+import io.cattle.platform.configitem.context.data.ContainerMetaData;
+import io.cattle.platform.configitem.context.data.HostMetaData;
+import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.constants.IpAddressConstants;
+import io.cattle.platform.core.model.Host;
+import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.model.IpAddress;
+import io.cattle.platform.core.model.ServiceExposeMap;
+import io.cattle.platform.core.model.tables.HostTable;
+import io.cattle.platform.core.model.tables.InstanceTable;
+import io.cattle.platform.core.model.tables.IpAddressTable;
+import io.cattle.platform.core.model.tables.ServiceExposeMapTable;
+import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
+import io.cattle.platform.db.jooq.mapper.MultiRecordMapper;
+import io.cattle.platform.object.util.DataAccessor;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings("unchecked")
+public class MetaDataInfoDaoImpl extends AbstractJooqDao implements MetaDataInfoDao {
+    @Override
+    public List<ContainerMetaData> getServicesContainersData(long accountId) {
+        MultiRecordMapper<ContainerMetaData> mapper = new MultiRecordMapper<ContainerMetaData>() {
+            @Override
+            protected ContainerMetaData map(List<Object> input) {
+                ContainerMetaData data = new ContainerMetaData();
+                data.setInstance((Instance) input.get(1));
+                data.setIp((IpAddress) input.get(3));
+                if (input.get(2) != null) {
+                    Host host = (Host) input.get(2);
+                    IpAddress hostIpAddress = (IpAddress) input.get(0);
+                    if (host != null) {
+                        Map<String, String> hostLabels = DataAccessor.fields(host)
+                                .withKey(InstanceConstants.FIELD_LABELS)
+                                .withDefault(Collections.EMPTY_MAP).as(Map.class);
+                        HostMetaData hostMetaData = new HostMetaData(hostIpAddress.getAddress(), host.getName(),
+                                hostLabels, host.getId());
+                        data.setHostMetaData(hostMetaData);
+                    }
+                }
+                if (input.get(4) != null) {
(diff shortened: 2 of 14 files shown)


alena1108:
@ibuildthecloud @vincent99 could you please validate answers.json content. I've implemented it based on the screenshot shared by Darren from the meeting two of you had. Could of extra fields that I've added:

* "state" for the container. As we are going to push all the containers data (except for removed ones). it might be important to display the state 
* "vip" for the service
* "kind" for the service
* "ports" field to the services (publicPort:privatePort format). We also have it per container (host:publicPort:privatePort format)
* use map instead of array for storing the "links". As link can have an alias

```
{
    "default": {
        "stacks": [
            {
                "environmentName": "Default",
                "name": "Default",
                "services": [
                    "regular",
                    "lb",
                    "external",
                    "externalwhost",
                    "alias"
                ]
            }
        ],
        "services": [
            {
                "name": "regular",
                "stackName": "Default",
                "kind": "service",
                "hostName": null,
                "vip": null,
                "ipAddresses": [],
                "sidekicks": [],
                "containers": [
                    "Default_regular_1"
                ],
                "links": {
                    "alias": "alias"
                },
                "ports": []
            },
            {
                "name": "lb",
                "stackName": "Default",
                "kind": "loadBalancerService",
                "hostName": null,
                "vip": null,
                "ipAddresses": [],
                "sidekicks": [],
                "containers": [
                    "Lb Agent"
                ],
                "links": {
                    "lb": "regular"
                },
                "ports": [
                    "100"
                ]
            },
            {
                "name": "external",
                "stackName": "Default",
                "kind": "externalService",
                "hostName": null,
                "vip": null,
                "ipAddresses": [
                    "172.18.0.1"
                ],
                "sidekicks": [],
                "containers": [],
                "links": {},
                "ports": []
            },
            {
                "name": "externalwhost",
                "stackName": "Default",
                "kind": "externalService",
                "hostName": "google.com",
                "vip": null,
                "ipAddresses": [],
                "sidekicks": [],
                "containers": [],
                "links": {},
                "ports": []
            },
            {
                "name": "alias",
                "stackName": "Default",
                "kind": "dnsService",
                "hostName": null,
                "vip": null,
                "ipAddresses": [],
                "sidekicks": [],
                "containers": [],
                "links": {
                    "alias": "lb"
                },
                "ports": []
            }
        ],
        "containers": [
            {
                "name": "Default_regular_1",
                "primaryIp": "10.42.65.14",
                "ips": [
                    "10.42.65.14"
                ],
                "ports": [],
                "serviceName": "regular",
                "state": "running"
            },
            {
                "name": "Lb Agent",
                "primaryIp": "10.42.217.3",
                "ips": [
                    "10.42.217.3"
                ],
                "ports": [
                    "127.0.0.1:100:100"
                ],
                "serviceName": "lb",
                "state": "running"
            }
        ]
    },
    "10.42.217.3": {
        "container": {
            "name": "Lb Agent",
            "primaryIp": "10.42.217.3",
            "ips": [
                "10.42.217.3"
            ],
            "ports": [
                "127.0.0.1:100:100"
            ],
            "serviceName": "lb",
            "state": "running"
        },
        "service": {
            "name": "lb",
            "stackName": "Default",
            "kind": "loadBalancerService",
            "hostName": null,
            "vip": null,
            "ipAddresses": [],
            "sidekicks": [],
            "containers": [
                "Lb Agent"
            ],
            "links": {
                "lb": "regular"
            },
            "ports": [
                "100"
            ]
        },
        "stack": {
            "environmentName": "Default",
            "name": "Default",
            "services": [
                "regular",
                "lb",
                "external",
                "externalwhost",
                "alias"
            ]
        }
    },
    "10.42.65.14": {
        "container": {
            "name": "Default_regular_1",
            "primaryIp": "10.42.65.14",
            "ips": [
                "10.42.65.14"
            ],
            "ports": [],
            "serviceName": "regular",
            "state": "running"
        },
        "service": {
            "name": "regular",
            "stackName": "Default",
            "kind": "service",
            "hostName": null,
            "vip": null,
            "ipAddresses": [],
            "sidekicks": [],
            "containers": [
                "Default_regular_1"
            ],
            "links": {
                "alias": "alias"
            },
            "ports": []
        },
        "stack": {
            "environmentName": "Default",
            "name": "Default",
            "services": [
                "regular",
                "lb",
                "external",
                "externalwhost",
                "alias"
            ]
        }
    }
}
```

vincent99:
- `state` means you're going to have to push down a new answers.json every time a container's state changes... Seems like potentially not a good idea for performance.
- We usually do `hostname` as one word, not inter-caps two.  Even though it's next to `ipAddresses` and `stackName` :smile:
- Should all the keys be snake_case?  They all turn into URLs (e.g. `/services/0/stack_name/`) so case-sensitivity is not very desirable.

sonchang:
If the containers have the stackName in addition to the serviceName, then I should be able to still get all the info I need by just querying all the containers.  Rather than query both "services" and "containers" and combine the resulting data.

alena1108:
@vincent99 

>> state means you're going to have to push down a new answers.json every time a container's state changes... Seems like potentially not a good idea for performance.

Not exactly. Config update is triggered by the list of predefined processes, not on every state change. I still think displaying the state might be a good idea as there might be cases when we want to return Stopped instances in metadata response, but we want to signal the client that the instance is actually stopped

>>We usually do hostname as one word, not inter-caps two. Even though it's next to ipAddresses and stackName :smile:

One liner change, np :) 

>> Should all the keys be snake_case? They all turn into URLs (e.g. /services/0/stack_name/) so case-sensitivity is not very desirable.

will fix

alena1108:
@sonchang 

will fix

sonchang:
quick question.  If I query for 'container', will this just bring up the containers running on this host?  Since I'll need container information for all containers within a service.  @vincent99 @ibuildthecloud

alena1108:
@sonchang it will bring up the containers for the service. But I also see that having host_id field on container would be a good idea.

vincent99:
> Not exactly. Config update is triggered by the list of predefined processes, not on every state change.

Ok, but if you provide it users will expect it to be accurate, and updated in a timely manner.  You could imagine someone using that to do autoscaling or monitoring. 

> it will bring up the containers for the service.

Am I missing something?  `/{version}/container` is the information about the (singular) container that is making the request.  It doesn't tell you about the other containers that make up the service.

There doesn't seem to be any simple way to get "all the containers for the current service", you'd want to use JSON or it will be many requests.. `/{version}/containers/`.filter((container) => { return container.serviceName == `/{version}/service/name`; })

Seems like there an issue here with service names not being unique across stacks...  We used `stackName/serviceName` as the primary identifier for labels because of that.

ibuildthecloud:
@alena1108 Let's drop state.  One thing that has come up in discussion w/ users is being about to get the host labels is useful.  I don't know where we should put that.  @vincent99 ideas?

vincent99:
`/{version}/host` with info about the current host?  Any reason why not show labels for containers & services too?

alena1108:
@vincent99 no reason, just missed it - labels. I'll add the host info as well.

@ibuildthecloud I'll drop the state

ibuildthecloud:
Labels for services and containers would be good.  Why do we have network/volumes from and sidekick.  I'd like to start with a minimum amount and then add things.  I can't really think why is need this metadata

alena1108:
@vincent99 @sonchang sorry, overlooked "container" as "service".  So under container you will see the information about only this container + its service.

Its when you query service, you get the info about all the containers of this service, not only containers running on this host.

alena1108:
@ibuildthecloud I can remove sidekicks for now, and leave volumes_from, network_from

ibuildthecloud:
Id like to drop networks/container from.  We can always easily add more fields.  But removing is hard

alena1108:
@ibuildthecloud ok, will leave sidekicks and remove the other two

ibuildthecloud:
Wait @alena1108 do we need sidekick?  I'd still want to remove that too

alena1108:
@ibuildthecloud it was on the whiteboard drawing from your meeting with @vincent99 ...but sure, can remove that one

ibuildthecloud:
@alena1108 Oh, maybe there was a reason.  Let's keep it

alena1108:
@ibuildthecloud if there is a use case requiring displaying sidekicks, do you think volumesFrom/networksFrom can be relevant too? As people often use sidekicks in volumesFrom/networkFrom context

ibuildthecloud:
I think the reason we put sidekick is because a sidekick service is not a real service, so we wanted to know what the real service was.  I'd still avoid network/volumes.

alena1108:
@ibuildthecloud ok

ibuildthecloud:
@vincent99 @alena1108 seems like we should add a DNS name for 169.254.169.250.  Ideas?? We could just use `rancher-metadata`.  I thought maybe `metadata` but that might be too generic.

alena1108:
@ibuildthecloud @vincent99 I like rancher-metadata better as a) it implies its our internal server b) it would never conflict with anything else

vincent99:
`rancher-metadata` sounds good, if you're willing to call it metadata and not carefully-worded-introspection-API.

ibuildthecloud:
Metadata is fine.  Introspection api backfired.  People thought it sounded too complicated.

alena1108:
Added "version" field to "default" section:

```
{"default": {
 	"version": "2-aae727c7fbc55f447f3e8c49cc4e50dd806913bc2cbd72e34e64649f294ca977",
...........
```

@sonchang

alena1108:
Dont merge yet, going to address some performance issues during the weekend

alena1108:
Performance issue is addressed, ready to be merged.

PR #747, opened 2015-08-13
Account for networkMode = "default" in docker 1.8

cjellick:
addresses https://github.com/rancher/rancher/issues/1791

### DIFF
--- a/code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
+++ b/code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
@@ -172,7 +172,7 @@ public class DockerTransformerImpl implements DockerTransformer {
                     NETWORK_MODE_HOST.equals(inspectNetMode) ||
                     NETWORK_MODE_NONE.equals(inspectNetMode)) {
                 netMode = inspectNetMode;
-            } else if (StringUtils.isBlank(inspectNetMode)) {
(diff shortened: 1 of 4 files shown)


code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
```diff
@@ -172,7 +172,7 @@ void setNetworkMode(Instance instance, ContainerConfig containerConfig, HostConf
                     NETWORK_MODE_HOST.equals(inspectNetMode) ||
                     NETWORK_MODE_NONE.equals(inspectNetMode)) {
                 netMode = inspectNetMode;
-            } else if (StringUtils.isBlank(inspectNetMode)) {
+            } else if (NETWORK_MODE_DEFAULT.equals(inspectNetMode) || StringUtils.isBlank(inspectNetMode)) {
```
alena1108: Would it be backwards compatible with docker 1.7 if some users still use it? or do we force the upgrade?

ibuildthecloud:
LGTM

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
cjellick: It id backwards compatible. We still handle the 1.7 way.

PR #748, opened 2015-08-14
Bug fixes

alena1108:
1) Host networking + Alias service fixes

https://github.com/rancher/rancher/issues/1792
https://github.com/rancher/rancher/issues/1793

2) Account purge 

Cleanup only non-service instances as service instances get removed during environment.remove process triggered earlier

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/dao/InstanceDao.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/dao/InstanceDao.java
@@ -1,5 +1,6 @@
 package io.cattle.platform.core.dao;
 
+import io.cattle.platform.core.model.Account;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.model.Service;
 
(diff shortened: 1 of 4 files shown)


code/iaas/model/src/main/java/io/cattle/platform/core/dao/InstanceDao.java
```diff
@@ -19,4 +20,6 @@
      */
     List<? extends Service> findServicesFor(Instance instance);
 
+    List<? extends Instance> listNonRemovedInstances(Account account, boolean forService);
```
sonchang: Is there somewhere else that makes this call passing in forService=true? (subsequent PR?)

sonchang:
LGTM

code/iaas/model/src/main/java/io/cattle/platform/core/dao/InstanceDao.java
alena1108: @sonchang No subsequent PR, I've just wanted this method to handle both cases - return non-service and service instances.

cjellick:
LGTM

PR #749, opened 2015-08-14
Allow scaleup/scaledown for service in updating-active state

alena1108:
So now you can change the scale for the service stuck in updating-active state, w/o deactivating this service first. 

Also fixes https://github.com/rancher/rancher/issues/1757


ibuildthecloud:
LG(mostly in a hacky way)TM.  I'm okay with merging this even though the solution isn't ideal.

PR #763, opened 2015-08-18
DB query performance improvement

sonchang:


ibuildthecloud:
Thanks, I'll test this out now

PR #767, opened 2015-08-18
Don't trust the database

hibooboo2:
rancher/rancher#1820
When getting projectMembers from database if they are no longer valid from external system do not blow up.


sonchang:
LGTM.  Once again, there should be another reviewer for this.

PR #755, opened 2015-08-17
Various bug fixes

cjellick:
Addresses:
https://github.com/rancher/rancher/issues/1702
and https://github.com/rancher/rancher/issues/1815 
and partially: https://github.com/rancher/rancher/issues/1741. I say partially because it will address the symptom explained in that issue (containers showing up as stopped vs deleted), but it won't address _why_ containers' delete events are getting missed during validation tests.

PR #768, opened 2015-08-18
Improve missing container delete functionality

cjellick:
Allow stopped containers that were removed from the host to be deleted in rancher if rancher has a record of their external id (docker id).


cjellick:
This is a follow up to https://github.com/rancher/cattle/pull/755
which we discovered did not delete containers that were gone from host and stopped in rancher.

ibuildthecloud:
LGTM

PR #770, opened 2015-08-19
add rancherCompose and dockerCompose fields to environment resource

wlan0:
@cjellick, @ibuildthecloud  - This adds the `import` field to service. This field will hold the bytes of rancher-compose.yml file.

### DIFF
--- a/resources/content/schema/base/environment.json
+++ b/resources/content/schema/base/environment.json
@@ -7,6 +7,12 @@
              "attributes" : {
                 "scheduleUpdate" : true
             }
+        },
+        "dockerCompose":{
+            "type": "string"
+        },
+        "rancherCompose":{
+            "type": "string"
         }
     },
     "resourceActions" : {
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
I don't think this needs to be a db column.  Also don't we need two fields?  One for docker-compose and another for rancher-compose

wlan0:
@ibuildthecloud Are you suggesting that this field be in the `data` column?  

 I originally intended to encode both files into a single archive and base64 encode the bytes and set it to this field. I apologize for not mentioning that accurately in my previous description.

ibuildthecloud:
I don't think I'd want them archived.  It just makes it harder to interact with the API of they need to be archived.

wlan0:
@ibuildthecloud I've changed this commit entirely from its initial form.

There are two new fields - `rancherCompose` and `dockerCompose` in the `env` resource

ibuildthecloud:
These don't need to be DB fields.

wlan0:
test_container_port_start failed. It is not related to this PR.

PR #773, opened 2015-08-19
Setup ssl default params on haproxy only when ssl listeners are present

alena1108:
To fix the upgrade bug

https://github.com/rancher/rancher/issues/1834


sonchang:
Any reason why we don't have sslproto as part of the map that we pass into the template?  I guess I'd prefer less logic if possible in the template.  But as a quick fix, LGTM

alena1108:
@sonchang you are right, java code would be cleaner. Let me revamp the fix

PR #776, opened 2015-08-20
[WIP] Add upgrade logic for private registry agents

cloudnautique:
This takes the approach that:
1.) the script is running in rancher-agent
2.) A user has started running it on a private registry and had to set the variable.
3.) All future upgrades will be from a private registry, though it could be switched, and this variable would still be used.

upgrade Logic for rancher/rancher#1840

### DIFF
--- a/resources/content/config-content/bootstrap/bootstrap.sh
+++ b/resources/content/config-content/bootstrap/bootstrap.sh
@@ -104,6 +104,11 @@ upgrade()
     fi
 }
 
+get_running_image()
+{
+    echo $(docker inspect -f '{{.Config.Image}}' rancher-agent)
+}
+
 cd $(dirname $0)
 
 for conf_file in "${CONF[@]}"; do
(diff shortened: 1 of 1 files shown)


resources/content/config-content/bootstrap/bootstrap.sh
```diff
@@ -104,6 +104,11 @@ upgrade()
     fi
 }
 
+get_running_image()
+{
+ echo $(docker inspect rancher-agent|jq -r .[].Config.Image)
```
ibuildthecloud: Can you switch to `docker inspect -f '{{.Config.Image}}'` and use four spaces and not a tab

resources/content/config-content/bootstrap/bootstrap.sh
```diff
@@ -136,6 +141,7 @@ while [ $# != 0 ]; do
     shift 1
 done
 
+export RANCHER_AGENT_IMAGE="$(get_running_image)"
```
ibuildthecloud: Can you move this after `check_debug` but before `print_config`

PR #777, opened 2015-08-20
Only use github at the correct time.

hibooboo2:


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
```diff
@@ -65,25 +65,16 @@ public Account getAccountFromJWT() {
...
-            if (!StringUtils.equalsIgnoreCase("bearer", StringUtils.trim(tokenArr[0]))) {
-                return null;
-            }
-            toParse = tokenArr[1];
-        } else if (tokenArr.length == 1) {
-            toParse = tokenArr[0];
-        } else {
-            toParse = jwtKey;
-        }
+        toParse = removeBearer(jwtKey);
```
alena1108: Nit: can we check if jwtKey is empty right before attempting this call? As I can see that inside removeBearer you are checking for an empty value, and then you do the same on line 75

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/internal/rancher/RancherIdentitySearchProvider.java
```diff
@@ -94,7 +93,7 @@ public Identity getIdentity(String id, String scope) {
 
     @Override
     public boolean isConfigured() {
-        return !SecurityConstants.SECURITY.get();
+        return true;
```
alena1108: could you explain this change? Why before it wasn't a requirement - for this provider to be enabled by default?

hibooboo2: This change is because validation when auth is on apikeys will not get rancher_ids and as a result not work.

alena1108:
Besides my "Nit" comment, LGTM

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
```diff
@@ -65,25 +65,16 @@ public Account getAccountFromJWT() {
...
         String dbJwt = retrieveJwt(toParse);
         if (dbJwt != null){
             toParse = dbJwt;
         }
+        toParse = removeBearer(toParse);
+        if (StringUtils.isEmpty(jwtKey)) {
```
sonchang: I don't understand why this was moved down here.  jwtKey hasn't changed since the start of the function.  The only difference I can see is an exception can be thrown before we reach down here.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
```diff
@@ -104,6 +95,25 @@ public Account getAccountFromJWT() {
...
 
+    private String removeBearer(String jwtKey) {
+        String toParse;
+        if (StringUtils.isBlank(jwtKey)){
+            return jwtKey;
+        }
+        String[] tokenArr = jwtKey.split("\\s+");
+        if (tokenArr.length == 2) {
+            if (!StringUtils.equalsIgnoreCase("bearer", StringUtils.trim(tokenArr[0]))) {
+                return null;
```
sonchang: Are you sure you want to return null here?  Seems like you'd either want to return the jwtKey or possibly throw an exception?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/internal/rancher/RancherIdentitySearchProvider.java
sonchang: I don't understand your explanation.  I think you're missing a lot of punctuation in your sentence.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/GithubIdentitySearchProvider.java
```diff
@@ -81,13 +82,18 @@
 
     @Override
     public Set<Identity> getIdentities(Account account) {
-        if (!isConfigured() && !GithubConstants.CONFIG.equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get())) {
+        if (!isConfigured() || !GithubConstants.CONFIG.equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get()) ||
+                !GithubConstants.USER_SCOPE.equalsIgnoreCase(account.getExternalIdType())) {
             return new HashSet<>();
         }
-        githubUtils.findAndSetJWT();
```
sonchang: Can you provide some background or explain the changes in this file?  Not questioning your changes.  Just need some background to understand.  Thanks!

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
hibooboo2: if it doesn't start with bearer and has a space it is not a jwt or a id for one so we want to return null. We don't want an exception because we throw those in a different place.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
hibooboo2: It should be toparse good catch.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
hibooboo2: The function checks if it is empty.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/GithubIdentitySearchProvider.java
hibooboo2: The reason i am making this change is that if i get to get the jwt and it is not set it will throw an error. We don't want that. because we can first attempt to create a jwt for them and if we cant then we just return no identities. The old code in this file would blow up before it could attempt to make the jwt. This will not blow up unless they do not have access based on their jwt anymore.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/internal/rancher/RancherIdentitySearchProvider.java
hibooboo2: The change is so that when auth is on a request still gets a rancher_id with its set of identities in the policy. Without that identity it cannot access projects that are rancher_id member only which breaks apiKeys ability to access projects and do things they might want / need to be able to do.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
sonchang: If we return null, we're passing that null to retrieveJwt(toParse) which ends up passing it to authTokenDao.getTokenByKey(jwtKey) which ends up passing a null into an SQL query

sonchang:
LGTM

PR #778, opened 2015-08-20
Fix an issue where when auth is off if you look at projects with github/ ldap members you get a 400.

hibooboo2:


alena1108:
LGTM

sonchang:
Sorry James.  I'm a little too lazy to trace back the code.  Can you briefly explain the 2 use cases for the untransform() and why we ignore the transformation in one case and not the other?  Thanks!

hibooboo2:
@sonchang the untransform method that changed really should be called transform as it calls transform which ultimately validates that the identity is valid. 
But in one case we want to throw an error another we do not.

hibooboo2:
The case where we do is if it is listing. don't is when they are creating members

sonchang:
Yeah.  I was able to gather that much from the code changes.  But why check for the listing but not when creating members?

hibooboo2:
i do check when listing but if it fails it isnt an error that should return 400 it should change the displayed result. But on creation it should 400

sonchang:
Got it.  Thanks!  LGTM

PR #779, opened 2015-08-20
Make simulator report proper instance state

cjellick:
A recent change to the ping logic to delete container that are not on host has exposed a flaw in our agent simulator. It was completely "forgetting" stopped instances and not reporting them, instead of reporting them as stopped.

### DIFF
--- a/code/implementation/simulator/agent-connection/src/main/java/io/cattle/platform/agent/connection/simulator/impl/SimulatorStartStopProcessor.java
+++ b/code/implementation/simulator/agent-connection/src/main/java/io/cattle/platform/agent/connection/simulator/impl/SimulatorStartStopProcessor.java
@@ -2,6 +2,7 @@ package io.cattle.platform.agent.connection.simulator.impl;
 
 import io.cattle.platform.agent.connection.simulator.AgentConnectionSimulator;
 import io.cattle.platform.agent.connection.simulator.AgentSimulatorEventProcessor;
+import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.eventing.model.Event;
 import io.cattle.platform.eventing.model.EventVO;
 import io.cattle.platform.json.JsonMapper;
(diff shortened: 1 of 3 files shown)


cjellick:
This could cause random (but pretty often) test failures that were checking to see if a simulated instance stayed stopped after being stopped because the ping logic was removing it.

code/implementation/simulator/agent-connection/src/main/java/io/cattle/platform/agent/connection/simulator/impl/SimulatorStartStopProcessor.java
```diff
@@ -44,9 +45,9 @@ public Event handle(final AgentConnectionSimulator simulator, Event event) throw
         final Object uuid = CollectionUtils.getNestedValue(event.getData(), "instanceHostMap", "instance", "uuid");
         if (uuid != null) {
             if (add && !FORGET.matcher(eventString).matches()) {
-                simulator.getInstances().add(uuid.toString());
+                simulator.getInstances().put(uuid.toString(), InstanceConstants.STATE_RUNNING);
             } else {
-                simulator.getInstances().remove(uuid.toString());
+                simulator.getInstances().put(uuid.toString(), InstanceConstants.STATE_STOPPED);
             }
```
alena1108: perhaps silly question: is Stopped is the only opposite state to Running in this context? Can it be be reported as Removed for example?

cjellick: Since the only events the simulator handles are "compute.instance.activate" "compute.instance.deactivate" (its in this file but hidden by the collapse), those are the only states we care about.

alena1108:
LGTM

cjellick:
Dont merge until we have a green build on it. I wanted to see what else failed :-)

sonchang:
LGTM

sonchang:
LGTM

PR #783, opened 2015-08-21
Retrive token that is the newest non expired token.

hibooboo2:


sonchang:
LGTM.  Feel bad missing it the first time around

cjellick:
This brings up the fact that we are definitely missing tests that exercise token expiration. I'm trying to think how we could test this.

cjellick:
I will LGTM so as to unblock the release an this is obviously the right fix, but we should figure out how to explicitly test token expiration.

hibooboo2:
Validation test. That changes the token expiry setting to something small with api. Then trys  to auth with expired token. I will right a validation test and submit a pr right now.

hibooboo2:
Actually I will do it as an integration test and it can be tested when we implement localauth.

cjellick:
That would immediately expire all tokens and if another token were used (if we run concurrently) those tests would fail, right?

PR #789, opened 2015-08-22
Have simulator report externalId

cjellick:


tests/integration/cattletest/core/test_allocation.py
```diff
@@ -130,6 +130,10 @@ def test_allocation_failed_on_start(super_client, new_context):
     assert c1.transitioning == 'no'
     assert c1.transitioningMessage is None
 
+    c2 = client.wait_success(c2.remove())
```
cjellick: Why the change to this test? Because this test sets a host's total compute to one, deploys two containers to the host, deallocates one container from the host and verifies that the other container can be allocated properly. However, the container that was deallocated is actually still on the host. So, it is reported to the Ping montior, but the ping monitor doesnt think it is on the host because it doesnt have an instance host map entry. So, the ping logic does a no-op start for the container. But, because the host's compute is set to one, that always fails the allocation step.
So, this just cleans up the container at the end of the test so that we can avoid a bunch of allocation failure errors while we wait for the context to be cleaned up.

cjellick:
Just test_lb_instance.py::test_validate_ports failed. Which is an acknowledged known failure.

alena1108:
LGTM

PR #794, opened 2015-08-25
CreateIndex for service instances

alena1108:
Put createIndex on a service container indicating the order in which it was launched. It doesn't have to be continuous, just always growing.

@ibuildthecloud I preferred to read last index from the services containers by selecting instancd.create_index field, over the approach where we store current create_index on the service object itself (to eliminate one extra db update - service update with the updated index). But in case we need to display the last index on the service's basis  - either in metadata or API - let me know, and I can rework the logic.


ibuildthecloud:
@alena1108 I think it's actually easier and more consistent to have the column on service as opposed to selecting all the instances and finding the next.  If you inject to your DAO the lockingConfiguration (Jooq object) and then do select (using the locking config), setCreateIndex(getCreateIndex() + 1), and then record.persist() it will do the opportunistic locking for you.

alena1108:
@ibuildthecloud thank you for the tip. Will update the PR

alena1108:
@ibuildthecloud reworked

PR #795, opened 2015-08-25
Remove stuff older than 16 hrs.

hibooboo2:


code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanUp.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import com.netflix.config.DynamicLongProperty;
+
+
+public class TableCleanUp extends AbstractJooqDao {
+
+    private static final Log logger = LogFactory.getLog(TableCleanUp.class);
+    private static final DynamicLongProperty CLEAN_UP_TIME = ArchaiusUtil.getLong("api.table.row.age");
```
ibuildthecloud: Change to something consistent with "task.purge.after.seconds",   I'd want to break this out into two settings.  process_instance.purge.after.seconds and events.purge.after.seconds

code/iaas/model/src/main/resources/META-INF/cattle/core-model/defaults.properties
```diff
@@ -17,3 +17,5 @@ object.link.ignore.serviceexposemaprecord.accountid=true
 object.link.ignore.serviceconsumemaprecord.accountid=true
 object.link.ignore.serviceexposehostmaprecord.accountid=true
 object.link.name.hostlabelmaprecord.labelid.override=hostLabels
+
+api.table.row.age=57600000
```
ibuildthecloud: Need to add setting "task.database.cleanup.schedule"

ibuildthecloud: Oops, sorry.  I see that above

code/iaas/config-item/server/src/main/resources/META-INF/cattle/config-server/config-item-defaults.properties
```diff
@@ -5,6 +5,7 @@ task.retry.timeout.schedule=1
 
 task.config.item.migration.schedule=900
 task.config.item.sync.schedule=15
+task.database.cleanup.schedule=15
```
ibuildthecloud: Let run this every 15 minutes.

PR #799, opened 2015-08-26
Put rancher-metadata to default section of answers.json

alena1108:
So standalone containers can also resolve it


cjellick:
LGTM

PR #800, opened 2015-08-26
Metadata fixes

alena1108:
1) Populate stack and service info for the instance.
Also use one method for getting service and non service instance using outer join with service/instance map table, instead of having separate methods. Because latter caused the bug when service instance was returned as a standalone one, and it caused duplicated entries for the same instance to be present in the answers.json file

https://github.com/rancher/rancher/issues/1880
https://github.com/rancher/rancher/issues/1882

2) Fixed bug in setting ports metadata for service and container

https://github.com/rancher/rancher/issues/1883


cjellick:
LGTM

sonchang:
LGTM

PR #801, opened 2015-08-26
More metadata bug fixes

alena1108:
1) Include stack name to the link name:

https://github.com/rancher/rancher/issues/1887

looks like https://github.com/rancher/rancher/issues/1889 has been fixed along the way

2) Fixed swapped stack_name and service_name for container:

https://github.com/rancher/rancher/issues/1888

3) Don't include createIndex to service metadata, keep it in container metadata only as its more relevant there. We can always add it on per user request later.

https://github.com/rancher/rancher/issues/1891

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/impl/ServiceMetadataInfoFactory.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/impl/ServiceMetadataInfoFactory.java
@@ -84,7 +84,10 @@ public class ServiceMetadataInfoFactory extends AbstractAgentBaseContextFactory 
         
         List<ServiceMetaData> servicesMD = new ArrayList<>();
         for (Map<String, ServiceMetaData> service : services.values()) {
-            servicesMD.addAll(service.values());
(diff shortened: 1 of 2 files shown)


code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/impl/ServiceMetadataInfoFactory.java
```diff
@@ -210,16 +211,19 @@ protected void getSidekicksInfo(Service service, List<String> sidekicks, List<St
         }
     }
 
-    protected void getLinksInfo(Map<Long, Service> idToService, Service service, Map<String, String> links) {
-        List<? extends ServiceConsumeMap> consumedMaps = consumeMapDao.findConsumedServices(service.getId());
+    protected void setLinksInfo(Map<Long, Map<String, ServiceMetaData>> services,
```
sonchang: super tiny, but looking at the method signature, I wouldn't have known it modified the contents of serviceMD.
But possibly rename to setLinksInfoInServiceMetaData() or setServiceMetaDataLinks()?

sonchang:
LGTM

PR #806, opened 2015-08-27
Null check when trying to check access on a projectMember.

hibooboo2:
rancher/rancher#1826

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectMemberResourceManager.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectMemberResourceManager.java
@@ -55,14 +55,23 @@ public class ProjectMemberResourceManager extends AbstractObjectResourceManager 
         Policy policy = (Policy) ApiContext.getContext().getPolicy();
         String id = RequestUtils.makeSingularStringIfCan(criteria.get("id"));
         if (StringUtils.isNotEmpty(id)) {
-            ProjectMember projectMember = authDao.getProjectMember(Long.valueOf(id));
+            ProjectMember projectMember;
+            try {
+                 projectMember = authDao.getProjectMember(Long.valueOf(id));
+            } catch (NumberFormatException e) {
+                throw new ClientVisibleException(ResponseCodes.NOT_FOUND);
+            }
+            if (projectMember == null) {
+                throw new ClientVisibleException(ResponseCodes.NOT_FOUND);
+            }
             if (!authDao.hasAccessToProject(projectMember.getProjectId(), policy.getAccountId(),
                     policy.isOption(Policy.AUTHORIZED_FOR_ALL_ACCOUNTS), policy.getIdentities())) {
                 throw new ClientVisibleException(ResponseCodes.NOT_FOUND);
             }
-            Identity identity  = identityManager.projectMemberToIdentity(projectMember);
+            Identity identity = identityManager.projectMemberToIdentity(projectMember);
             policy.grantObjectAccess(identity);
             return Arrays.asList(identity);
+
         }
         String projectId = RequestUtils.makeSingularStringIfCan(criteria.get("projectId"));
         List<? extends ProjectMember> members;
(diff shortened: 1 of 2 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectMemberResourceManager.java
```diff
@@ -55,14 +55,23 @@ protected Object listInternal(SchemaFactory schemaFactory, String type, Map<Obje
         Policy policy = (Policy) ApiContext.getContext().getPolicy();
         String id = RequestUtils.makeSingularStringIfCan(criteria.get("id"));
         if (StringUtils.isNotEmpty(id)) {
-            ProjectMember projectMember = authDao.getProjectMember(Long.valueOf(id));
+            ProjectMember projectMember;
+            try {
+                 projectMember = authDao.getProjectMember(Long.valueOf(id));
+            } catch (NumberFormatException e) {
+                return null;
```
cjellick: In case of garbage, probably raise NOT_FOUND. Theoretically more secure since malicious users would not be able to differentiate between a legit id that isnt there and a garbage id.

cjellick: I think you just said that this will 404 higher up anyway. Is that right? Why not also raise Client visible here for consistency?

tests/integration/cattletest/core/test_projects.py
```diff
@@ -372,6 +372,28 @@ def test_project_deactivate(user_clients, project, members):
     assert project.state == 'inactive'
 
 
+def test_project_member_invalid(project, admin_user_client, members):
+    client = client_for_project(project, admin_user_client)
+    project.setmembers(members=members)
+    diff_members(members, get_plain_members(project.projectMembers()))
+    membersGot = get_plain_members(project.projectMembers())
```
cjellick: Can you convert all these two the python convention of underscores?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectMemberResourceManager.java
hibooboo2: Raising the exception ends the request. If it is a request that is using the manager internally it would end that request which is bad. The 404 that is raised here is auth related but if they try to list something that doesn't exist they shouldn't lose their request.
Since other places internally use the projectManager to list members.

tests/integration/cattletest/core/test_projects.py
```diff
@@ -372,6 +372,28 @@ def test_project_deactivate(user_clients, project, members):
...
+    for member in membersGot:
+        if (member['role'] == 'owner'):
+            newMembers.append(member)
+        else:
+            oldMembers.append(member)
+    project.setmembers(members=newMembers)
+    diff_members(newMembers, get_plain_members(project.projectMembers()))
+    with pytest.raises(ApiError) as e:
+        client.by_id('projectMember', 'garbageId')
+    assert e.value.error.status == 404
```
cjellick: This assert should be indented so that it is executed as part of the with. Same below.

cjellick: Nevermind. We do it this way every else in our tests. Guess I'm wrong

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectMemberResourceManager.java
cjellick: I don't get the inconsistency then. Why is it ok to raise the exception if projectMember is null but not ok to do so if the id is garbage? IMO, they should behave exactly the same.

hibooboo2: So you are right about garbage id. That should throw ClientVisible. But a null member should return null not throw the 404.

cjellick: OK, that makes more sense. I could live with that.

cjellick:
LGTM

ibuildthecloud:
LGTM

PR #807, opened 2015-08-27
fix host label scheduling with case sensitivity

sonchang:
@cjellick @alena1108 
Super tiny fix for https://github.com/rancher/rancher/issues/1844


alena1108:
LGTM :)

PR #808, opened 2015-08-27
add error state for environment resource

wlan0:
@ibuildthecloud

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
@@ -2,6 +2,7 @@ package io.cattle.platform.servicediscovery.api.filter;
 
 import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.dao.NetworkDao;
+import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.iaas.api.filter.common.AbstractDefaultResourceManagerFilter;
 import io.cattle.platform.object.ObjectManager;
@@ -26,6 +27,8 @@ import java.util.List;
 import java.util.Map;
 
 import javax.inject.Inject;
+
+import org.apache.commons.lang3.StringUtils;
 
 public class ServiceCreateValidationFilter extends AbstractDefaultResourceManagerFilter {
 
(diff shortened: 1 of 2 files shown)


wlan0:
The failures are unrelated tests.

alena1108:
LGTM, but will merge after validation or integration test(s) for new state transitions are in place

cjellick:
LGTM. Agree with @alena1108. 

@wlan0 Since you're going to test this via validation tests, I'm guessing this puts you in a position where this won't be merged until the rest rancher-compose-executor is merged or very close to merged.

Is that going to cause you any major problems? Seems like it should be ok.

wlan0:
@cjellick @alena1108 Thanks! Yes, it is fine to wait until rancher-compose-executor is merged.

alena1108:
Just realized we would need one more fix. In ServiceCreateValidationFilter, make sure that service can be added to the environment only when its in Active state.

wlan0:
@alena1108 I added a validation step for checking that the environment is not in error state when creating services.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
```diff
@@ -71,6 +76,14 @@ public Object create(String type, ApiRequest request, ResourceManager next) {
         return super.create(type, request, next);
     }
 
+    protected void validateEnvironment(Service service) {
+        Environment env = objectManager.loadResource(Environment.class, service.getEnvironmentId());
+        if (StringUtils.equals(env.getState(), "error")) {
```
alena1108: We should probably read error state from CommonStatesConstants instead of hardcoding it.

wlan0: I agree, but Error state is not defined in CommonStatesConstants. So, I hardcoded it.

alena1108: Oh, right, its a custom state.

wlan0:
The test failure is random/unrelated

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
ibuildthecloud: @wlan0 Just add "error" to the CommonStatesConstants and then use that.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/ServiceCreateValidationFilter.java
```diff
@@ -71,6 +76,14 @@ public Object create(String type, ApiRequest request, ResourceManager next) {
...
+            throw new ValidationErrorException(ValidationErrorCodes.ACTION_NOT_AVAILABLE,
```
ibuildthecloud: Is this an action?  @alena1108 When does this code get called?

PR #810, opened 2015-08-28
Add support for local db auth.

hibooboo2:


vincent99:
Passwords being stored in cleartext instead of hashed?  Think about how you'll cleanly support upgrading to a version that does hashing/encryption now.

ibuildthecloud:
We won't merge this before the encryption/hashing

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -69,27 +71,62 @@ public Account getAdminAccount() {
...
-                .where(ACCOUNT.STATE.eq(CommonStatesConstants.ACTIVE)
-                        .and(ACCOUNT.NAME.contains(name))
-                        .and(ACCOUNT.KIND.eq(AccountConstants.USER_KIND)
-                                .or(ACCOUNT.KIND.eq(AccountConstants.ADMIN_KIND))))
-                .orderBy(ACCOUNT.ID.asc()).fetch());
-        return accounts;
+    public List<Account> searchUsers(String username) {
+        try {
+            List<Account> accounts = new ArrayList<>();
+            accounts.addAll(create()
```
alena1108: Nit: instead of defining/populating List data structure here, you can just return the result of select directly:

```
return create().select..
``

File: code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java [hunk 2]
```diff
@@ -69,27 +71,62 @@ public Account getAdminAccount() {
...
+                    .join(CREDENTIAL)
+                    .on(CREDENTIAL.ACCOUNT_ID.eq(ACCOUNT.ID))
+                    .where(
+                            ACCOUNT.STATE.eq(CommonStatesConstants.ACTIVE)
+                                    .and(CREDENTIAL.STATE.eq(CommonStatesConstants.ACTIVE))
+                                    .and(CREDENTIAL.PUBLIC_VALUE.eq(username)))
+                    .and(CREDENTIAL.KIND.eq("password"))
+                    .fetchOneInto(AccountRecord.class);
+        } catch (InvalidResultException e) {
+            throw new ClientVisibleException(ResponseCodes.CONFLICT, "MultipleOfUsername");
```
alena1108: How would it be possible? Should we disallow creating users with the same name to begin with?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -150,7 +187,37 @@ public Account createAccount(String name, String kind, String externalId, String
...
+    public Identity getIdentity(Long id) {
+        Account account = getAccountById(id);
+        if (account == null || account.getKind().equalsIgnoreCase(ProjectConstants.TYPE)) {
+            return null;
+        }
+        Credential credential = create()
+                .selectFrom(CREDENTIAL)
+                .where(CREDENTIAL.KIND.equalIgnoreCase("password")
+                        .and(CREDENTIAL.ACCOUNT_ID.eq(id))
+                .and(CREDENTIAL.STATE.equalIgnoreCase(CommonStatesConstants.ACTIVE))).fetchAny();
```
alena1108: Want to make sure I understand this correctly. Here by fetchAny we don't care what identity is returned? So in case of multiple account credentials, it can be any of them? What use case requires this behavior?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
hibooboo2: No we do not. Credentials are not guaranteed to be unique. As such neither are Passwords.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
hibooboo2: You are correct the behavior here is odd. Not sure what to do in the case of multiple credentials of type password for a user. @ibuildthecloud @vincent99 @cjellick any ideas?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
vincent99: Can we just fix this; it's crazy for there to not be a unique index on publicValue.

hibooboo2: @ibuildthecloud Said we cannot make then unique index as you could have same publicvalue that has been deactivated. As per the last time i mentioned this to him with regards to credentials/ apikeys in general.

vincent99: If it's `inactive` then it's not deleted yet, so that publicValue *should* still be reserved.

@ibuildthecloud I reiterate that it is absurd and dangerous to complicate authentication logic with extra code to handle multiple credentials in multiple states for multiple accounts with the same publicValue.  Here be dragons.  There should be a hard unique on publicValue so there is only one possible entry to authenticate against.

Once a credential deleted/purged that publicValue *could* be eligible for reuse, but that's a feature we don't necessarily need.

 - The simple option is to just not support that, the name is taken forever.
 - If someone *really* wants that name, they can still go manipulate the database.
 - Later we can move the row to a separate non-unique table of deleted credentials for audit-trail/the cloud only scales up.
 - Or you could put in something hacky like on delete update the public value to "*unixTimestamp*oldValue", using a character that we don't allow for API keys or usernames.

tests/integration/cattletest/core/test_local_auth.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __call__(self, r):
+        # modify and return the request
+        r.headers['Authorization'] = 'Bearer ' + self.jwt
+        if self.prj_id is not None:
+            r.headers['X-API-Project-Id'] = self.prj_id
+        return r
+
+
+def turn_on_off_local_auth(admin_client, request):
```
cjellick: Is it safe to do this given that we run tests in parallel and any test doing this could be running at the same time as any other test in the suite?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
ibuildthecloud: Give the scope of this feature I agree w/ @vincent99.  @hibooboo2 just add a unique DB constraint.

tests/integration/cattletest/core/test_local_auth.py
```diff
@@ -0,0 +1,187 @@
...
+    assert len(identities) > 0
+    local = False
+    for identity in identities:
+        if (identity.externalIdType == 'rancher_id'):
+            local = True
+    assert local
+    return start_client
+
+
+def makeUser(admin_user_client, name_base='user '):
```
cjellick: no camel case

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -69,27 +71,60 @@ public Account getAdminAccount() {
...
+                    .select(ACCOUNT.fields())
+                    .from(ACCOUNT)
+                    .join(CREDENTIAL)
+                    .on(CREDENTIAL.ACCOUNT_ID.eq(ACCOUNT.ID))
+                    .where(
+                            ACCOUNT.STATE.eq(CommonStatesConstants.ACTIVE)
+                                    .and(CREDENTIAL.STATE.eq(CommonStatesConstants.ACTIVE))
+                                    .and(CREDENTIAL.PUBLIC_VALUE.eq(publicValue))
+                                    .and(CREDENTIAL.SECRET_VALUE.eq(secretValue)))
+                    .and(CREDENTIAL.KIND.equalIgnoreCase("password"))
```
ibuildthecloud: Please create a constant in CredentialConstants for "password"

tests/integration/cattletest/core/test_local_auth.py
hibooboo2: You are correct this could cause problems. Maybe these need to be validation tests?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
alena1108: @ibuildthecloud @vincent99 @hibooboo2 when adding a constraint, we should add handling for db upgrade scenario from the version where duplicates were allowed.

hibooboo2: @alena1108 How would you suggest handling the upgrade scenario?

alena1108: @hibooboo2 Thats a tricky one. Applying unique key would fail when duplicates are present. And in this case you can't run pre-upgrade cleanup on user's behalf as duplicates weren't introduced by some bug in our code where you would know which data is safe to remove; we've just let user create duplicated values, and now don't know which one of them is valid. 

@vincent99 @ibuildthecloud any tips on that part?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
vincent99: I would say not allow multiple `password` credentials for a user in the API.  That will have to be enforced in the API though, simple db uniqueness won't work.  Then when you're reading them back, if you get more than one row (which should be impossible), bail out.

I know this differs from what we wrote out yesterday, and probably not easy to implement in the framework.  But for discussion, as an alternative to `POST /v1/passwords` to create and `GET /v1/passwords?accountId=:id` + `POST /v1/passwords/:id?action=changeSecret` to create and change passwords, we could have a single `/v1/accounts/:id?action=setPassword {publicValue: 'me', secretValue: 'newPass', [oldSecretValue: 'oldPass']}` action that creates-or-updates the `password` resource.  Then the API consumer doesn't have to know about the separate resource at all and it's clear that there is a single user/password for that account.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
vincent99: What happens when it makes the alter and fails?  Does cattle not start up?  Will it continue and work if you then remove the duplicate from the database and start again?

If it does something rational like that, then I say put it as a footnote in the release notes "if you see `<this error>` on upgrade, use `<steps>` to remove duplicates".  The table is used for apikeys today, and only admin has permission to create specific values, and most people outside the company wouldn't know it was possible to begin with.  So there should be no to very few people with actual duplicates in their database.

hibooboo2: @vincent99 I agree with this method not sure how @ibuildthecloud feels, or even maybe how @will-chan would feel. @alena1108 What do you say?

tests/integration/cattletest/core/test_local_auth.py
ibuildthecloud: @cjellick @hibooboo2 Once the tests start the apikeys are setup so flipping auth on/off shouldn't generally be a problem.  Could be a problem for these auth tests though.  Does this need to be fixed?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/GithubTokenCreator.java
```diff
@@ -72,23 +71,9 @@ public Token getGithubToken(String accessToken) {
...
-                    }
-                }
-            }
-        } else {
-            account = authDao.getAdminAccount();
-            authDao.updateAccount(account, null, AccountConstants.ADMIN_KIND, userAccountInfo.getAccountId(), GithubConstants.USER_SCOPE);
-            authDao.ensureAllProjectsHaveNonRancherIdMembers(userAccountInfo.toIdentity(GithubConstants.USER_SCOPE));
+        Account account = githubUtils.getOrCreateAccount(user, identities, null, true);
+        if (account == null){
+            throw new ClientVisibleException(ResponseCodes.INTERNAL_SERVER_ERROR, "FailedToGetAccount");
```
ibuildthecloud: Should this really be a 500?  401?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/ChangeSecretActionHandler.java
```diff
@@ -0,0 +1,16 @@
+package io.cattle.platform.iaas.api.auth.integration.local;
+
+import io.cattle.platform.api.action.ActionHandler;
+import io.github.ibuildthecloud.gdapi.request.ApiRequest;
+
+public class ChangeSecretActionHandler implements ActionHandler {
+    @Override
+    public Object perform(String name, Object obj, ApiRequest request) {
+        return null;
```
ibuildthecloud: Implementation TBD?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/LocalAuthConfigManager.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    protected Object createInternal(String type, ApiRequest request) {
+        if (!StringUtils.equalsIgnoreCase(LocalAuthConstants.CONFIG, request.getType())) {
+            return null;
+        }
+        Map<String, Object> config = CollectionUtils.toMap(request.getRequestObject());
+        String username = (String) config.get("username");
```
ibuildthecloud: FYI, if LocalAuthConfig was an interface you could just do `request.proxyRequestObject(LocalAuthConfig.class)` and then you can just do `goo.getName()` and not have to be casting all these strings and hardcoding fields names.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/LocalAuthTokenCreator.java
```diff
@@ -0,0 +1,85 @@
...
+    AuthDao authDao;
+    @Inject
+    TokenService tokenService;
+    @Inject
+    LocalAuthUtils localAuthUtils;
+    @Inject
+    ObjectManager objectManager;
+
+    @Override
+    public Token getToken(ApiRequest request) {
```
ibuildthecloud: White space man... It's free.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/LocalAuthUtils.java
```diff
@@ -0,0 +1,33 @@
+package io.cattle.platform.iaas.api.auth.integration.local;
+
+import io.cattle.platform.core.constants.ProjectConstants;
+import io.cattle.platform.iaas.api.auth.TokenUtils;
+
+import java.util.List;
+
+public class LocalAuthUtils extends TokenUtils {
```
ibuildthecloud: Not that you have to change this, but most java developers expect classes that are named `*Utils` to just have static methods and not actually be instantiated.

tests/integration/cattletest/core/test_local_auth.py
```diff
@@ -0,0 +1,187 @@
+from common_fixtures import *  # NOQA
```
ibuildthecloud: Some whitespace would help.  Most methods just look like blobs of code.

tests/integration/cattletest/core/test_local_auth.py
hibooboo2: Yes it does will be a problem for the auth tests. I  think these may need to be validation tests.

PR #813, opened 2015-08-28
Service selectors

alena1108:
2 selector kinds are supported for the service, here are the API object fields:

* selectorLink
* selectorContainer

First one defines which services get linked to it. Second selector defines the criteria that container has to satisfy in order to join the service.

Selector format: 

```
key in (value1, value2)
key notin (value1, value2)
key
key = value
key != value
and any combination of them can be specified in coma separated way:
key != value, key in (value1, value2), key = value
```
 
For selectorLink, validation is done against labels defined on the service; for selectorContainer  - labels defined on the container.

Service can be:

* a combination of managed - containers started based on service launchConfig/scale - and unmanaged - containers joined service based on selectors
* selector only service identified by presence of following param in launch config:

```
image: rancher/none
```

Containers joining the service based on selectors, are not managed by the service replication controller. It means:

* service.deactivate won't stop the container
* service.reconcile won't restart the container if it dies
* service.remove won't remove the unmanaged container; it will just get disassociated from the service, and will continue to run as standalone. 

Selectors get represented in docker-compose.yml as labels.

```
io.rancher.service.selector.container: foo=bar
io.rancher.service.selector.link: foo in (bar)
```

https://github.com/rancher/rancher/issues/2239

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/ServiceDeploymentPlannerFactoryImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/ServiceDeploymentPlannerFactoryImpl.java
@@ -1,5 +1,6 @@
 package io.cattle.platform.servicediscovery.deployment.impl;
 
+import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.servicediscovery.api.constants.ServiceDiscoveryConstants;
 import io.cattle.platform.servicediscovery.api.util.ServiceDiscoveryUtil;
@@ -21,17 +22,34 @@ public class ServiceDeploymentPlannerFactoryImpl implements ServiceDeploymentPla
         }
 
         Service service = services.get(0);
-        Map<String, String> serviceLabels = ServiceDiscoveryUtil.getServiceLabels(service, context.allocatorService);
-        String globalService = serviceLabels.get(ServiceDiscoveryConstants.LABEL_SERVICE_GLOBAL);
-
-        if (service.getKind().equalsIgnoreCase(ServiceDiscoveryConstants.KIND.EXTERNALSERVICE.name())
+        boolean isGlobalDeploymentStrategy = isGlobalDeploymentStrategy(context, service);
+        boolean isSelectorOnlyStrategy = isSelectorOnlyStrategy(context, service);
+        if (isSelectorOnlyStrategy
+                || service.getKind().equalsIgnoreCase(ServiceDiscoveryConstants.KIND.EXTERNALSERVICE.name())
                 || service.getKind().equalsIgnoreCase(ServiceDiscoveryConstants.KIND.DNSSERVICE.name())) {
-            return new ExternalServiceDeploymentPlanner(services, units, context);
-        } else if (globalService != null) {
+            return new NoOpServiceDeploymentPlanner(services, units, context);
+        } else if (isGlobalDeploymentStrategy) {
             return new GlobalServiceDeploymentPlanner(services, units, context);
         } else {
             return new DefaultServiceDeploymentPlanner(services, units, context);
         }
     }
 
+    protected boolean isGlobalDeploymentStrategy(DeploymentServiceContext context, Service service) {
+        Map<String, String> serviceLabels = ServiceDiscoveryUtil.getServiceLabels(service, context.allocatorService);
+        String globalService = serviceLabels.get(ServiceDiscoveryConstants.LABEL_SERVICE_GLOBAL);
+        return globalService != null;
+    }
+
+    protected boolean isSelectorOnlyStrategy(DeploymentServiceContext context, Service service) {
+        boolean selectorOnly = false;
+        Object imageUUID = ServiceDiscoveryUtil.getServiceDataAsMap(service,
+                ServiceDiscoveryConstants.PRIMARY_LAUNCH_CONFIG_NAME,
+                context.allocatorService).get(InstanceConstants.FIELD_IMAGE_UUID);
+        if (service.getSelectorContainer() != null
+                && (imageUUID == null || imageUUID.toString().equalsIgnoreCase(ServiceDiscoveryConstants.IMAGE_NONE))) {
+            selectorOnly = true;
+        }
+        return selectorOnly;
+    }
 }
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/SelectorInstanceCreatePostListener.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/SelectorInstanceCreatePostListener.java
@@ -1,0 +1,80 @@
+package io.cattle.platform.servicediscovery.process;
+
+import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
+import io.cattle.platform.core.constants.CommonStatesConstants;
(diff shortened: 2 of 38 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/ServiceDeploymentPlannerFactoryImpl.java
```diff
@@ -21,17 +22,34 @@ public ServiceDeploymentPlanner createServiceDeploymentPlanner(List<Service> ser
...
-        } else if (globalService != null) {
+            return new NoOpServiceDeploymentPlanner(services, units, context);
+        } else if (isGlobalDeploymentStrategy != null) {
             return new GlobalServiceDeploymentPlanner(services, units, context);
         } else {
             return new DefaultServiceDeploymentPlanner(services, units, context);
         }
     }
 
+    protected String isGlobalDeploymentStrategy(DeploymentServiceContext context, Service service) {
```
cjellick: odd to have an is* method that returns a string

resources/content/db/changelog.xml
```diff
@@ -64,5 +64,7 @@
     <include file="db/core-057.xml"/>
     <include file="db/core-058.xml"/>
     <include file="db/core-059.xml"/>
+    <include file="db/core-060.xml"/>
```
cjellick: Guessing you know this already, but 60 is already on master, so I guess you need to rebase.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/ServiceDeploymentPlannerFactoryImpl.java
alena1108: Will fix, thank you

resources/content/db/changelog.xml
alena1108: Ah, thanks :) was just trying to please @ibuildthecloud who dislikes me using custom names before the final rebase ;)

alena1108:
@ibuildthecloud @cjellick could you please make a final review? Extra fixes Craig has made, are going to be merged as a part of his PR.

cjellick:
@alena1108 Here's another thing that @ibuildthecloud and I noticed: in the UI, a service is not dynamically updated with the new containers that joined the service based on their labels. Meaning in this expected flow:
1. Create selector based service
2. Create container with matching label
3. Container is added to service
4. UI for service automatically shows new container
The 4th step doesnt happened.

@ibuildthecloud said that this is likely due to the fact that we are not generating a service udpate event for the UI to consume. He said that triggering a service reconcile could fix this. But please talk to him about it because I may be misinterpreting what he aid.

alena1108:
@cjellick spoke to @ibuildthecloud about it. Calling service reconcile won't work as the underlying logic will never run - and will never update the service state - as it will think that there is nothing to reconcile because instances joining the service are not deployed as a part of the original service deployment. What we will do instead is - trigger state.change event for the service directly from the code where instance joins the service.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/SelectorInstanceCreatePostListener.java
```diff
@@ -0,0 +1,63 @@
...
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        Instance instance = (Instance) state.getResource();
+        List<Service> services = objectManager.find(Service.class, SERVICE.ACCOUNT_ID, instance.getAccountId(),
+                SERVICE.REMOVED, null);
+
+        for (Service service : services) {
+            if (sdService.isServiceInstance(service, instance)) {
+                continue;
+            }
+            if (sdService.isSelectorContainerMatch(service, instance.getId())) {
```
cjellick: Are you going to add the state.change event in this PR or another?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/selector/SelectorUtils.java
```diff
@@ -0,0 +1,137 @@
...
+            return key;
+        }
+
+        private List<String> getValue() {
+            return value;
+        }
+    }
+
+    public static boolean isSelectorMatch(String selector, Map<String, String> labels) {
+        List<SelectorConstraintDefinition> constraints = getSelectorConstraints(selector);
```
cjellick: Is it worth protecting against a constraints.length() == 0 situation or is that impossible/highly unlikely?
Theoretically, if constraints.length() == 0, you should just return false?

resources/content/db/changelog.xml
```diff
@@ -65,5 +65,6 @@
     <include file="db/core-058.xml"/>
     <include file="db/core-059.xml"/>
     <include file="db/core-060.xml"/>
+    <include file="db/core-100.xml"/>
```
cjellick: Need to move to 61

cjellick:
@alena1108 are you making the label merging change in this PR or a separate one?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/SelectorInstanceCreatePostListener.java
alena1108: this PR

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/selector/SelectorUtils.java
alena1108: Sure, wouldn't hurt to have.

resources/content/db/changelog.xml
alena1108: yup

alena1108:
@cjellick labels merging fix will come as a separate bug fixes PR (which will most likely get merged to master before this one)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/selector/SelectorUtils.java
```diff
@@ -0,0 +1,145 @@
...
+        // key
+        // key = value
+        // key != value
+        // and any combination of them can be specified in coma separated way:
+        // key != value, key in (value1, value2), key = value
+
+        // the code below temporarily replaces , with random uuid
+        // so we can split by "," to get all the selectors
+        boolean inList = false;
+        for (String s : selector.split("")) {
```
ibuildthecloud: Can you refactor to just scan through the char's and not rely on split.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/SelectorServiceCreatePostListener.java
```diff
@@ -0,0 +1,86 @@
...
+        for (Service targetService : services) {
+            // skip itself
+            if (targetService.getId().equals(service.getId())) {
+                continue;
+            }
+            if (sdService.isSelectorLinkMatch(service, targetService)) {
+                sdService.addServiceLink(service, new ServiceLink(targetService.getId(), null));
+            }
+            if (sdService.isSelectorLinkMatch(targetService, service)) {
+                sdService.addServiceLink(targetService, new ServiceLink(service.getId(), null));
```
ibuildthecloud: It might be possiblie to end up with two map entries in a race condition

PR #820, opened 2015-08-30
Slow query optimization

alena1108:
Optimized the query retrieving instances on host network by:

1) retrieving host network of the account and referencing it in the query by id, instead of verifying network.kind info in the query.
2) Added "removed=null" filter for every table participating in join
3) Excluded instance table from the join as we can get instanceId from instancehostmap table itself.

When you run mysql explain on a new sql statement, join type=All is no longer on the picture - fixed by 1). Also "rows" number is reduced - fixed by 2)


alena1108:
@ibuildthecloud

PR #825, opened 2015-08-31
Service discovery bug fixes

alena1108:
1) Error out when there is still a need to reconcile after the reconcile
is finished.

to avoid service.reconcile going into an infinite loop

https://github.com/rancher/rancher/issues/1933

2) RestartPolicy=No in service container context- respect the flag for containers only in HA cases - when
service.update is caused by service.reconcile triggers.

https://github.com/rancher/rancher/issues/1935

3) Define create-only service containers using label "io.rancher.container.start_once" (with value=true) instead of utilizing the restartPolicy param. Makes the bug below no longer relevant: 

https://github.com/rancher/rancher/issues/1752

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
@@ -68,6 +68,7 @@ public class ServiceDiscoveryConstants {
     public static final String LABEL_SERVICE_GLOBAL = "io.rancher.scheduler.global";
     public static final String LABEL_SERVICE_REQUESTED_HOST_ID = "io.rancher.service.requested.host.id";
     public static final String LABEL_SERVICE_LAUNCH_CONFIG = "io.rancher.service.launch.config";
+    public static final String LABEL_SERVICE_CONTAINER_CREATE_ONLY = "io.rancher.container.start_once";
 
     public static final String PRIMARY_LAUNCH_CONFIG_NAME = "io.rancher.service.primary.launch.config";
     public static final String LABEL_SIDEKICK = "io.rancher.sidekicks";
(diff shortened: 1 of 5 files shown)


code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/constants/ServiceDiscoveryConstants.java
```diff
@@ -68,6 +68,7 @@
     public static final String LABEL_SERVICE_GLOBAL = "io.rancher.scheduler.global";
     public static final String LABEL_SERVICE_REQUESTED_HOST_ID = "io.rancher.service.requested.host.id";
     public static final String LABEL_SERVICE_LAUNCH_CONFIG = "io.rancher.service.launch.config";
+    public static final String LABEL_SERVICE_CONTAINER_CREATE_ONLY = "io.rancher.service.container.start_once";
```
ibuildthecloud: Can you change to `io.rancher.container.start_once` to be more consistent with `io.rancher.container.hostname_override`

alena1108: sure

alena1108:
@ibuildthecloud Darren, don't merge this one yet, debugging the TF to see if its related to the changes I've made

ibuildthecloud:
@alena1108 is @vincent99 required to make any changes now that we have `start_once`?  Or is there some crappy logic he can remove now?

alena1108:
@ibuildthecloud  we can leave UI as it is because the only change @vincent99 made is to set radio button restartPolicy=Always by default. Changing it back to No won't make a difference as this param is disregarded in service reconcile anyway.

So  "start_once=true" behavior would be supported only by rancher-compose now unless someone asks for the UI support (this functionality will most likely be used by for data containers as a part of sidekick feature, which is not in UI yet)

PR #831, opened 2015-09-02
Add readonly role to projects.

hibooboo2:


tests/integration/cattletest/core/test_authorization.py
```diff
@@ -156,6 +158,23 @@ def test_project_types(project_client):
     test_user_types(project_client, adds={'subscribe'},
                     removes={'userPreference'})
 
+def test_readonly_types(admin_user_client):
+    context = create_context(admin_user_client, kind="readonly")
+    client = context.user_client
+    types = test_user_types(client, adds={'subscribe'},
+                            removes={'userPreference'})
+    for type in types:
```
ibuildthecloud: `admin_user_client.schema.types` is a map of schema name to schema object.  You should just read the create/update/delete fields from the schema.  Also check all the fields are r/o.

PR #834, opened 2015-09-03
bump machine, host-api and go-machine-service versions

cjellick:


ibuildthecloud:
@cjellick looks like a TF from @hibooboo2 local auth changes I merged last night.  Wonder if the auth tests should be ran nonparallel

PR #839, opened 2015-09-03
Prevent NPE in ContainerEventCreate

cjellick:
In simulated tests, we werent maintaing enough data in the simulated hosts. Specifically, image, but this PR maintains all the same info that a real host maintains.

### DIFF
--- a/code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
+++ b/code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
@@ -30,7 +30,9 @@ import io.cattle.platform.object.meta.ObjectMetaDataManager;
 import io.cattle.platform.object.process.ObjectProcessManager;
 import io.cattle.platform.object.process.StandardProcess;
 import io.cattle.platform.object.util.DataAccessor;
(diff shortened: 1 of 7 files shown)


code/iaas/ha/src/main/java/io/cattle/platform/ha/monitor/impl/PingInstancesMonitorImpl.java
```diff
@@ -224,6 +226,9 @@ void determineSyncActions(Map<String, KnownInstance> knownInstances, ReportedIns
                 ReportedInstance ri = new ReportedInstance();
                 ri.setExternalId(ki.getExternalId());
                 ri.setUuid(ki.getUuid());
+                Object imageUuid = CollectionUtils.getNestedValue(ki.getData(), DataUtils.FIELDS, FIELD_IMAGE_UUID);
```
alena1108: Craig, do you think we can be more force about certain fields being required for ReportedInstance (in order to avoid possible NPEs in follow up code)? May be present them as the fields in ReportedInstance constructor? Otherwise we might forget to set those fields in the future code where new events are generated.

cjellick: Yes. WIll do.

PR #840, opened 2015-09-03
Run flake8 before tests.

hibooboo2:


ibuildthecloud:
I don't want this change.

hibooboo2:
Why not?

ibuildthecloud:
Because we don't have notifications setup so fail fast behavior ends up wasting time.

PR #846, opened 2015-09-04
DSA private key bug fix: regex format change for .pem file generation on haproxy

alena1108:
To respect other key formats besides RSA

https://github.com/rancher/rancher/issues/1980


alena1108:
@ibuildthecloud @cjellick , could you please review? It needs to go to RC

PR #851, opened 2015-09-08
Filter publicvalue only allow unique.

hibooboo2:


ibuildthecloud:
@hibooboo2 flake8

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -72,6 +71,15 @@
                                             transitioning="deactivating" done="inactive" />
     <process:process name="agent.reconnect" resourceType="agent" start="active,activating" transitioning="reconnecting" done="active" />
 
+
+    <!-- Credential -->
+    <process:process name="credential.create" resourceType="credential" start="requested" transitioning="registering" done="inactive" delegate="credential.activate" />
```
ibuildthecloud: The `<process:defaultProcess />` element has an `exclude` attribute.  I would feel better if you did `exclude="restore"` as it is less error prone than manually defining all of these processes.

PR #854, opened 2015-09-09
Read on create only returns null instead of hiding field.

hibooboo2:


ibuildthecloud:
@hibooboo2 tests are failing

ibuildthecloud:
@hibooboo2 I don't quiet understand.  It seems like the old behavior is correct.  Why do we want to return null as opposed to not sending the field?  Is this causing a downstream issue?  @vincent99 Did you complain about this one?

hibooboo2:
It is causing isssues yes.

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/model/impl/WrappedResource.java
```diff
@@ -68,13 +68,16 @@ protected void init() {
...
             Field field = entry.getValue();
-            if (!field.isIncludeInList() || (!StringUtils.equalsIgnoreCase(method, "post") && field.isReadOnCreateOnly())) {
+            if (!field.isIncludeInList()) {
                 continue;
             }
             Object value = additionalFields.remove(name);
             if (value == null) {
                 value = field.getValue(obj);
             }
+            if (!StringUtils.equalsIgnoreCase(method, "post") && field.isReadOnCreateOnly()){
```
ibuildthecloud: There is a constant or method somewhere for "post"

ibuildthecloud:
sounds like there was discussion amongst @cjellick and @vincent99 regarding this.  I'm okay with the change then.

PR #857, opened 2015-09-09
Bug fix for default/optional certificate swap

alena1108:
When search for existing certificate, consider is_default flag as well
https://github.com/rancher/rancher/issues/2010


ibuildthecloud:
There is a failing test

alena1108:
@ibuildthecloud fixing

PR #859, opened 2015-09-10
Change default pictures to identicons.

hibooboo2:


code/framework/api/src/main/java/io/cattle/platform/api/auth/Identity.java
```diff
@@ -8,9 +9,12 @@
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.springframework.util.DigestUtils;
```
ibuildthecloud: ah.... don't import spring.  don't need to change, but don't import spring classes in the future.

hibooboo2: How do i use DigestUtils then?

ibuildthecloud: You don't.  Find a different approach.  We will eventually drop Spring.

kaos: :stuck_out_tongue_closed_eyes:

PR #861, opened 2015-09-10
Fix for setting truncated io.rancher.container.hostname_override when instance.name > 64 chars and a domainName is specified for a container

prachidamle:
If domainName is set on a container, the hostname considers the domainName as well. So fix is to truncate the instanceName further by including the domainName. The hostName should fit in 64 chars including the domainName.

rancher/rancher#1934

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/DefaultDeploymentUnitInstance.java
@@ -5,6 +5,7 @@ import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.core.model.ServiceExposeMap;
+import io.cattle.platform.docker.constants.DockerInstanceConstants;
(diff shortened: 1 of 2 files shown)


prachidamle:
@ibuildthecloud, @cjellick Can you review and merge this one please, thanks!

ibuildthecloud:
@prachidamle I'm confused, why do we need to check domain name?

prachidamle:
Looks like when domainName + instanceName is exceeding 64 chars, the container does not start. @sangeethah reopened rancher/rancher#1934 for this case.

So somewhere down the line, we must be adding domainName to the hostname...?

ibuildthecloud:
@prachidamle I would guess that domainName itself can't be longer than 64 characters.  for that though we don't need code.  we can set the maxLength to 64 in the schema

prachidamle:
@ibuildthecloud the bug is not about domainName being > 64.

When hostName_override flag is set AND (domainName + <stackName_serviceNam_serviceNumber>)   is exceeding 64 chars, then container does not start.

Original fix truncated <stackName_serviceNam_serviceNumber> to 64.
But @sangeethah used a domainName for her container and the fix did not work. So need to consider the domainName while truncating.

prachidamle:
@ibuildthecloud  Although, the other question is why is domainName being used in hostName, which I am not sure where to check.

ibuildthecloud:
@prachidamle Okay, I just looked at Docker they do `${hostname}.${domain}` in the call to set a hostname.  So this PR makes sense now.

PR #867, opened 2015-09-11
Delete projectMember entries on account purge.

hibooboo2:


hibooboo2:
This will fix rancher/rancher#2021

PR #871, opened 2015-09-14
use auth in websocket test

cjellick:


ibuildthecloud:
@hibooboo2 This build passed because we were missing auth, but the odd thing is why is auth turned on?  Do you know how auth gets turned on in the tests?  I thought it was off.

PR #890, opened 2015-09-22
Fixed DNS programming for the accounts created before Rancher beta

alena1108:
The account were missing networks, and the DNS code used to rely on the
fact that docker host network support is in

https://github.com/rancher/rancher/issues/2024


alena1108:
@ibuildthecloud

PR #898, opened 2015-09-24
InService upgrade

alena1108:
API changes:

* 2 new parameters to service.upgrade API: launchConfig (type=launchConfig) and secondaryLaunchConfigs (type=array[launchConfig])
* service.upgarde - toService and finalScale are no longer required. toService and launchConfig/secondaryLaunchConfigs are mutually exclusive. First option implies to-service upgrade, second - in service
* new field - version - got added to service.launchConfig and instance objects. If the parameter varies for service launchConfig and its instances, it means that instances are yet to be upgraded.

As no parameters can actually be changed on the service's launch config today (like image, other deploy params), so launchConfig presence itself determines that the inService upgrade needs to be done

In service upgrade can be performed in batch size, just the way its done for a regular upgrade. One thing to be aware of - for multi launch config service (service with sidekicks), batch = deployment Unit. Meaning that instances of the same deployment unit will get removed/redeployed together, in a batch.

https://github.com/rancher/rancher/issues/1874

@vincent99 @ibuildthecloud


ibuildthecloud:
@alena1108 Regarding batch size, that means if you have a sidekick and do a batch size of 1 (and upgrade both launchConfigs), you will see two containers get deleted at a time because of this?

alena1108:
@ibuildthecloud yes

PR #900, opened 2015-09-25
Increased ResourceLockChange timeout

alena1108:
Also removed logging exception cause for the exceptions logged in debug
mode - @ibuildthecloud need your opinion on this particular change. I've put this to omit logging stack traces for exceptions like io.cattle.platform.lock.exception.FailedToAcquireLockException

https://github.com/rancher/rancher/issues/1996


ibuildthecloud:
Could you at least log the cause message if the cause is not null?  I believe the cause is usually the most important part.

alena1108:
test_container_request_ip has failed

PR #907, opened 2015-09-29
Bug fixes

alena1108:
1) Don't export "scale" if global deployment flag is present

https://github.com/rancher/rancher/issues/1842

2) Made certificate name unique within the account scope as rancher-compose accesses certificates by their names, and we have to guard against having duplicates:

https://github.com/rancher/rancher/issues/2042


ibuildthecloud:
@alena1108 Can you just add "unique: true" to the json for certs?  Does that not work?

ibuildthecloud:
On the name field

alena1108:
@ibuildthecloud thank you, fixed and added more tests

PR #911, opened 2015-09-30
More stuff to metadata

alena1108:
1) Added /self/host and /hosts sections to defaults
2) Added "host_name" field to the container object (similarly we do  for service_name and stack_name)

```
  "create_index": 1,
   "service_name": "alena",
    "stack_name": "Default"
    "host_name": "myhost",
    "ips": [
        "10.42.34.58"
    ],.....
```

@ibuildthecloud @vincent99 - regarding the second change - not sure if we should expect host name to be unique within the account, and whether if we should instead introduce "uuid"/"host_uuid" to host/container objects respectively like @cjellick suggested earlier. 

@cloudnautique 

https://github.com/rancher/rancher/issues/2189

UPDATE: 2) won't be added as a part of this PR; we can add it on per request later (if use case comes up)

One more item is added as a part of the PR - "host_uuid" field to host object

### DIFF
--- a/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/service/impl/HostCreateRemoveNicLookup.java
+++ b/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/service/impl/HostCreateRemoveNicLookup.java
@@ -1,0 +1,58 @@
+package io.cattle.platform.agent.instance.service.impl;
+
+import static io.cattle.platform.core.model.tables.InstanceHostMapTable.INSTANCE_HOST_MAP;
+import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
+import static io.cattle.platform.core.model.tables.NetworkTable.NETWORK;
+import static io.cattle.platform.core.model.tables.NicTable.NIC;
+import io.cattle.platform.agent.instance.service.InstanceNicLookup;
+import io.cattle.platform.core.constants.NetworkConstants;
+import io.cattle.platform.core.model.Host;
+import io.cattle.platform.core.model.Network;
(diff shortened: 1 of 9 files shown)


cjellick:
@alena1108 I would still like to see uuid. I think I need the guaranteed uniqueness. We can discuss quickly after standup

cjellick:
This was the output I got for /hosts:
```
root@fusion-171:/# curl -i http://rancher-metadata/latest/hosts
HTTP/1.1 200 OK
Date: Thu, 01 Oct 2015 17:38:55 GMT
Content-Length: 284
Content-Type: text/plain; charset=utf-8

0=fusion-171
1=fusion-171
2=fusion-171
3=fusion-171
4=fusion-171
5=fusion-171
6=fusion-171
7=fusion-171
8=fusion-171
9=fusion-171
10=fusion-171
11=fusion-171
12=fusion-171
13=fusion-171
14=fusion-171
15=fusion-171
16=fusion-171
17=fusion-171
18=fusion-171
19=fusion-171
20=fusion-171
```
...I only have a single entry in the host table

alena1108:
@cjellick hm, let me look at it

alena1108:
@cjellick Looks like one more fix is needed as a part of this PR - metadata should be updated on the host.create event to update /hosts target

cjellick:
LGTM

cjellick:
`test_anti_affinity_sidekick` failed. Any chance its related?

alena1108:
@cjellick most likely not, I'm going to re-run the test though just in case.

alena1108:
Realized that host_uuid field on the container is something that is needed for external-dns feature. About to add it in.

ibuildthecloud:
@alena1108 Can you change `host_name` to `hostname`?

PR #919, opened 2015-10-05
Add dynamic capabilities to apicontext.

hibooboo2:
Capabilities added this way take precedence over existing ones. Meaning that old capabilities are not used.

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
@@ -459,7 +459,9 @@ public abstract class AbstractObjectResourceManager extends AbstractBaseResource
         String state = ObjectUtils.toString(attributes.get(ObjectMetaDataManager.STATE_FIELD), null);
         String currentState = io.cattle.platform.object.util.ObjectUtils.getState(obj);
 
-        if (!StringUtils.isBlank(capability) && !DataAccessor.fieldStringList(obj, ObjectMetaDataManager.CAPABILITIES_FIELD).contains(capability)) {
+        if (!StringUtils.isBlank(capability) && !(ApiContext.getContext().getCapabilities(obj) != null ?
+                ApiContext.getContext().getCapabilities(obj).contains(capability) :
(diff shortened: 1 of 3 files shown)


code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
```diff
@@ -459,7 +459,9 @@ protected boolean isValidAction(Object obj, Action action) {
         String state = ObjectUtils.toString(attributes.get(ObjectMetaDataManager.STATE_FIELD), null);
         String currentState = io.cattle.platform.object.util.ObjectUtils.getState(obj);
 
-        if (!StringUtils.isBlank(capability) && !DataAccessor.fieldStringList(obj, ObjectMetaDataManager.CAPABILITIES_FIELD).contains(capability)) {
+        if (!StringUtils.isBlank(capability) && !(ApiContext.getContext().getCapabilities(obj) != null ?
```
ibuildthecloud: Why don't you just make (ApiContext.getContext().getCapabilities(obj) never return null.  Then you don't need this null check

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/context/ApiContext.java
```diff
@@ -91,4 +97,19 @@ public TransformationService getTransformationService() {
     public void setTransformationService(TransformationService transformationService) {
         this.transformationService = transformationService;
     }
+
+    public void addCapability(Object object, String capability) {
+        if (dynamicCapabilities.get(object) == null){
```
ibuildthecloud: We have a dumb util method for this: `io.cattle.platform.util.type.CollectionUtils.addToMap(Map<K, V>, K, T, Class<? extends Collection>)`

code/framework/api/src/main/java/io/cattle/platform/api/resource/AbstractObjectResourceManager.java
hibooboo2: The null check allows me to use it if they aren't null and if it is null it uses the one on the object. That is why.

PR #921, opened 2015-10-06
Fix issue creating extra apikeys.

hibooboo2:
No apikeys created for users and admins by default.
Address rancher/rancher#2103


ibuildthecloud:
@hibooboo2 There are test failures

PR #934, opened 2015-10-09
Fixed regression bug in selectors code

alena1108:
Making LB instances fail to deploy as no image is being set in launch config for those. Integration test didn't caught it as for LB instance we are always adding simulator image.

https://github.com/rancher/rancher/issues/2284

TBD - in the future as a part of LB refactoring, we should probably update launch config with the network-instance image information.


cjellick:
LGTM

PR #935, opened 2015-10-09
Rollback support for insservice upgrade

alena1108:
Rollback support + Strategies (inService/toService) to the upgrade method

https://github.com/rancher/rancher/issues/2277
https://github.com/rancher/rancher/issues/1874

API changes
=============
* New API methods on service:

```
rollback () - service will get rolled back with the same interval/batch size as the upgrade was done with
cancelrollback()
finishUpgrade
```

Other changes:
===========
* During inservice upgrade, create a new instances' batch before acting on old ones (used to be the other way around, it it resulted in service being degraded): https://github.com/rancher/rancher/issues/2324

* Stop old instances instead of removing them. Removal is triggered when "finishUpgrade" method is called

* When  the upgrade is finished, service goes to "upgraded" state instead of going to active state directly. To complete the upgrade, "finishUpgrade" method should be called.

* When service is in "canceled*"/"upgraded states", user gets a chance to perform manual actions on the instances of this service without service.reconcile interfering. 

* Fixed consecutive upgrade: https://github.com/rancher/rancher/issues/2296

* InService upgrade taking long time to finish: https://github.com/rancher/rancher/issues/2293

* Allow launchConfig.labels update for external/Alias service via service.update command: https://github.com/rancher/rancher/issues/2322

@ibuildthecloud any parameters that are the part of launchConfig (image, labels, etc) can be changed using service.upGRADE command for all kinds of services. The parameters that are set on the service itself - scale, externalIps, etc - can be changed using service.upDATE command.


ibuildthecloud:
@alena1108 rollingback should probably be rolling-back.  I'm guessing that is two words and not a made up new contraction.

ibuildthecloud:
Is there an advantage to having two state for canceled-rollback and canceled-upgrade.  What if there was just one `canceled` state.

alena1108:
@ibuildthecloud 

1) regarding introducing common "canceled" state. There are 2 things:

* by having 2 states, it would be clear to see which process got cancelled
* canceled-upgrade is the only valid state for executing the rollback. If we merge canceled-upgrade and canceled-rollback states together, user can call rollback on the service on which he just canceled the rollback (confusing)

2) I'll change the name to rolling-back

ibuildthecloud:
@alena1108 strategy shouldn't be required to be set, default to inplace

ibuildthecloud:
hmmm... this is a breaking API change

ibuildthecloud:
strategies need to be set to nullable in the schema

alena1108:
@ibuildthecloud 

1) set nullable=true on the strategies
2) got rid of duplicated actions (aka finishrollback/finishRollback)

alena1108:
@ibuildthecloud one thing is yet to clarify - Stack state transitions. I've updated it so it replicates service's state transition; but based on our last conversation, you would like to see common end state (aka "unfinished") for stack.cancelupgrade/cancelrollback processes.

code/packaging/app/keys.properties
```diff
@@ -0,0 +1,2 @@
+#Wed Oct 21 00:07:46 UTC 2015
+AES256Key=41a314a9e17483b069855dbbad7b2751e1f937e985092c13e88f3a7b760c46a5
```
ibuildthecloud: Remove this file

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -150,22 +150,28 @@
...
+    <process:process name="environment.cancelupgrade" resourceType="environment" start="upgrading" transitioning="canceling-upgrade" done="canceled-upgrade" />
+    <process:process name="environment.rollback" resourceType="environment" start="upgraded,canceled-upgrade" transitioning="rolling-back" done="active" />
+    <process:process name="environment.cancelrollback" resourceType="environment" start="rolling-back" transitioning="canceling-rollback" done="canceled-upgrade" />
+    <process:process name="environment.finishupgrade" resourceType="environment" start="upgraded,canceled-upgrade" transitioning="finishing-upgrade" done="active" />
+
     
     <!-- Service Discovery Service -->
     <process:process name="service.create" resourceType="service" start="requested" transitioning="registering" done="inactive" />
-    <process:process name="service.activate" resourceType="service" start="inactive" transitioning="activating" done="active" />    
+    <process:process name="service.activate" resourceType="service" start="inactive," transitioning="activating" done="active" />
```
ibuildthecloud: remove the trailing comma

code/iaas/model/src/main/java/io/cattle/platform/core/addon/InServiceUpgradeStrategy.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.List;
+
+@Type(list = false, parent = "serviceUpgradeStrategy")
+public class InServiceUpgradeStrategy extends ServiceUpgradeStrategy {
+    Object launchConfig;
+    List<Object> secondaryLaunchConfigs;
+    Object previousLaunchConfig;
+    List<Object> previousSecondaryLaunchConfigs;
+    boolean isRolling;
```
ibuildthecloud: rename to startFirst

PR #937, opened 2015-10-09
Update to SSL security parameters

alexderz:
For #2286, Added SSL Cipher suite list to Global section and disable SSLv3 and TLSv10 in Frontend bind statement.


ibuildthecloud:
LGTM.  I believe we will merge this after v0.42 is released, so it will go into v0.43

alena1108:
LGTM

sdlarsen:
Looks solid to me as well.

PR #945, opened 2015-10-14
External storage pool and volume events

cjellick:
At this point, storage pool events work, but still need to implement
volume events.


code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -0,0 +1,209 @@
...
+    @Inject
+    VolumeDao volumeDao;
+    @Inject
+    HostDao hostDao;
+    @Inject
+    JsonMapper jsonMapper;
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        objectManager.children(null, null);
```
alena1108: what does this line do?

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -0,0 +1,209 @@
...
+            log.warn("External event doesn't have an external id: {}", event);
+            return null;
+        }
+
+        if (ExternalEventConstants.KIND_VOLUME_EVENT.equals(event.getKind())) {
+            handleVolumeEvent(event, state, process);
+        } else if (ExternalEventConstants.KIND_STORAGE_POOL_EVENT.equals(event.getKind())) {
+            handleStoragePoolEvent(event, state, process);
+        } else {
+            new IllegalStateException();
```
alena1108: Should we set some message for this exception?

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -0,0 +1,209 @@
...
+                if (storagePool == null) {
+                    log.warn("Unknown storage pool. Returning. External id: {}", spExtId);
+                    return;
+                }
+                Volume volume = volumeDao.findVolumeByExternalId(storagePool.getId(), event.getExternalId());
+                switch (event.getEventType()) {
+                case ExternalEventConstants.TYPE_VOLUME_CREATE:
+                    if (volume == null) {
+                        Map<String, Object> volumeData = CollectionUtils.toMap(DataUtils.getFields(event).get("volume"));
+                        if (volumeData == null || volumeData.isEmpty()) {
```
alena1108: I believe "toMap" action never returns null; only an empty map, so the first check can be dropped

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -0,0 +1,209 @@
...
+
+                        volumeDao.createVolumeInStoragePool(volumeData, storagePool);
+                    }
+                    break;
+                case ExternalEventConstants.TYPE_VOLUME_DELETE:
+                    if (volume != null) {
+                        try {
+                            objectProcessManager.scheduleStandardProcess(StandardProcess.DEACTIVATE, volume,
+                                    ProcessUtils.chainInData(new HashMap<String, Object>(), PROC_VOL_DEACTIVATE, PROC_VOL_REMOVE));
+                        } catch (ProcessCancelException e) {
```
alena1108: Seems dangerous to catch processCancelException w/o doing anything. If the volume was in Activating state at the moment, Deactivate might fail to schedule. Should we just let an exception to go through?

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    protected void handleStoragePoolEvent(final ExternalEvent event, ProcessState state, ProcessInstance process) {
+        lockManager.lock(new ExternalEventLock(STORAGE_POOL_LOCK_NAME, event.getAccountId(), event.getExternalId()), new LockCallbackNoReturn() {
+            @Override
+            public void doWithLockNoResult() {
+                StoragePool storagePool = storagePoolDao.findStoragePoolByExternalId(event.getAccountId(), event.getExternalId());
+                if (storagePool == null) {
+                    Map<String, Object> spData = CollectionUtils.toMap(DataUtils.getFields(event).get(FIELD_STORAGE_POOL));
+                    if (spData == null || spData.isEmpty()) {
```
alena1108: Same nit: not null check is not needed

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
cjellick: nothing. wow. how have i not seen that

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
cjellick: I catch and ignore here because these external events run in the background without the user having a way address ones that end up continually rescheduling.

So, the idea is that this makes a "best effort" attempt at scheduling the process for volume or storage pool. We do the same thing with container events.

In a future PR, we'll add sync/ping logic similar to what we have for containers that runs periodically to make sure things sync up.

I do believe it is worth adding an info level log statement though.

Seem reasonable?

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void createVolumeInStoragePool(Map<String, Object> volumeData, StoragePool storagePool) {
+        try {
+            Volume volume = resourceDao.createAndSchedule(Volume.class, volumeData);
+            Map<String, Object> vspm = new HashMap<String, Object>();
+            vspm.put("volumeId", volume.getId());
+            vspm.put("storagePoolId", storagePool.getId());
+            resourceDao.createAndSchedule(VolumeStoragePoolMap.class, vspm);
+        } catch (ProcessCancelException e) {
```
cjellick: Im going to move this catch out into the handler and log it.

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
alena1108: sure, lets log it for now

cjellick:
Test failure is legitimate and related to my code. Investigating

alena1108:
LGTM

PR #946, opened 2015-10-14
Bug fixes

alena1108:
1) Validate against Removed/Removing states for environment object, when add a service to it

https://github.com/rancher/rancher/issues/2334
https://github.com/rancher/rancher/issues/1644

2) integration test fix

### DIFF
--- a/tests/integration/cattletest/core/test_healthcheck.py
+++ b/tests/integration/cattletest/core/test_healthcheck.py
@@ -149,6 +149,10 @@ def test_health_check_bad_external_timestamp(super_client, context, client):
 def test_health_check_bad_agent(super_client, context, client):
     # Create another host to get the agent from that host
     host2 = super_client.reload(register_simulated_host(context))
+    # register one more host to ensure
+    # there is at least one more host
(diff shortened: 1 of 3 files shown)


tests/integration/cattletest/core/test_healthcheck.py
```diff
@@ -149,6 +149,10 @@ def test_health_check_bad_external_timestamp(super_client, context, client):
 def test_health_check_bad_agent(super_client, context, client):
     # Create another host to get the agent from that host
     host2 = super_client.reload(register_simulated_host(context))
+    # register one more host to ensure
+    # there is at least one more host
+    # to schedule healtcheck on
+    super_client.reload(register_simulated_host(context))
```
cjellick: Why do you reload without assigning back to context?

alena1108: @cjellick you are right, there is no need to reload here. The purpose of this line is only to register simulated host; will apply the fix

cjellick:
LGTM

PR #951, opened 2015-10-15
Add stack upgrade API

cjellick:


cjellick:
API looks fine to me, but @vincent99 is the one that has to use it :-)

code/iaas/logic/src/main/resources/META-INF/cattle/core-process/spring-core-processes-context.xml
```diff
@@ -151,8 +151,12 @@
     <!-- Service Discovery Environment -->
     <process:process name="environment.create" resourceType="environment" start="requested" transitioning="activating" done="active" />
     <process:process name="environment.update" resourceType="environment" start="active" transitioning="updating-active" done="active" />
-    <process:process name="environment.remove" resourceType="environment" start="requested, active, activating, updating-active, error, erroring" transitioning="removing" done="removed" />
+    <process:process name="environment.remove" resourceType="environment" start="requested, active, activating, updating-active, error, erroring, upgrading, canceling-upgrade, canceling-rollback, rolling-back" transitioning="removing" done="removed" />
     <process:process name="environment.error" resourceType="environment" start="activating" transitioning="erroring" done="error" />
+    <process:process name="environment.upgrade" resourceType="environment" start="active" transitioning="upgrading" done="active" />
```
alena1108: I have very similar state transitions for service.upgrade, with only one difference: canceling-upgrade and canceling-rollback have "canceledupgrade" and "canceledrollback" end states:

https://github.com/rancher/cattle/pull/935

 The reason is - "active" state doesn't really indicate that the cancel-upgrade or cancel-rollback has failed. Also canceledupgrade and canceledrollback states let user look at the service and fix things; and service becomes "frozen" while in this state. 

Do you think it would make sense to do something similar for the environment? So when its in "canceled*" state, no new services can be added to it?

alena1108:
Will be merged as a part of https://github.com/rancher/cattle/pull/959 (needed to resolve merge conflicts)

PR #956, opened 2015-10-16
Replace single $ with $$ in exported yml files

alena1108:
https://github.com/rancher/rancher/issues/2351

@ibuildthecloud just to confirm: we have to replace in both docker-compose.yml and rancher-compose.yml? 

Also is there a chance that we override user defined value where $ is expected to be single? Like in "command" field for example


ibuildthecloud:
This applies to both files.  I can't think of anyway that we will do the wrong thing by turning $ into $$.

alena1108:
@ibuildthecloud ok, then the PR is ready for a review. TF happened in external volume event test @cjellick

PR #969, opened 2015-10-21
Make All auth types more similar.

hibooboo2:
Fix various bugs with authentication and projects.


alena1108:
@hibooboo2 could you put some brief information about the bugs that PR fixes? Or make the ref to the github tickets.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/PasswordDaoImpl.java
```diff
@@ -51,46 +54,65 @@ public Credential changePassword(Credential password, ChangePassword changePassw
...
+                .and(CREDENTIAL.PUBLIC_VALUE.eq(username)
+                .and(CREDENTIAL.KIND.equalIgnoreCase(CredentialConstants.KIND_PASSWORD)))
+                .fetchOne();
+
+        SelectConditionStep<CredentialRecord> query =
+                create().selectFrom(CREDENTIAL)
+                        .where(CREDENTIAL.PUBLIC_VALUE.eq(username)
+                                .and(CREDENTIAL.STATE.eq(CommonStatesConstants.ACTIVE)
+                                        .and(CREDENTIAL.KIND.eq(CredentialConstants.KIND_PASSWORD))));
+        if (credential != null ) {
```
alena1108: @hibooboo2 why do we make 2 similar queries on line 62 and line 78? If its just to figure out if the there are duplicated entries for the same account,  can we use count() combined with groupBy() clause here?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/ApiAuthenticator.java
```diff
@@ -55,11 +55,15 @@ public void handle(ApiRequest request) throws IOException {
...
         Account authenticatedAsAccount = getAccount(request);
-        if (authenticatedAsAccount == null) {
-            throwUnauthorizated();
+        if (authenticatedAsAccount == null ||
+                !StringUtils.equals(CommonStatesConstants.ACTIVE, authenticatedAsAccount.getState())) {
+            throw new ClientVisibleException(ResponseCodes.UNAUTHORIZED);
         }
 
         Set<Identity> identities = getIdentities(authenticatedAsAccount);
+        if (identities == null || identities.size() == 0) {
```
alena1108: Nit: looks like getIdentities can never return null, only an empty list, so not null check can be omitted here.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/ApiAuthenticator.java
```diff
@@ -150,7 +154,7 @@ protected Account getAccount(ApiRequest request) {
         for (IdentityTransformationHandler identityTransformationHandler : identityTransformationHandlers) {
             identities.addAll(identityTransformationHandler.getIdentities(account));
         }
-
+        identities.remove(null);
```
alena1108: This line looks a bit off; if we want to restrict the set from having not null elements, we can put a not null check at the place where element gets added to the set.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/PasswordDaoImpl.java
hibooboo2: Hmm you are right when i wrote this i was combining some code in a new way. I will have to rework this logic tomorrow. Not sure how I will get what this was supposed to accomplish as this will not accomplish what i was after.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/ApiAuthenticator.java
hibooboo2: I could but that would require like 7 null checks in different places. Sometimes it is an error for null identities and some times it is not. But in this case it is not, however we do not want nulls in the set.

hibooboo2:
So the bugs that i fixed in that pr are not bugs that have issues filed. They are bugs that I noticed as I was refactoring satisfy the idea of code reuse for different auth types. As before a large part of the code was the same but slightly different and executed in 3 places. This refactor makes it so that if it works with local Auth, github auth , or ldap auth it will work with all 3. Which means that the only thing specifically tested for each is the ability to create a token(communicate with the respective systems.) Where as before the tokens were made differently and the way it worked was not shared nearly as much.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/SecurityConstants.java
```diff
@@ -21,4 +21,5 @@
     public static final String TOKEN_VERSION = "v1";
     public static final DynamicLongProperty TOKEN_EXPIRY_MILLIS = ArchaiusUtil.getLong("api.auth.jwt.token.expiry");
     public static final String HAS_LOGGED_IN = "hasLoggedIn";
+    public static final String AUTH_ENABLER = "api.auth.enabler";
```
hibooboo2: Used to track who turned on authentication.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/GithubConfigManager.java
```diff
@@ -92,6 +100,19 @@ public GithubConfig getCurrentConfig(Map<String, Object> config) {
         if (config.get(CLIENT_ID) != null) {
             clientId = (String) config.get(CLIENT_ID);
         }
+        if (((Policy) ApiContext.getContext().getPolicy()).isOption(Policy.AUTHORIZED_FOR_ALL_ACCOUNTS)) {
+            if (StringUtils.isBlank(
+                    (String) ApiContext.getContext().getApiRequest().getAttribute(GithubConstants.GITHUB_ACCESS_TOKEN))) {
+                Identity gotIdentity = Identity.fromId(ArchaiusUtil.getString(SecurityConstants.AUTH_ENABLER).get());
```
hibooboo2: The reason that this as been added is so that if there is not an Apikey avaliable for github for a user who is an admin and is listing github config we are able to access github using the person who turned github auth on. That way there are no 500's due to NPE. We could also just not list the saved identities, however this does not seem like a good options/ the right one.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/GithubTokenCreator.java
```diff
@@ -35,59 +19,15 @@
...
-        objectManager.persist(account);
-        account = objectManager.reload(account);
-        String accountId = (String) ApiContext.getContext().getIdFormatter().formatId(objectManager.getType(Account.class), account.getId());
-        Date expiry = new Date(System.currentTimeMillis() + SecurityConstants.TOKEN_EXPIRY_MILLIS.get());
-        String jwt = tokenService.generateEncryptedToken(jsonData, expiry);
-        //LEGACY: Used for old Implementation of projects/ Identities. Remove when vincent changes to new api.
-        return new Token(jwt, user.getName(), null, teamsAccountInfo, SecurityConstants.SECURITY.get(),
-                GithubConstants.GITHUB_CLIENT_ID.get(), account.getKind(), SecurityConstants.AUTH_PROVIDER.get(),
-                accountId, new ArrayList<>(identities), user);
+        return githubUtils.createToken(githubClient.getIdentities(accessToken), null);
```
hibooboo2: This logic was half repeated from other classes as a result it was moved to tokenUtils and getIdentities in the githubClient.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/GithubUtils.java
```diff
@@ -20,6 +24,25 @@ protected String accessToken() {
     }
 
     @Override
+    protected void postAuthModification(Account account) {
+        ApiRequest request = ApiContext.getContext().getApiRequest();
```
hibooboo2: This is to ensure that after an account is authenticated with github and rancher the api key for github that we need on subsequent requests is saved to the database. This used to happen in the token creator however it is more appropriate here.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/resource/GithubClient.java
```diff
@@ -201,7 +207,10 @@ public GithubAccountInfo getGithubUserByName(String username) {
     }
 
     public GithubAccountInfo getGithubOrgByName(String org) {
-        String gitHubAccessToken = (String) ApiContext.getContext().getApiRequest().getAttribute(GithubConstants.GITHUB_ACCESS_TOKEN);
+        String githubAccessToken = (String) ApiContext.getContext().getApiRequest().getAttribute(GithubConstants.GITHUB_ACCESS_TOKEN);
```
hibooboo2: Better null checks for accessToken so that when it is not present the correct error is presented to the user.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/github/resource/GithubClient.java
```diff
@@ -318,4 +330,27 @@ public GithubAccountInfo getUserOrgById(String id) {
     public String getName() {
         return GithubConstants.GITHUB_CLIENT;
     }
+
+    public Set<Identity> getIdentities(String accessToken) {
```
hibooboo2: This was added so that there is a clearer pattern seen in the token creator classes.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
```diff
@@ -254,29 +257,42 @@ public String getAccessToken() {
 
     protected abstract String accessToken();
 
-    public Account getOrCreateAccount(Identity gotIdentity, Set<Identity> identities, Account account, boolean createAccount) {
-        boolean hasAccessToAProject = authDao.hasAccessToAnyProject(identities, false, null);
+    public Account getOrCreateAccount(Identity user, Set<Identity> identities, Account account) {
```
hibooboo2: The logic in this method was reworked so that the behavior across providers is the same. Meaning that account creation for a provider is specified with in the Implementation of token utils as dictated by createAccount(). As well as when the authentication is being turned on we record who turns it on, as well as instead of recycling the same admin account we create a new one for the user or use their existing one, unless the authentication mode they are using does not allow for creating an account in which case we throw an error. This case happens when local auth is being turned on and someone tries to create a token using a username and password that has not been associated with a pre existing account.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
```diff
@@ -284,4 +300,55 @@ public Account getOrCreateAccount(Identity gotIdentity, Set<Identity> identities
         }
         return account;
     }
+
+
+    public Token createToken(Set<Identity> identities, Account account) {
```
hibooboo2: This logic was moved from 3 different locations where the token was created in almost the exact same manner. Now the token is created in the same manner, while allowing any modification that is needed to the associated account to be done before the token is returned but only after the user has been authenticated.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -200,7 +200,7 @@ public Account createAccount(String name, String kind, String externalId, String
     }
 
     @Override
-    public Identity getIdentity(Long id) {
+    public Identity getIdentity(Long id, boolean obfuscate) {
```
hibooboo2: This method was modified with this flag so that when called we know if the returned value will be the one for server use or client.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
```diff
@@ -422,7 +424,7 @@ public boolean isProjectMember(long projectId, Long usingAccount, boolean isAdmi
                 List<? extends ProjectMember> toDelete = create().selectFrom(PROJECT_MEMBER).where(allMembers).fetch();
                 for (ProjectMember member : toDelete) {
                     objectProcessManager.executeStandardProcess(StandardProcess.DEACTIVATE, member, null);
-                    objectProcessManager.executeStandardProcess(StandardProcess.REMOVE, member, null);
+                    objectProcessManager.scheduleStandardProcess(StandardProcess.REMOVE, member, null);
```
hibooboo2: This fixes errors that were occurring when idempotent checks were on.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/Token.java
```diff
@@ -1,8 +1,9 @@
 package io.cattle.platform.iaas.api.auth.identity;
 
 import io.cattle.platform.api.auth.Identity;
+import io.cattle.platform.iaas.api.auth.SecurityConstants;
```
hibooboo2: The changes in this class were to remove deprecated token creation methods that are no longer needed.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -57,15 +61,21 @@ private Token createToken(ApiRequest request) {
             throw new ClientVisibleException(ResponseCodes.BAD_REQUEST,
                     "codeInvalid", "Code provided is invalid.", null);
         }
+
+        Identity[] identities = token.getIdentities();
```
hibooboo2: This is to ensure that any identities that are returned in the token are ones that are meant for the client and not the server.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/DefaultAuthorizationProvider.java
```diff
@@ -58,27 +58,31 @@ public SchemaFactory getSchemaFactory(Account account, Policy policy, ApiRequest
                 return schemaFactory;
             }
         }
-        List<? extends ProjectMember> projectMembers = null;
-        if (account != null && account.getKind().equalsIgnoreCase(ProjectConstants.TYPE)){
+        List<? extends ProjectMember> projectMembers;
```
hibooboo2: This ensures that the proper role is actually returned.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/LdapTokenCreator.java
```diff
@@ -68,36 +59,6 @@ public String providerType() {
         return LdapConstants.CONFIG;
     }
 
-    public Token getTokenByIdentities(Set<Identity> identities){
```
hibooboo2: Moved to tokenutils.createToken.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/TokenUtils.java
```diff
@@ -284,4 +300,55 @@ public Account getOrCreateAccount(Identity gotIdentity, Set<Identity> identities
...
+        postAuthModification(account);
+
+        authDao.updateAccount(account, account.getName(), account.getKind(), user.getExternalId(), user
+                .getExternalIdType());
+
+        Map<String, Object> jsonData = new HashMap<>();
+        jsonData.put(TokenUtils.TOKEN, tokenType());
+        jsonData.put(TokenUtils.ACCOUNT_ID, user.getExternalId());
+        jsonData.put(TokenUtils.ID_LIST, identitiesToIdList(identities));
+        account = objectManager.reload(account);
```
alena1108: I think it would make sense to make authDao.updateAccount() to return an updated account. Would save you and extra DB call that you do by objectManager.reload(account)

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthDaoImpl.java
alena1108: I think this needs to be done in a slightly different fashion. When obfuscate is passed in, the code implies that the ApiContext might not be null at this point (line 213). Just pass in IdFormatter to the call; and if its null - return the id non-obfuscated.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/PasswordDaoImpl.java
```diff
@@ -51,46 +54,65 @@ public Credential changePassword(Credential password, ChangePassword changePassw
...
-                .execute();
+        if (query.fetchAny() != null){
+            throw new ClientVisibleException(ResponseCodes.CONFLICT, "UsernameIsTaken", "Username: " + username +" is taken", null);
+        }
 
         Map<Object, Object> properties = new HashMap<>();
 
-        if (StringUtils.isBlank(username) || StringUtils.isBlank(password)) {
-            throw new ClientVisibleException(ResponseCodes.BAD_REQUEST, "UserCredentialCreation", "Cannot have blank username or password.", null);
+        if (credential != null) {
```
alena1108: What are we doing here? :) cleaning up the credentials for the existing account?

hibooboo2: Basically. Was originally a different query. Need to rework this entire method, been thinking about how to do this all day. I think I almost have it.

alena1108:
@hibooboo2 failed test is concerning

alena1108:
LGTM, but only if the test is green ;)

PR #972, opened 2015-10-22
Register instance to lb on exposemap.create

alena1108:
used to be enabled only on the instance.start event for service instance. For regular, non selector based containers, exposeMap exists at this point, so lb target gets created. For selector based container, instance.start schedules serviceExposeMap.creation, so the map might not exist when this code is called.


alena1108:
@ibuildthecloud

PR #975, opened 2015-10-22
Added lock to register service instance to Lb

alena1108:
https://github.com/rancher/rancher/issues/2399

@ibuildthecloud I want it to be the last fix for standalone LB code


ibuildthecloud:
@alena1108 do you want this to be a blocking lock?  The lock you have right now will immediately fail, basically tryLock.

alena1108:
@ibuildthecloud makes sense

PR #984, opened 2015-10-23
Bug fixes

alena1108:
1) Cleanup instance from planner on its removal

When instance with missing dependencies gets removed up by deployment planner, we should remove it from the planner's list of instances. Otherwise it would participate in further deployment, and further down will hit the timeout waiting for this instance to hit valid state (Running/Stopped, etc)

2) Don't schedule service config.update on service in updating-active/activating state. Reconcile process invoked by update/activate operation, can result in some instances being stopped/destroyed. And stop/destroy processes have config.upgrade hooks resulting in subsequent reconcile requests. 

At the end of every activate/update process, we double check that the reconcile have been finished successfully, and if not - the process gets rescheduled. 

This 2 fixes above are supposed to fix the bugs related to service reconcile slowness, like: 

https://github.com/rancher/rancher/issues/2044

3) Validate service selector on instance.restore

https://github.com/rancher/rancher/issues/2338


ibuildthecloud:
I feel like you need better PR titles.  More uplifting like "Vast improvement in functional correctness"

code/packaging/app/keys.properties
```diff
@@ -0,0 +1,2 @@
+#Fri Oct 23 19:03:46 UTC 2015
+AES256Key=910c3458c5b2ef9eff389efba8421e4153ab755d6379d637d3aae549f3893c1f
```
ibuildthecloud: NO!!!!!! not the key again....  pesky file

alena1108: Argh, my bad. Fix is coming

alena1108:
@ibuildthecloud :) will improve the naming in the future PRs

PR #993, opened 2015-10-26
Broke down big integration test into a set of smaller ones

alena1108:
@ibuildthecloud


ibuildthecloud:
I don't know... now the build is 22 minutes :)  So was one big one faster?

alena1108:
@ibuildthecloud weird, as now being split, it should be available for parallel execution (it thats enabled on drone). Number of calls hasn't been changed. I'm going to make one more change - reduce the interval to 500ms

alena1108:
@ibuildthecloud 100ms is the minLimit in cattle for interval. So I'm using this value across all my tests

PR #996, opened 2015-10-27
externalDns flag (boolean, default=true) to service

alena1108:
If set, and external dns provider is registered to cattle, service will
get registered to public DNS

Introduced this flag so user has a choice to "turn off" dns for certain
services.

@ibuildthecloud

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/ServiceMetaData.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/ServiceMetaData.java
@@ -33,6 +33,7 @@ public class ServiceMetaData {
     Map<String, String> labels;
     Map<String, Object> metadata;
     Integer scale;
+    Boolean external_dns = true;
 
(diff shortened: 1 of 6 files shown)


code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/ServiceMetaData.java
```diff
@@ -52,6 +53,8 @@ public ServiceMetaData(Service service, String serviceName, Environment env, Lis
         this.create_index = service.getCreateIndex();
         this.metadata = metadata;
         this.scale = DataAccessor.fieldInteger(service, ServiceDiscoveryConstants.FIELD_SCALE);
+        this.external_dns = DataAccessor.fields(service)
```
ibuildthecloud: do you want DataAccessor.fieldBool or fieldBoolean?

alena1108:
@ibuildthecloud sure, will convert the flag to label too

alena1108:
I'm going to convert this flag to label, so closing the PR.

PR #999, opened 2015-10-27
Support kubernetes

cjellick:
Handle the creation and updating of services created outside of Rancher


resources/content/db/changelog.xml
```diff
@@ -72,5 +72,6 @@
     <include file="db/core-065.xml"/>
     <include file="db/core-066.xml"/>
     <include file="db/core-067.xml"/>
+    <include file="db/core-101.xml"/>
```
cjellick: Self reminder: need to fix this before merging

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/AgentInstanceDao.java
```diff
@@ -30,4 +30,5 @@
 
     void populateNicAndIp(NetworkServiceInfo service);
 
+    List<Long> getAgentProvider(String providedServiceLabel, Long accountId);
```
alena1108: Nit: if accountId is never meant to be null, can we change Long to long in the signature?

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventManagerImpl.java
```diff
@@ -0,0 +1,199 @@
...
+        if (serviceData != null) {
+            service.putAll(serviceData);
+        }
+
+        try {
+            String create = objectProcessManager.getStandardProcessName(StandardProcess.CREATE, Service.class);
+            String activate = objectProcessManager.getStandardProcessName(StandardProcess.ACTIVATE, Service.class);
+            ProcessUtils.chainInData(service, create, activate);
+            resourceDao.createAndSchedule(Service.class, service);
+        } catch (ProcessCancelException e) {
```
alena1108: What is the impact of processCancelException (event not being processed)? Do you think we should log it?

alena1108:
LGTM

PR #1004, opened 2015-10-30
Killed standalone LB

alena1108:


alena1108:
Issues that will get fixed along:
https://github.com/rancher/rancher/issues/2047

alena1108:
@ibuildthecloud could you please review. Don't get scared by file change number, its mostly code removal.

ibuildthecloud:
rebase?

ibuildthecloud:
OMG  +450 −8,467.  Best PR ever!!!!!!

alena1108:
@ibuildthecloud I know, even my OCD inner self got quiet after the 100th removed file.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/factory/impl/AgentInstanceBuilderImpl.java
```diff
@@ -50,7 +52,11 @@ public AgentInstanceBuilderImpl(AgentInstanceFactoryImpl factory, Instance insta
         this(factory);
         this.accountId = instance.getAccountId();
         this.zoneId = instance.getZoneId();
-        this.uri = "event:///instanceId=" + instance.getId();
+        String uriPrefix = "event";
+        if (StringUtils.equals(SystemContainer.LoadBalancerAgent.name(), instance.getSystemContainer())) {
```
ibuildthecloud: This is a hack.  Can you just pass in the uri or uri prefix somehow?

ibuildthecloud:
@alena1108 We can't actually drop the tables in this release.  Then rollback isn't possible.  We can either just leave it as is for this release (drop the tables in a future release), or modify code/iaas/model/src/test/resources/codegen.xml to exclude the tables we don't use anymore.  I prefer the later.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/factory/impl/AgentInstanceBuilderImpl.java
alena1108: @ibuildthecloud I can use system label to pass this info if you are ok with this approach

alena1108:
@ibuildthecloud disabled *load_balancer* tables in codegen.xml

PR #1006, opened 2015-10-30
Fix external event test

cjellick:
Fixes random failures in this test caused by a race condition in
attempting to load the event's original account id.


alena1108:
LGTM

PR #1018, opened 2015-11-06
Fix npe caused by null string comparison.

hibooboo2:


alena1108:
LGTM

ibuildthecloud:
In general do CONSTANT.equals(foo). Then you can't get a NPE.  it's better than using the util method.

cjellick:
LGTM

PR #1020, opened 2015-11-06
Added option http-server-close for websockets.

rmcleay:
Without adding "option http-server-close", HAProxy seems to not work with websockets.

Unfortunately, this setting does ignore the keepalive setting of
containers that the load balancer is linked to.

For more information, see http://blog.silverbucket.net/post/31927044856/3-ways-to-configure-haproxy-for-websockets and http://blog.haproxy.com/2012/11/07/websockets-load-balancing-with-haproxy/

### DIFF
--- a/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
+++ b/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
@@ -19,6 +19,7 @@ defaults
 	option	tcplog
         option  dontlognull
         option  redispatch
+        option http-server-close
         option forwardfor
         retries 3
         timeout connect 5000


ibuildthecloud:
This looks good, but this does point out that we are missing `timeout tunnel`

ibuildthecloud:
@rmcleay Hmmm... you have the website fearthecow.net and you are contributing to a project named cattle.  Was this just one of a multi-step process in overcoming your bovinophobia :smile:

ibuildthecloud:
ping @alena1108

alena1108:
@ibuildthecloud @rmcleayI would suggest we hold on with this change. I'm going to work on https://github.com/rancher/rancher/issues/1871 ("LB (haproxy): support for custom parameters in default section") early next week. It will enable user to define all haproxy default settings his application requires, on per service basis.

ibuildthecloud:
@alena1108  I don't think there is any particular downside of having this in the defaults by default.

alena1108:
@ibuildthecloud no downside; its just keep-alive is the default http mode on haproxy, so I wanted to stay aligned in our defaults.

ibuildthecloud:
@alena1108 I don't quite get what your saying.  This change still does client keep-alive.  Just the haproxy -> server connection is closed.  This should have minimal impact because 1) we typically do SSL termination 2) the load balancer sits very close to the container.

I really think websockets should work out of the box.  I don't want users to have to do something to turn it on.  If this option is needed, I'd like it to be the default.

rmcleay:
@alena1108 

I tend to agree with @ibuildthecloud. This should have minimal impact on performance as option http-server-close closes the connection on the internal network side of the load balancer, but keeps the connection alive to the customer.

Nice picture by the way!

alena1108:
@ibuildthecloud @rmcleay ok ok :) I'm convinced. LGTM

alena1108:
@ibuildthecloud 
>>  I don't quite get what your saying.

Elaborating. haproxy supports following http modes:
http-tunnel
http-server-close
forceclose
httpclose
http-keep-alive - this is the default http mode

Considering the use cases we want to support out of the box, having httpclose set by Rancher as default, is legit.

PR #1026, opened 2015-11-07
Add auditing service to cattle.

hibooboo2:
rancher/rancher#1267


alena1108:
@hibooboo2 could you briefly explain what this service is doing? Or provider a ref to the github ticket if there is one.

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditLogsRequestHandler.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final Logger log = LoggerFactory.getLogger(AuditLogsRequestHandler.class);
+    @Inject
+    AuditService auditService;
+
+    @Override
+    public void handle(ApiRequest request) throws IOException {
+        if (!Schema.Method.GET.isMethod(request.getMethod())){
+            request.setAttribute("requestEndTime", System.currentTimeMillis());
+            Policy policy = (Policy) ApiContext.getContext().getPolicy();
+            log.info("Logged ApiRequest from: " + policy.getUserName());
```
alena1108: Is policy.userName unique in the same Rancher instance?

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+                break;
+            }
+        }
+        if (user == null && policy.getIdentities().size() == 1){
+            user = policy.getIdentities().iterator().next();
+        }
+        long runtime = ((long) request.getAttribute("requestEndTime")) - ((long)request.getAttribute("requestStartTime"));
+        String authType = (String) request.getAttribute(AccountConstants.AUTH_TYPE);
+        String resourceId =request.getResponseObject() instanceof Resource ? ((Resource) request.getResponseObject()).getId(): "1a0";
+        Long parsedResourceId = resourceId == null ? (long) 0 : Long.valueOf(ApiContext.getContext().getIdFormatter().parseId(resourceId));
```
alena1108: Nit: (long)0 can be written as 0L

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/ApiAuthenticator.java
```diff
@@ -145,7 +147,7 @@ protected Account getAccount(ApiRequest request) {
             }
         }
 
-        return account;
+        return null;
```
alena1108: why we switched from returning account to returning null?

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+    public void logResourceModification(Resource resource, Map<String, Object> data, String eventType, String description) {
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        try {
+            jsonMapper.writeValue(os, resource);
+            data.put("resource", os.toString());
+        } catch (IOException e) {
+            log.error("Failed to write response object to log.");
+            data.put("resource", "{\"object\":\"Failed to convert response object to json.\"}");
+        }
+        auditLogDao.create(resource.getType(), resource.getId() != null ? Long.valueOf(resource.getId()): (long) 0,
```
alena1108: To clarify: resourceId would be null for Create calls? Do you think it would be cleaner if you allow to pass null values to the auditLogDao.create, and then convert null to 0L there? So you don't have to do it in all the methods calling  auditLogDao.create

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+        for (Identity identity: policy.getIdentities()){
+            if (identity.getExternalIdType().contains("user")){
+                user = identity;
+                break;
+            }
+        }
+        if (user == null && policy.getIdentities().size() == 1){
+            user = policy.getIdentities().iterator().next();
+        }
+        long runtime = ((long) request.getAttribute("requestEndTime")) - ((long)request.getAttribute("requestStartTime"));
```
alena1108: Would requestEndTime always set on the request? I just see that in this PR you set it for only non GET methods, and wonder if there are chances for NPE when cast to (long):

if (!Schema.Method.GET.isMethod(request.getMethod())){
+            request.setAttribute("requestEndTime", System.currentTimeMillis());

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditLogsRequestHandler.java
hibooboo2: Usernames are unique in a rancher instance. User1 is always user1.(Assuming auth types are never changed.)

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
hibooboo2: You are right I only set it on non GET Methods. But it is also only logged on non GET. So  it will always be set.

hibooboo2: But i will add a check for the request method being get. And not log if it is.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/impl/ApiAuthenticator.java
hibooboo2: The previous logic always returned account here which was always null. This is more clear that that is the case.

alena1108: ok

alena1108:
we need tests as well :)

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditService.java
```diff
@@ -0,0 +1,20 @@
...
+    void logRequest(ApiRequest request, Policy policy);
+
+    /**
+     * @param data {@link Map} -  The {@link Object}'s that will be set as the fields on the created
+     *             {@link io.cattle.platform.core.model.AuditLog} using the {@link io.cattle.platform.core.model.AuditLog}.setData.
+     *             Where each Key {@link String} will
+     *             be a the name of the field and the {@link Object} will be the Value.
+     * @param resource @see Resource
+     */
+    void logResourceModification(Resource resource, Map<String, Object> data, String eventType, String description);
```
ibuildthecloud:
```
void logApiRequest(ApiRequest request, Policy policy)
void logAuditEvent(Object resource, Map<String, Object> data, AuditEventType eventType, String description. Long accountId)
```

code/framework/auditing/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    <artifactId>cattle-framework-auditing</artifactId>
+    <parent>
+        <artifactId>cattle-parent</artifactId>
+        <groupId>io.cattle</groupId>
+        <version>0.5.0-SNAPSHOT</version>
+        <relativePath>../../parent/pom.xml</relativePath>
+    </parent>
+    
+    <dependencies>
+        <dependency>
```
ibuildthecloud: only jooq and cattle-iaas-api-logic is actually required here.  Plus add
```xml
        <dependency>
            <groupId>javax.servlet</groupId>
            <artifactId>servlet-api</artifactId>
            <scope>provided</scope>
        </dependency>```

File: code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditLogsRequestHandler.java (hunk 1 above)
Comment (ibuildthecloud) [inline code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditLogsRequestHandler.java]: Don't log this, it just spams the logs

**Comment by ibuildthecloud** [comment]
@hibooboo2 You need to blacklist the type `publish` otherwise you write a million records to the DB for each message published by an agent.

**Comment by ibuildthecloud** [comment]
@hibooboo2 When are the audit logs purged, if ever?

File: code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditLogOutPutFilter.java [hunk 8]
```diff
@@ -0,0 +1,43 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+
+public class AuditLogOutPutFilter implements ResourceOutputFilter {
+    @Override
+    public Resource filter(ApiRequest request, Object original, Resource converted) {
+        if (original instanceof AuditLogRecord) {
+            AuditLogRecord auditLogRecord = (AuditLogRecord) original;
+            if (auditLogRecord.getResourceId() != null) {
+                String id = (String) ApiContext.getContext()
```
ibuildthecloud: id is not used

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(AuditLogsRequestHandler.class);
+
+    @Override
+    public void logRequest(ApiRequest request, Policy policy) {
+        if (Schema.Method.GET.isMethod(request.getMethod())) {
+            return;
+        }
+            Map<String, Object> data = new HashMap<>();
+        Map requestObject = CollectionUtils.toMap(request.getRequestObject());
```
ibuildthecloud: Missing generic type `Map<?, ?>`.  This causes a warning.

code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanUp.java
```diff
@@ -50,16 +52,24 @@ public void cleanUp(){
                 .where(CONTAINER_EVENT.CREATED.lt(expiredEvents)
                 .and(CONTAINER_EVENT.STATE.eq(CommonStatesConstants.CREATED)))
                 .execute();
-
         long endContainerEvent = System.currentTimeMillis();
+
+        long startAuditLog = System.currentTimeMillis();
+        int deletedAuditLogRecords = create()
+                .delete(AUDIT_LOG)
+                .where(AUDIT_LOG.CREATED.lt(new Date(System.currentTimeMillis() - (AUDIT_LOG_TIME.get() * 1000 * 86400))))
```
ibuildthecloud: What is 86400 for?

ibuildthecloud:
@hibooboo2 You need to blacklist type `configContent` also

ibuildthecloud:
And `containerEvent`

ibuildthecloud:
And `serviceEvent`

ibuildthecloud:
And `externalEvent`

code/framework/auditing/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+            <groupId>io.cattle</groupId>
+            <artifactId>cattle-iaas-model</artifactId>
```
ibuildthecloud: model isn't needed, it's a transitive dependency

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
```diff
@@ -0,0 +1,160 @@
...
+            jsonMapper.writeValue(os, objectToPlace);
+            data.put(fieldForObject, os.toString());
+        } catch (IOException e) {
+            log.error(errMsg);
+            data.put("requestObject", "{\"object\":\"" + errMsg + "\"");
+        }
+    }
+
+    private boolean containsIgnoreCase(Set<String> list, String string){
+        for(String i : list){
```
ibuildthecloud: just lower case everything.  so have the set be all lowercase and then lower case the value you are checking.  then its `O(1)`

hibooboo2: How would it be 0(1) when the method that is used to do contains is:

```
    final Entry<K,V> getEntry(Object key) {
        if (size == 0) {
            return null;
        }

        int hash = (key == null) ? 0 : hash(key);
        for (Entry<K,V> e = table[indexFor(hash, table.length)];
             e != null;
             e = e.next) {
            Object k;
            if (e.hash == hash &&
                ((k = e.key) == key || (key != null && key.equals(k))))
                return e;
        }
        return null;
    }
```

PR #1034, opened 2015-11-10
Disabled launchConfig.restartPolicy in the API

alena1108:
@ibuildthecloud thought it would be better to just disable restartPolicy in the API for service.laucnhConfig instead of just ignoring it when build a container + export a config. This way its more clear to the end user

https://github.com/rancher/cattle/pull/1021
https://github.com/rancher/rancher/issues/2585


ibuildthecloud:
@alena1108 Do you need to revert any of your previous PR?

alena1108:
@ibuildthecloud decided to not revert it as it had tests. So simply removed the class responsible for formatting rancher restart policy to corresponding compose field as there is no need for this class any more - restart policy will never be set on the exported service.

PR #1035, opened 2015-11-10
Don't pick host with agent in Reconnecting state

alena1108:
when schedule instance start by service with global flag = true.
Also don't schedule healtcheck on the network instance running on the
host that has agent in reconnecting state

https://github.com/rancher/rancher/issues/2622
https://github.com/rancher/rancher/issues/2516


alena1108:
@ibuildthecloud

PR #1039, opened 2015-11-11
Services inside stacks are discoverable to each other by default

alena1108:
* service is discoverable to every other service within the same stack by default. Service links would only be useful if user wants to override DNS name
* every container is discoverable by its name to every other container in the same stack
*  to make service discoverable to outside of the stack, links should still be used.
* for service to register to the LB, links still be used.
https://github.com/rancher/rancher/issues/2628


alena1108:
@ibuildthecloud this PR is ready to be merged, but only after https://github.com/rancher/cattle/pull/1048 is merged.

PR #1043, opened 2015-11-11
Reschedule healtchecks when host goes to reconnecting state

alena1108:
When agent goes through Reconnecting process:

1) as a pre-handler, deregister all the healtchecks scheduled on the host
2) as a post handler, register all the instances with health check enabled, for health check again. This particular scenario is to cover the case when its the only host in the system going through Reconnecting process (as health check gets de-registered on step 1)
3) Cleanup health check info on the instance removal


alena1108:
https://github.com/rancher/rancher/issues/2622

PR #1044, opened 2015-11-12
Trigger healtcheck update on healthcheckhostmap.create

alena1108:
https://github.com/rancher/rancher/issues/2641

@ibuildthecloud:

* added that missing NicLookup for healtcheckhostmap.create.

* removed obsolete code registering instance for healtcheck in HostRemovePreHandler. Instance is being registered automatically as a part of HealthcheckInstanceHostMapRemovePostHandler

* when agent reconnects back to cattle, schedule for healtchecks all instances of host's account (in case healtcheck is missing for some of them, so it can be set on this host). That would fix the scenario when there are only 2 hosts in the system, and host2 is going down while host1 is coming up taking over all host2's healtchecks https://github.com/rancher/rancher/issues/2642


ibuildthecloud:
Assuming green LGTM.

PR #1047, opened 2015-11-12
Revert "Fix issue in which not all port ports get exposed"

ibuildthecloud:
Reverts rancher/cattle#1016


ibuildthecloud:
Going with #1050 instead.

PR #1049, opened 2015-11-13
Fix allocation bugs introduced by shared storage

cjellick:
Bugs fixed (not recorded by QA because the feature isn't exposed yet):
- When specifying dataVolumeMounts, allocator cannot incrementally add a StoragePoolVolumeMap for each dataVolumeMount.
- When more than one storage pools is associated with a host, a race condition can cause the allocator to fail because the allocator would pick different storage pools on different runs
- When more than one storage pools is associated with a host, the placement of volumes to a storage pool was non-deterministic. They could land on any pool associated to the host.


cjellick:
@ibuildthecloud @alena1108 @wlan0 Can you review?

PR #1055, opened 2015-11-14
Enabled health check for LB service

alena1108:
https://github.com/rancher/rancher/issues/1721

### DIFF
--- a/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
+++ b/code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
@@ -120,6 +120,12 @@ public class LoadBalancerInstanceManagerImpl implements LoadBalancerInstanceMana
                     params.put(DockerInstanceConstants.FIELD_NETWORK_CONTAINER_ID,
                             data.get(DockerInstanceConstants.FIELD_NETWORK_CONTAINER_ID));
                 }
+                
+
+                if (data.get(InstanceConstants.FIELD_HEALTH_CHECK) != null) {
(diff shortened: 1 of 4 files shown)


code/iaas/lb-instance/src/main/java/io/cattle/platform/lb/instance/service/impl/LoadBalancerInstanceManagerImpl.java
```diff
@@ -120,6 +120,12 @@
                     params.put(DockerInstanceConstants.FIELD_NETWORK_CONTAINER_ID,
                             data.get(DockerInstanceConstants.FIELD_NETWORK_CONTAINER_ID));
                 }
+                
+
```
ibuildthecloud: There's two blank lines... I don't know if I can handle that much white space.

resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
```diff
@@ -83,6 +83,7 @@ backend ${listener.uuid}_${backend.uuid}_backend
         
 </#list>
 </#list>
-<#else>
-listen web 0.0.0.0:9
 </#if>
+
+listen default 0.0.0.0:42
```
ibuildthecloud: Why?  Can this ever conflict if I really want to listen on port :42?

alena1108: You are right, it can conflict. I just wanted to reserve some port for LB health check, so user won't have to define health check for each LB service in his environment, and health check will be enabled for every LB service by default. And we can just document that port 42 is reserved for health checks.  I can re-work it the other way - let user specify the port/health check, and configure haproxy accordingly if you think that would be a more valid solution.

PR #1062, opened 2015-11-16
Lookup named volumes on instance.create

cjellick:
Adds logic to support properly scheduling containers that are using
shared volumes. If a named volume is specified in dataVolumes like
this: `vol1:/container/path`, then during instance.create, look to see
if that volume exists in a non-local storage pool and if it does, use
that volume by adding an entry to dataVolumeMounts.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
@@ -1,0 +1,111 @@
+package io.cattle.platform.process.instance;
+
+import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.dao.StoragePoolDao;
+import io.cattle.platform.core.dao.VolumeDao;
+import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.model.StoragePool;
+import io.cattle.platform.core.model.Volume;
+import io.cattle.platform.engine.handler.HandlerResult;
+import io.cattle.platform.engine.handler.ProcessPreListener;
+import io.cattle.platform.engine.process.ProcessInstance;
+import io.cattle.platform.engine.process.ProcessState;
+import io.cattle.platform.object.ObjectManager;
+import io.cattle.platform.object.process.ObjectProcessManager;
+import io.cattle.platform.object.util.DataAccessor;
+import io.cattle.platform.process.common.handler.AbstractObjectProcessLogic;
+import io.cattle.platform.util.exception.ExecutionException;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Named
+public class InstanceVolumeLookupPreCreate extends AbstractObjectProcessLogic implements ProcessPreListener {
+
+    private static final Logger log = LoggerFactory.getLogger(InstanceVolumeLookupPreCreate.class);
+    private static final String LOCAL = "local";
+
+    @Inject
+    ObjectManager objectManager;
+    @Inject
+    ObjectProcessManager processManager;
+    @Inject
+    StoragePoolDao storagePoolDao;
+    @Inject
+    VolumeDao volumeDao;
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { "instance.create" };
+    }
+
+    /**
+     * Looks for shared volumes in the dataVolumes field and converts them to dataVolumeMounts.
+     * If volumeDriver is set:
+     *  if set to local, do nothing
+     *  if set to anything else, restrict volume lookup to a matching storage pool
+     *  if blank, do not restrict volume lookup to a particular storage pool.
+     * If volume is found, remove from dataVolumes list and put it in dataVolumeMounts. 
+     * If volume is not found, just leave it in dataVolumes.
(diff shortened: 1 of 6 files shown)


cjellick:
Ping @ibuildthecloud @alena1108

cjellick:
If you guys review this, let me merge tomorrow morning. I want to run it through a few quick real-world tests against a convoy cluster.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
```diff
@@ -0,0 +1,111 @@
...
+        long accountId = instance.getAccountId();
+        if (StringUtils.isNotEmpty(vd)) {
+            StoragePool sp = storagePoolDao.findStoragePoolByDriverName(accountId, vd);
+            if (sp != null) {
+                storagePoolId = sp.getId();
+            }
+        }
+
+        Map<Object, Object> data = new HashMap<>();
+        List<String> dataVolumes = DataAccessor.fieldStringList(instance, InstanceConstants.FIELD_DATA_VOLUMES);
```
alena1108: Should we add .withDefault(Collections.EMPTY_LIST) here? or may be its better to check for not null when do data.put on lines 85 and 86 to avoid storing null values unnecessarily.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
```diff
@@ -0,0 +1,111 @@
...
+                if (parts.length > 1) {
+                    try {
+                        Volume vol = volumeDao.findSharedVolume(accountId, storagePoolId, parts[0]);
+                        if (vol != null) {
+                            dataVolumeMounts.put(parts[1], vol.getId());
+                            continue;
+                        }
+                    } catch (IllegalStateException e) {
+                        log.error(e.getMessage());
+                        objectProcessManager.scheduleProcessInstance(InstanceConstants.PROCESS_REMOVE, instance, null);
```
alena1108: to speed up the removal, we can call scheduleProcessInstanceAsync here. Also curious whether we should handle situation when remove process fails to schedule here. The follow up exception will never be thrown

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
```diff
@@ -53,4 +60,45 @@ public void createVolumeInStoragePool(Map<String, Object> volumeData, StoragePoo
...
+        if (storagePoolId == null) {
+            volumes = create()
+            .select(VOLUME.fields())
+            .from(VOLUME)
+            .join(VOLUME_STORAGE_POOL_MAP)
+                .on(VOLUME_STORAGE_POOL_MAP.VOLUME_ID.eq(VOLUME.ID))
+            .join(STORAGE_POOL)
+                .on(VOLUME_STORAGE_POOL_MAP.STORAGE_POOL_ID.eq(STORAGE_POOL.ID)
+                .and(STORAGE_POOL.KIND.notIn(LOCAL_POOL_KINDS)))
+            .where(VOLUME.NAME.eq(volumeName)
```
alena1108: should probably also add not null check for STORAGE_POOL table.

cjellick: Doesn't the fact that I'm doing a join (aka inner join) on the storage_pool table guarantee that storage pool will not be null?

...I may be mistaken

cjellick: I had @vincent99 confirm: the join takes care of the null check.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
cjellick: `fieldStringList` already handles that and always returns an initialized list.

alena1108: @cjellick nice! learned something new :)

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
alena1108: @cjellick sorry, Craig, my bad. My thoughts clearly ran somewhere while I was typing that comment. We should check for Storage_pool.REMOVED field not being null, not that record being null.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
cjellick: Will number my responses so its easier for you to respond to them:
1. What is the impact of using `scheduleProcessInstanceAsync`? Nothing other than forcing the schedule to run immediately? Does it still behave like normal scheduling (as opposed to executing)? Also, in this case, _not_ speeding up the removal _might_ be beneficial in that it leaves the container in an error state a little longer, which turns into a decent visual cue in the UI that something went wrong (bc when its in an error state, the UI display an exclamation point).

2. This is basically the same logic as `io.cattle.platform.process.instance.InstanceStart.stopOrRemove(ProcessState, Instance, ExecutionException)`, which is executed if we cannot start a container (like if py-agent errors out). We don't handle the failure to schedule remove there and that logic is well established/exercised, so I think I'm comfortable not handling it here.

alena1108:
LGTM

PR #1064, opened 2015-11-16
Don't update haproxy config for unhealthy lb instances

alena1108:
https://github.com/rancher/rancher/issues/2694


cjellick:
LGTM.

PR #1065, opened 2015-11-16
Don't wait for lb config update

alena1108:
during the instance start. Instance start shouldn't be delayed because it takes time to program itself to the (n) lb services in the system or/and if lb service itself goes through activating/updating state. 

So with this fix, target instance.start will just schedule haproxy config item update. This fix is already in "Killed standalone LB" PR, so just porting it to master as health check use cases can be affected by the pre-fix behavior.  

https://github.com/rancher/rancher/issues/2697


cjellick:
LGTM

PR #1067, opened 2015-11-17
Fix race condition for volume external event

cjellick:
If an external volume event comes in at the same time as we are doing
native-docker population, it is possible to hit a race condition where
a volume is created twice. This fix prevents that by doing an
in-transaction lookup of the volume in the storage pool before creating
it.

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
@@ -53,8 +53,12 @@ public class VolumeDaoImpl extends AbstractJooqDao implements VolumeDao {
     }
 
     @Override
-    public void createVolumeInStoragePool(Map<String, Object> volumeData, StoragePool storagePool) {
-        Volume volume = resourceDao.createAndSchedule(Volume.class, volumeData);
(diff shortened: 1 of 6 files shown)


alena1108:
LGTM.

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
```diff
@@ -53,8 +53,12 @@ public Volume findVolumeByExternalId(Long storagePoolId, String externalId) {
     }
 
     @Override
-    public void createVolumeInStoragePool(Map<String, Object> volumeData, StoragePool storagePool) {
-        Volume volume = resourceDao.createAndSchedule(Volume.class, volumeData);
+    public void createVolumeInStoragePool(Map<String, Object> volumeData, String volumeName, StoragePool storagePool) {
```
alena1108: For other reviewers ref: all calls within this method will be executed as a part of same transaction as method is registered for this sort of execution here:

https://github.com/rancher/cattle/blob/4e25899064cd801a2066b856ed1e682d719a7579/code/iaas/model/src/main/resources/META-INF/cattle/system-services/spring-core-model-system-services-context.xml#L54

alena1108:
LGTM

PR #1068, opened 2015-11-17
add engineOpts and dockerVersion fields in machine

wlan0:
@cjellick @alena1108 

These fields are required for allowing a user to specify a docker version, and other docker engine options when performing the 'Add host' action for a cloud provider.


alena1108:
@wlan0 do we ever expect those fields to be set by the user? If so, we should add the permissions to user-auth.

wlan0:
These fields are required for allowing a user to specify a docker version, and other docker engine options when performing the 'Add host' action for a cloud provider.

wlan0:
@alena1108 Set 'create' permissions for "project" user on these fields. The "project" user will create these fields.

cjellick:
LGTM

PR #1070, opened 2015-11-17
Fixed bug in Global Service Deployment planner and other bug fixes around reconnecting agents

alena1108:
1) Cleanup instances that exist on an inactive host when do global deployment

https://github.com/rancher/rancher/issues/2711

2) Update: one more fix coming as a part of this PR - for standalone LB fix. Wait for the LB instance full start only when lb instance creation is requested by loadbalancerhostmap.create. This fix wouldn't be needed (and would be removed) once the "Killed standalone LB" PR is merged.

https://github.com/rancher/rancher/issues/2712

3) Update: trigger global service reconcile on agent reconnect:

https://github.com/rancher/rancher/issues/2713


cjellick:
LGTM

PR #1071, opened 2015-11-17
Fix volume storage pool allocation

cjellick:
If the volume's driver matches an existing storage pool driver name,
restrict the allocation to that storage pool;
otherwise, allow it to allocate to any pool that is not managed by
rancher.

### DIFF
--- a/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/BaseConstraintsProvider.java
+++ b/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/BaseConstraintsProvider.java
@@ -4,16 +4,16 @@ import io.cattle.platform.allocator.dao.AllocatorDao;
 import io.cattle.platform.allocator.service.AllocationAttempt;
 import io.cattle.platform.allocator.service.AllocationLog;
 import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.constants.VolumeConstants;
 import io.cattle.platform.core.dao.GenericMapDao;
+import io.cattle.platform.core.dao.StoragePoolDao;
 import io.cattle.platform.core.model.Host;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.model.Nic;
 import io.cattle.platform.core.model.StoragePool;
-import io.cattle.platform.core.model.StoragePoolHostMap;
 import io.cattle.platform.core.model.Subnet;
 import io.cattle.platform.core.model.Vnet;
 import io.cattle.platform.core.model.Volume;
-import io.cattle.platform.core.model.VolumeStoragePoolMap;
 import io.cattle.platform.object.ObjectManager;
 import io.cattle.platform.object.util.DataAccessor;
 import io.cattle.platform.object.util.DataUtils;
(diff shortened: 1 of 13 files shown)


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/BaseConstraintsProvider.java
```diff
@@ -106,49 +110,59 @@ protected void addComputeConstraints(AllocationAttempt attempt, List<Constraint>
     protected void addStorageConstraints(AllocationAttempt attempt, List<Constraint> constraints) {
         for (Map.Entry<Volume, Set<StoragePool>> entry : attempt.getPools().entrySet()) {
             Volume volume = entry.getKey();
-            List<? extends VolumeStoragePoolMap> vspms = mapDao.findNonRemoved(VolumeStoragePoolMap.class, Volume.class, volume.getId());
-            boolean alreadyMappedToPool = vspms.size() > 0;
+            boolean alreadyMappedToPool = entry.getValue().size() > 0;
             VolumeValidStoragePoolConstraint volumeToPoolConstraint = new VolumeValidStoragePoolConstraint(volume, alreadyMappedToPool);
-            if (alreadyMappedToPool) {
```
cjellick: @ibuildthecloud I added all this alreadyMappedToPool in my last PR when fixing allocation bugs. But, I realized that my `mapDao.findNonRemoved(VolumeStoragePoolMap.class ...)` lookup was just duplicating what was already done here: https://github.com/rancher/cattle/blob/0c4066f9fd2652f99d29989c2a29065f0378c20e/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AbstractAllocator.java#L161
and that the only real change we needed was that if the pools set returned by entry.getValue() was not empty, make the constraint an "exact" match.

So, my changes in this method actually simplify the logic back to before my last fix but then add the additional constraints needed to properly allocate stand-alone volumes.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/BaseConstraintsProvider.java
```diff
@@ -106,49 +110,59 @@ protected void addComputeConstraints(AllocationAttempt attempt, List<Constraint>
...
+            for (StoragePool pool : entry.getValue()) {
+                volumeToPoolConstraint.getStoragePools().add(pool.getId());
+                storagePoolToHostConstraint(constraints, pool);
+            }
+
+            if (!alreadyMappedToPool) {
+                String driver = DataAccessor.fieldString(volume, VolumeConstants.FIELD_VOLUME_DRIVER);
+                boolean restrictToUnmanagedPool = true;
+                if (StringUtils.isNotEmpty(driver) && !VolumeConstants.LOCAL_DRIVER.equals(driver)) {
+                    StoragePool pool = storagePoolDao.findStoragePoolByDriverName(volume.getAccountId(), driver);
```
alena1108: Do you think its worth to log the fact that if no pool satisfying volume's driver was found, so it was placed on unmanaged pool?

cjellick: @alena1108 I don't think it needs logged because not finding a pool is  a normal use case. Users can put whatever they want in the volumeDriver field and if it doesn't map to storage pool, it just means that they want to use a volume driver that we are not orchestrating.

For example, if they've already deployed flocker on the hosts as a docker plugin, they would put flocker as the volume driver and that would be just fine.

alena1108:
LGTM

PR #1073, opened 2015-11-17
Lower case only io.rancher labels keys

alena1108:
https://github.com/rancher/rancher/issues/2704


ibuildthecloud:
LGTM (assuming green)

alena1108:
Update: lower case not only affinity labels' keys, but any rancher system label (the one which key starts with "io.rancher")

PR #1075, opened 2015-11-18
Send detailed volume data on instance.activate

cjellick:
So that we can create volumes on-demand. Also, fix back-population to
account for volumes that don't yet belong to a storage pool.


ibuildthecloud:
LGTM

alena1108:
LGTM

PR #1076, opened 2015-11-18
Added publicIp/port information to host/service

alena1108:
https://github.com/rancher/rancher/issues/2695

With this PR, "service" and "host" objects get one extra API field called "publicEndpoints". Each endpoint has 2 fields:

* port (int)
* ipAddress (string)

Field name is debatable, I would agree with any name having better representation of ip/port combo.

Field gets populated for service/host on every nic.activate/nic.deactivate event for the instance having port exposed. Supported for both service and standalone instances; in standalone instance case only host object will get an update.

This model will be valid even in the future, when we add support for multiple interfaces for hosts + mapping port to a particular interface.

@vincent99 @ibuildthecloud API police, please review

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/constants/PortConstants.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/constants/PortConstants.java
@@ -5,4 +5,7 @@ public class PortConstants {
     public static final String KIND_USER = "userPort";
     public static final String KIND_IMAGE = "imagePort";
 
+    public static final String PROCESS_PORT_ACTIVATE = "port.activate";
+    public static final String PROCESS_PORT_DEACTIVATE = "port.deactivate";
+
 }
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/HostDaoImpl.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/HostDaoImpl.java
@@ -1,14 +1,25 @@
 package io.cattle.platform.core.dao.impl;
 
-import static io.cattle.platform.core.model.tables.HostTable.*;
+import static io.cattle.platform.core.model.tables.HostIpAddressMapTable.HOST_IP_ADDRESS_MAP;
+import static io.cattle.platform.core.model.tables.HostTable.HOST;
+import static io.cattle.platform.core.model.tables.IpAddressTable.IP_ADDRESS;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.dao.HostDao;
 import io.cattle.platform.core.model.Host;
+import io.cattle.platform.core.model.IpAddress;
+import io.cattle.platform.core.model.tables.records.HostRecord;
+import io.cattle.platform.core.model.tables.records.IpAddressRecord;
 import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
+import io.cattle.platform.object.ObjectManager;
 
 import java.util.List;
 
+import javax.inject.Inject;
+
 public class HostDaoImpl extends AbstractJooqDao implements HostDao {
+    
+    @Inject
+    ObjectManager objectManager;
 
     @Override
     public List<? extends Host> getHosts(Long accountId, List<String> uuids) {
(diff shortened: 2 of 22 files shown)


vincent99:
Description sounds good to me..

ibuildthecloud:
I like it!   Need to review one more time on a screen larger than 4"

code/iaas/model/src/main/java/io/cattle/platform/core/addon/IpPortPublicEndpoint.java
```diff
@@ -0,0 +1,63 @@
+package io.cattle.platform.core.addon;
+
+import io.github.ibuildthecloud.gdapi.annotation.Type;
+
+@Type(list = false)
+public class IpPortPublicEndpoint {
```
ibuildthecloud: Can we change this to PublicEndpoint, as the type ipPortPublicEndpoint will be in the API, ipPort seems unneeded

code/iaas/model/src/main/java/io/cattle/platform/core/constants/PortConstants.java
```diff
@@ -5,4 +5,8 @@
     public static final String KIND_USER = "userPort";
     public static final String KIND_IMAGE = "imagePort";
 
+    public static final String PROCESS_PORT_ACTIVATE = "port.activate";
+    public static final String PROCESS_PORT_DEACTIVATE = "port.deactivate";
+
+    public static final String FIELD_PUBLIC_PORTS = "publicPorts";
```
ibuildthecloud: Is this constant used?

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/HostDaoImpl.java
```diff
@@ -1,14 +1,28 @@
...
+import io.cattle.platform.lock.LockManager;
+import io.cattle.platform.object.ObjectManager;
 
 import java.util.List;
 
+import javax.inject.Inject;
+
 public class HostDaoImpl extends AbstractJooqDao implements HostDao {
+    @Inject
+    LockManager lockManager;
```
ibuildthecloud: Is this used?

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/InstanceDaoImpl.java
```diff
@@ -143,4 +149,29 @@ public Instance getInstanceByUuidOrExternalId(Long accountId, String uuid, Strin
                 .fetchInto(InstanceRecord.class);
     }
 
+    @Override
+    public Long getInstanceHostId(long instanceId) {
+        List<? extends InstanceHostMap> maps = mapDao.findNonRemoved(InstanceHostMap.class, Instance.class,
+                instanceId);
+        if (!maps.isEmpty()) {
```
ibuildthecloud: nit: its easier to read code when the first block of a if statement is the true case.  so in this situation.

```java
if (maps.isEmpty()) {
  return null
}

return maps.get(0).getHostId();
```

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -298,4 +312,51 @@ public boolean isGlobalService(Service service) {
...
+            if (add) {
+                publicEndpoints.add(publicEndpoint);
+            } else {
+                publicEndpoints.remove(publicEndpoint);
+            }
+            DataUtils.getWritableFields(service).put(ServiceDiscoveryConstants.FIELD_PUBLIC_ENDPOINTS,
+                    publicEndpoints);
+            objectManager.persist(service);
+            Map<String, Object> data = new HashMap<>();
+            data.put(ServiceDiscoveryConstants.FIELD_PUBLIC_ENDPOINTS, publicEndpoints);
```
ibuildthecloud: We've copied and pasted this code in a couple places now.  Can we make it generic somewhere.  So you can just call `utils.TriggerStateChanged(eventService, resource)`.  And the call to eventService.publish should be deferred.

ibuildthecloud:
@alena1108 Just a couple small comments.  I like how this just keys off of port activate/deactivate.  Very clean, I hope it stays that simple.

code/iaas/model/src/main/java/io/cattle/platform/core/addon/IpPortPublicEndpoint.java
alena1108: sure

code/iaas/model/src/main/java/io/cattle/platform/core/constants/PortConstants.java
alena1108: nope, leftovers from the previous design, will remove

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/HostDaoImpl.java
alena1108: will cleanup

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
alena1108: sure

PR #1077, opened 2015-11-19
Fix query that looks up volumes by name

cjellick:
Query was unintentionally restricting results to volumes that had
already been allocated to a storage pool. But brands new volumes
created via the API are not allocated to a pool until after they are
used by a container.

Addresses rancher/rancher#2727
Also maybe rancher/rancher#2726, but I need to research that one more.


alena1108:
LGTM

ibuildthecloud:
LGTM

PR #1078, opened 2015-11-19
Better errors messages from invalid identities.

hibooboo2:
This addresses rancher/rancher#2031


ibuildthecloud:
Does this change the error message seen during login? Or only when looking up identities?

hibooboo2:
Only when looking up identities during a call of `project.setmembers`.

PR #1079, opened 2015-11-19
Fixes for bugs caused by Standalone LB removal refactoring 

alena1108:
1) Fixed stack.restart use case for lb service

https://github.com/rancher/rancher/issues/2735

2) Added not null check for ports: can be null for internal LB

https://github.com/rancher/rancher/issues/2737


cjellick:
LGTM

PR #1080, opened 2015-11-19
And the award for most confusing bug with simplest fix goes to...

cjellick:
Fix allocation query ordering

This query returns combinations of host and storage pool ids order by
compute_free. The code that uses this query expects hosts will be
grouped together by id such that results look like this:
```
host1, spA
host1, spB
host2, spA
host2, spC
```
but the secondary ordering by host was not guaranteed and you could end
up with a result set like this:
 ```
host1, spA
host2, spA
host1, spB
host2, spC
```


alena1108:
LGTM

cjellick:
Addreses
rancher/rancher#2728
rancher/rancher#2726
rancher/rancher#2725

PR #1081, opened 2015-11-19
Custom Lb config support, "defaults" section

alena1108:
This PR allows user to specify custom values for haproxy "defaults" section.

API changes: 

* loadBalancerConfig - embedded object passed as a field to loadBalancerService on its Create/Update - gets a new field "defaults" of type array[string]
* rancher-compose.yml representation:

<pre>
nginx:
  scale: 1
lb:
  scale: 1
  load_balancer_config:
    <b>defaults:
      - balance leastconn
      - retries 3
      - option keep-alive
      - option forwardfor
</b>
</pre>

The reason for going with array[string] as opposed to map[string] - there are duplicated keys like "option"

https://github.com/rancher/rancher/issues/1871
https://github.com/rancher/rancher/issues/876

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/LoadBalancerInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/LoadBalancerInfoDaoImpl.java
@@ -105,8 +105,8 @@ public class LoadBalancerInfoDaoImpl implements LoadBalancerInfoDao {
                     sourceProtocol = "https";
                 }
             }
-            listenersToReturn.add(new LoadBalancerListenerInfo(privatePort, sourcePort,
-                    sourceProtocol, targetPort));
+            listenersToReturn.add(new LoadBalancerListenerInfo(lbService, privatePort,
+                    sourcePort, sourceProtocol, targetPort));
         }
         return listenersToReturn;
     }
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/LoadBalancerListenerInfo.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/LoadBalancerListenerInfo.java
@@ -1,6 +1,6 @@
 package io.cattle.platform.configitem.context.data;
 
-import java.util.UUID;
+import io.cattle.platform.core.model.Service;
 
 public class LoadBalancerListenerInfo {
     Integer privatePort;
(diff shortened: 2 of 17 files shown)


alena1108:
@ibuildthecloud https://github.com/rancher/rancher/issues/2753

alena1108:
@ibuildthecloud just spoke to @vincent99. Along with "defaults" section users might want to customize haproxy sections like "userlists", "peers", etc. So it all comes down to allowing user to define their own haproxy config, or part of it as we can't allow to modify frontend/backend sections as we form/control them.

In order to support it, we can't limit customization to just "defaults" section. So instead of accepting list of strings, we should accept a string for loadBalancerConfig.defaults parameter (name is a subject for discussion). This string would represent the entire customizable section of haproxy.cfg, its value would look like:

```
global
	log 127.0.0.1 local0
    	log 127.0.0.1 local1 notice
        maxconn 4096

defaults
	log	global
	mode	tcp
	option	tcplog
        option  dontlognull
        option  redispatch
        option http-server-close
        option forwardfor
        retries 3
        timeout connect 5000
        timeout client 50000
        timeout server 50000
	errorfile 400 /etc/haproxy/errors/400.http
	errorfile 403 /etc/haproxy/errors/403.http
	errorfile 408 /etc/haproxy/errors/408.http
	errorfile 500 /etc/haproxy/errors/500.http
	errorfile 502 /etc/haproxy/errors/502.http
	errorfile 503 /etc/haproxy/errors/503.http
	errorfile 504 /etc/haproxy/errors/504.http
```

If parameter is not passed, we will apply Rancher defaults.

andyshinn:
I know you are probably not using Ruby. But if you are interested in accepting JSON in the API, there is an example of a HAProxy to JSON transformer at https://github.com/progrium/configurator/blob/master/transformers/haproxy-cfg. In the test directory you can see an example input and output.

alena1108:
@andyshinn thanks for suggestion, cool project.

We are going to go with string type for now. We are going to create an extra haproxy config file with the contents of this string, and pass it as an extra config on haproxy load. Something like:

service haproxy start -f defaultRancherConfig -f customConfig

andyshinn:
Great, that would be helpful and solve a lot of immediate use cases.

alena1108:
@vincent99 @ibuildthecloud I've updated the PR, could you please review? Now we accept haproxyConfig (type=string), and generate/pass one extra config when launch haproxy. I'm going to give it more testing to check if its possible to override even listen/backend sections.

alena1108:
Did more testing for haproxy accepting multiple configs. The results is not something we can go with:

1) Only Global and User sections defined in multiple files, are being merged.
2) "defaults" section parameters apply only to the listeners/backends defined in very same haproxy config file. Did test is by defining "balance roundrobin" in config1 and  "balance first" in config2. "balance first" didn't override algorithm for the backends in the first config.
3) If you have backends/listeners with duplicated names in both configs, haproxy errors out. But at the same time, if you have listener with diff names and the same listening port in both config, listener of config2 will override listener of config1.

I'm sure there are more uncovered merge related caveats, and as unfortunately its not documented on haproxy side, so we can't use "passing 2 configs" approach.

Most likely we will go with writing our own merging tool which appends user defined parameters to corresponding section of Rancher default haproxy config

alena1108:
The PR is updated with code merging 2 haproxy configs on the lb agent side. 
New use case flow: 

1) user passes haproxy config to loadBalancerService.haproxyConfig parameter. Important: indentation matters. Here are the rules:

```
defaults  >> section name should come as unindented
    balance roundrobin >> parameters/options should come indented
```

2) custom haproxy config will get merged with the config generated by Rancher, on network-agent side, and haproxy will be launched using this merged config.

alena1108:
Closing in favor of https://github.com/rancher/cattle/pull/1160

PR #1083, opened 2015-11-20
Omit duplicated "default" entries in dns config

alena1108:
When host has at least one container with host networking, "default" entry is generated for this container in java code (dnsEntry.sourceIpAddress.address="default" for such entry). We don't want to populate default "default" section in this case.

https://github.com/rancher/rancher/issues/2460

### DIFF
--- a/resources/content/config-content/hosts/content-home/etc/cattle/dns/answers.json.ftl
+++ b/resources/content/config-content/hosts/content-home/etc/cattle/dns/answers.json.ftl
@@ -1,17 +1,10 @@
 {
-    "default": {
-        "recurse": [
-            PARENT_DNS
-        ],
-        "a": {
-        	"rancher-metadata.": {"answer": ["169.254.169.250"]},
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
@alena1108 I want to confirm that if there are zero containers with `--net=host` that we still generate a useful default section for dns and metadata.  For example, quite often I run `docker run ...` on a host and talk to the metadata service because /self/host and some others still work.  I know this is for DNS, but wanted to confirm the behavior between both DNS and metadata

alena1108:
@ibuildthecloud confirming. If there are 0 containers with DNS metadata, old regular "default" section will be generated anyway:

```
 "default": {
        "recurse": [
            PARENT_DNS
        ],
        "a": {
        	"rancher-metadata.": {"answer": ["169.254.169.250"]},
		"rancher-metadata.rancher.internal.": {"answer": ["169.254.169.250"]}
		}
    }
```

PR #1084, opened 2015-11-20
Fix volumeDriver to pool to host allocation bug

cjellick:
If a volumeDriver is specified on an instance and that volumeDriver
resolves to a pool, allocation should be restricted to hosts belonging
to that storage pool.


ibuildthecloud:
LGTM

cjellick:
Addresses rancher/rancher#2725

PR #1091, opened 2015-11-20
Lookup volumes by driver, not pool id

cjellick:
When a volume is first created via the API, it is not in a storage
pool. If that volume is referenced by name in an instance, look it up
by driver instead of storage pool id, since it may not yet be in a
storage pool.

Bonus: disable volume delete event logic.
Double bonus: don't unnecessarily query for storage pools when driver
is local in allocatio logic.

Addresses rancher/rancher#2755


alena1108:
LGTM

PR #1093, opened 2015-11-21
Add a volume affinity label to assist in placement

cjellick:
Volume affinity
io.rancher.scheduler.affinity:volume = volume1,volume2

Behavior:
- If this label is present while scheduling, constrain the scheduling of the container to hosts where those volumes are already available, regardless of whether those volumes are part of shared or local pools.
- If the labels point at shared volume, the container can land on any host in the storage pool.
- If the values in this label conflict with each or with the values of dataVolumeMounts, it is a scheduling failure.
- If this label is present, looking up volumes in shared storage pools _still_ takes precedence over local storage pools.
- If a value is in the label but not in volumeMounts, the label is ignored.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
@@ -1,5 +1,6 @@
 package io.cattle.platform.process.instance;
 
+import static io.cattle.platform.core.model.tables.VolumeTable.*;
 import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.constants.VolumeConstants;
 import io.cattle.platform.core.dao.StoragePoolDao;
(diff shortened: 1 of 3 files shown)


cjellick:
Use case:
A user wants to guarantee that a container uses a specific LOCAL volume but doesn't want to or cant use dataVolumeMounts to specify its ID. If they were to just specify the volume in dataVolumes, we would not force the container to an existing local volume. If they add it to the label, we _will_ force the container to use an existing local volume.

This is important in rancher-compose, which cannot expose dataVolumeMounts.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceVolumeLookupPreCreate.java
```diff
@@ -62,8 +71,15 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
             return null;
         }
 
+        Object va = DataAccessor.fieldMap(instance, InstanceConstants.FIELD_LABELS).get(LABEL_VOLUME_AFFINITY);
+        Set<String> volumeAffinities = new HashSet<String>();
+        if (va != null) {
+            volumeAffinities.addAll(Arrays.asList(va.toString().split(",")));
+        }
+
         String driver = DataAccessor.fieldString(instance, InstanceConstants.FIELD_VOLUME_DRIVER);
```
ibuildthecloud: Should "" and null be considered the same as local driver?

PR #1094, opened 2015-11-22
Unringing this bell

cjellick:
Change named dataVolume lookup behavior

In general, change behavior so that container.volumeDriver is only used
for volume creation, not named volume lookup. More specifically, change
behavior such that:
- 'local' is treated exactly the same as blank when specified as
container.volumeDriver.
- volume lookup is not restricted to the shared storage pool matching
  the volumeDriver field.


ibuildthecloud:
LGTM

PR #1095, opened 2015-11-23
Fixes around publicEndpoints feature

alena1108:
1) Don't update public endpoint where publicIP or port is null

https://github.com/rancher/rancher/issues/2773
https://github.com/rancher/rancher/issues/2771

2) Update public endpoints on port.update

https://github.com/rancher/rancher/issues/2770


cjellick:
LGTM

PR #1099, opened 2015-11-23
Metadata: order containers by start_index, and other bug fixes

alena1108:
1) Metadata: order containers by start_index

https://github.com/rancher/rancher/issues/2774

@cloudnautique @ibuildthecloud this small PR is to ensure that container order in metadata is done based on start_index. Currently done in ascending order, please comment if you think it should be DESC.

2) Limit length for stack.name, service.name and sidekick.name to 63 chars as its the max allowed length for dns labels.

https://github.com/rancher/rancher/issues/2742

3) Handle publicEndpoint update for the case when privately exposed port gets updated with a public port:

https://github.com/rancher/rancher/issues/2770

4) Update service.publicEndpoints for all instances, not just managed:

https://github.com/rancher/rancher/issues/2787


cjellick:
LGTM

PR #1102, opened 2015-11-24
Order containers by create_index when form metadata

alena1108:
Instead of mistakenly put start_count


cjellick:
LGTM

alena1108:
Unrelated TF, @cjellick is looking into it.

PR #1103, opened 2015-11-24
Greenify master

cjellick:
Fix volume test by deleting container first

This test is purging volumes. When docker 1.9 is on the host, this translates to a `docker volume rm`, but this test is doing the volume purge while there are still containers using the volume. So, we need to delete the containers first so that the purge can finish cleanly.

We probably need to address this in a larger context by preventing volume purge if the volume has active container mounts. I noted this in #2796


alena1108:
LGTM

PR #1104, opened 2015-11-24
Add Wildcards on haproxy ACLs

sekipaolo:
As result of fixing https://github.com/rancher/rancher/issues/1503 we lost any way to apply more complex routing.
With wildcards we could have multiple lb layers management


alena1108:
@sekipaolo funny, was working on the same task yesterday night, and submitted this PR:

https://github.com/rancher/cattle/pull/1107

The PR lets you specify match conditions for both host and path supported prefixes:

* beg
* dir
* dom
* end
* reg
* sub

ParametersExample:

example.com{end}/path{beg}=81

@ibuildthecloud @vincent99 I'm fine with either approach, could you please take a look at both PRs?

alena1108:
@sekipaolo after talking to @vincent99 and @ibuildthecloud we've decided that it would require less modifications (and new syntax explanation) if we go your way of implementing things. So following translation would be done for "host" value in host name routing rule:

1) domain.com - will get translated to hdr(domain.com)
2) domain.com*. - will get translated to hdr_end(domain.com)
3) *.domain.com - will get translated to hdr_beg(domain.com)

1) is already in place; and 2) is added by your PR. Do you want me to implement 3), or you want to do it as a part of the same PR?

vincent99:
@alena1108 it should strip just the star, not the dots from the value... `*.domain.com` should become `hdr_end(.domain.com), and `domain.com.*` -> `hdr_beg(domain.com.)` (though that one makes little sense, something like `rancher.*` might be useful.)

If you strip off the dot then `otherdomain.com` would match `*.domain.com`

alena1108:
@vincent99 right, thank you

sekipaolo:
@alena1108  of course. I did a rebase because tests where failing due a different issue and I wasn't able to setup a test environment. I will reapply the PR soon. There are also 2 other features I would like to work on: add options for set a default backend and enable stats on the load balancer. Do you know if there is already some WIP for this?

alena1108:
@sekipaolo we have one WIP PR for haproxy: https://github.com/rancher/cattle/pull/1160 - feature allowing users to specify custom parameters for "defaults" and "global" haproxy sections.

Thank you for willing to contribute! Before writing an actual code, could you open a ticket for feature + feature implementation proposal on https://github.com/rancher/rancher/issues?

PR #1107, opened 2015-11-24
Fuzzy search for LB hostname routing rules

alena1108:
Path and domain can be appended with condition. Condition format {condition}. Example:
 * example.com{end}/path{beg}=81

supported prefixes:

* beg (prefix match)
* dir (subdir match)
* dom (domain match)
* end (suffix match)
* reg (regex match)
* sub (substring match)

Description for all prefixes can be found here:

https://cbonte.github.io/haproxy-dconv/configuration-1.5.html#7.3.6-req.hdr

Default value - exact match

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/util/LoadBalancerTargetPortSpec.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/util/LoadBalancerTargetPortSpec.java
@@ -7,6 +7,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 
 /**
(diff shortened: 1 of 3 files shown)


alena1108:
@vincent99 @ibuildthecloud ^^

vincent99:
oy... This is definitely useful but I hate that the targets are crazy label strings already, and this is 2 more options for both the API and UI to parse and stringify in separate languages and attempt to explain to a human in docs.

Can we continue discussing/work on making this kind of thing first-class API and YAML fields?  e.g. https://github.com/rancher/rancher/issues/2590

alena1108:
@vincent99 that would be the best, but I'm not sure when I can get to do that major LB refactoring. As fuzzy search is one of the most frequently requested features, so I've decided to give this PR a try.

alena1108:
Closing in favor of https://github.com/rancher/cattle/pull/1104

PR #1108, opened 2015-11-24
Add context to the callable.

hibooboo2:
Fix an issue where the context is null during a callable that requires it.
This as a result made it so that the identityprovider for github was unable to
get an access token for github.


ibuildthecloud:
You need to wrap the call in the in the Callable to call ManagedContext.callWithContext()

ibuildthecloud:
Use io.cattle.platform.util.exception.ExceptionUtils.throwRuntime(String, Throwable), don't print stack trace

PR #1116, opened 2015-11-30
Expose healthCheckInstanceHostMap to the API

alena1108:
1) Expose health check reported by hosts, via API

https://github.com/rancher/rancher/issues/2128

* added instanceId field to healthCheckInstanceHostMap, so the object links to instance directly via API
* changed initial state from healthy to initializing for healthCheckInstanceHostMap

2) Schedule health check on host only when Healthcheck network service provider is present there. Also:

* reconcile health check scheduling for all instances with health check enabled, when Healthcheck network service provider  (network-agent) gets deployed
* remove health check host maps on network-agent remove

https://github.com/rancher/rancher/issues/2287


ibuildthecloud:
@alena1108 Is there any way that we can just launch the network service agent?

alena1108:
@ibuildthecloud that would require some refactoring in way we do 2 things. 

1)  Refactor the way vnet is created for host. Today its triggered by user instance creation:

https://github.com/rancher/cattle/blob/0c4066f9fd2652f99d29989c2a29065f0378c20e/code/implementation/host-only-network/src/main/java/io/cattle/platform/networking/host/process/HostOnlyNicActivate.java#L65

2) AgentInstanceApply config items can be triggered by process on the resource that can be translated to nic on the same vnet as network agent:

https://github.com/rancher/cattle/blob/9b0872956bb9f7ac48b22d1113d8023d7cb54124/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/process/AgentInstanceApplyItems.java#L41

and that nic can't be the nic of agent instance itself. 

So initial healtcheck config can apply on the agent-instance initial start, but all subsequent changes won't be propagated. 

It all comes down to the requirement to have client nic that triggers network-agent creation and config item apply.

Let me know what you think

ibuildthecloud:
@alena1108 A bit longer term is that we need to break out the healthcheck into a system stack and deploy it separately.  Then we don't really have this issue because we will just globally schedule the healthcheck instances.  The current behavior you have is fine for the short term.

alena1108:
@ibuildthecloud makes sense

PR #1121, opened 2015-12-03
Performance improvements to DNS records generation

alena1108:
Query db just 2 times to get a) all services with their instances b) relationship between the services; store this info in memory. Form dns records based on this data


alena1108:
Also fixes https://github.com/rancher/rancher/issues/2890

PR #1122, opened 2015-12-03
Update metadata on instancehostmap.activate

alena1108:
as its when docker native containers' labels are being populated

https://github.com/rancher/cattle/blob/e9ca0a4f9946483dee4220f9f1f392c330bf7334/code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instancehostmap/DockerPostInstanceHostMapActivate.java#L127

@ibuildthecloud @yasker

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
@@ -94,7 +94,8 @@ public class MetaDataInfoDaoImpl extends AbstractJooqDao implements MetaDataInfo
                 .and(instance.REMOVED.isNull())
                 .and(instance.STATE.notIn(CommonStatesConstants.REMOVING, CommonStatesConstants.REMOVED))
                 .and(exposeMap.REMOVED.isNull())
-                .and(exposeMap.STATE.notIn(CommonStatesConstants.REMOVING, CommonStatesConstants.REMOVED))
+                .and(exposeMap.STATE.isNull().or(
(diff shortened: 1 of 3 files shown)


yasker:
Thanks @alena1108 !!

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
```diff
@@ -94,7 +94,8 @@ protected ContainerMetaData map(List<Object> input) {
                 .and(instance.REMOVED.isNull())
                 .and(instance.STATE.notIn(CommonStatesConstants.REMOVING, CommonStatesConstants.REMOVED))
                 .and(exposeMap.REMOVED.isNull())
-                .and(exposeMap.STATE.notIn(CommonStatesConstants.REMOVING, CommonStatesConstants.REMOVED))
+                .and(exposeMap.STATE.isNull().or(
```
ibuildthecloud: how can state be null?

alena1108: when exposeMap entry is null. I do left outer join with this table to get service and stand alone containers

ibuildthecloud: ah hah

alena1108:
@ibuildthecloud all the TFs are caused by the fact that instancehostmap.activate is a part of "agent.instance.services.processes" and "host.agent.instance.services.processes" now (and its the only config item present in both places). TFs happen only with idempotent checks on; network-agent instance start hangs as config items apply gets stuck.

What do you think can be the cause of this problem, and what would be the best way to workaround it?

PR #1125, opened 2015-12-03
Upgrade machine and go-machine-service

cjellick:


cjellick:
verbal lgtm from @ibuildthecloud

PR #1147, opened 2015-12-08
Upgrade gms to v0.23.0

cjellick:


cjellick:
Address 
rancher/rancher#2839
rancher/rancher#2576

Issue #1150, opened 2015-12-09
IPtables are effectively unconfigurable on agents.


stresler:
We're working on locking down our agent machines and have hit an issue that exists with Cattle (and Docker). 

When the agent starts it puts a catchall rule in nat PREROUTING table that directs all traffic through the CATTLE_PREROUTING chain. This is done here: 

https://github.com/rancher/cattle/blob/0c4066f9fd2652f99d29989c2a29065f0378c20e/resources/content/config-content/host-iptables/apply.sh#L17-L21

As a user we want to be able to add filtering rules in front of my container nat.  e.g. This dev environment should only accept these whitelisted IPs. 

However, in order to get upstream of the nat I would need to do filtering in the mangle table, which is universally a bad practice. 

I'd like to make a patch that adds an environment variable the agent could take: 

CATTLE_IPTABLES_INTERMEDIARY_CHAIN=xyzxyz

which once tracked through here: 

https://github.com/rancher/rancher/blob/master/agent/run.sh#L81-L110

I think should 'just work'. 

That could be documented with the understanding that this filter should direct anything that needs to be handled by rancher to CATTLE_PREROUTING.  

Eventually, I think Rancher will want to have something more sophisticated than this for managing iptables filtering, but this is sufficient to let users do sophisticated pre-nat filtering. 

I don't currently think that anything would need to be done for postrouting. 

I'll have a patch soon, but does this approach seem sane?

PR #1152, opened 2015-12-09
Adding CATTLE_IPTABLES_INTERMEDIARY_CHAIN support if the variable is …

stresler:
…defined.

PR so we can discuss Issue #1150

### DIFF
--- a/resources/content/config-content/host-iptables/apply.sh
+++ b/resources/content/config-content/host-iptables/apply.sh
@@ -14,11 +14,21 @@ fi
 
 apply_config iptables-restore -n etc/cattle/host-iptables
 
-for i in PRE POST; do
-    if ! iptables -t nat -n -L ${i}ROUTING | grep -q CATTLE_${i}ROUTING; then
-        iptables -t nat -I ${i}ROUTING -j CATTLE_${i}ROUTING
-    fi
-done
+if ! iptables -t nat -L | grep Chain | grep -q CATTLE_HOOK_PREROUTING
+    iptables -t nat -N CATTLE_HOOK_PREROUTING
+fi
+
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
@stresler I wonder if we could make this simpler in that we just always have a special chain that call first.  So if the chain doesn't exist we create it (same as you do here), but otherwise we don't alter the chain.  Then there is no real configuration the user has to do of the agent, they would just have to know to put rules in that special chain if they want prefiltering.

stresler:
That would totally suffice. I could modify this patch if you want. I expect that special chain would just have a similar catchall at the bottom of it to get everything to CATTLE_PREROUTING next?

ibuildthecloud:
@stresler If you could modify this PR that would be great.  It's basically the same approach and structure that you specified, but just assume there is no configuration an CATTLE_IPTABLES_INTERMEDIARY_CHAIN is always set to something fixed.

stresler:
If this suits I'll squash it. Do you guys have a commit message methodology you prefer?

ibuildthecloud:
@stresler We prefer PRs to be squashed.

ibuildthecloud:
@stresler And not picky about messages.

stresler:
Well, I did something wrong here.  I'll need to figure out how I mangled this PR first thing tomorrow though. Need to head out for the day.

stresler:
I'm not sure that this PR is causing a 'No Route To Host' in Drone.  Is there a way to trigger a re-run of that test?

stresler:
Hold off on this. Experimenting some more.

stresler:
This is not going to be the right way to deal with this issue. I'll update #1150

vincent99:
FYI the "no route to host" thing is just our drone being broken today.

PR #1160, opened 2015-12-10
Custom defaults and global for haproxy

alena1108:


resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
```diff
@@ -12,6 +12,9 @@ global
 	user haproxy
 	group haproxy
 	daemon
+	<#if haproxyConfig?? && haproxyConfig.global??>
```
ibuildthecloud: for future reference, freemarker has a shorter syntax for this I believe it is `(haproxyConfig.global)??`

PR #1165, opened 2015-12-11
Random public port assignment to a service (if not specified)

alena1108:
https://github.com/rancher/rancher/issues/2097

1) Port range is defined on per environment (project) basis. New API field is introduced for the project:

servicesPortRange (string, optional, default value "49153-65535", CRU permissions)

2) If service has a port defined as "80" or "80/tcp" (with no public port specified), public port will get automatically assigned to the port from environment.servicesPortRange

3) Port gets assigned to the service on its creation, and the assignment gets removed after service is removed. Once disassociated, the port can be used by other services.

Random port can't have duplicates among other ports generated in random manner.
We don't validate random port against public ports specified by the user explicitly. The assumption is that the servicesPortRange is reserved for random allocation, and user shouldn't specify ports from this range when creates service/standalone instance.

@ibuildthecloud @vincent99 please review

Also split test_svc_discovery test into smaller files

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
@@ -7,11 +7,13 @@ import io.cattle.platform.core.addon.LoadBalancerServiceLink;
 import io.cattle.platform.core.addon.PublicEndpoint;
 import io.cattle.platform.core.addon.ServiceLink;
 import io.cattle.platform.core.constants.CommonStatesConstants;
+import io.cattle.platform.core.constants.InstanceConstants;
 import io.cattle.platform.core.constants.LoadBalancerConstants;
 import io.cattle.platform.core.constants.NetworkConstants;
 import io.cattle.platform.core.constants.SubnetConstants;
 import io.cattle.platform.core.dao.LabelsDao;
 import io.cattle.platform.core.dao.NetworkDao;
+import io.cattle.platform.core.model.Account;
 import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.core.model.Host;
 import io.cattle.platform.core.model.Instance;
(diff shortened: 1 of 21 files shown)


alena1108:
We would need a UI support for modifying default port range for the environment @vincent99

msh100:
To expand on this a little more, it would also be nice if there was the ability to increment port numbers for a service (assuming this isn't what's happening here). That way you get more predictable port numbering patterns.

For example, start range number + container ID in service.

ibuildthecloud:
@msh100 The way this works is that the service gets a random port assigned and that port is used with every single container.  If the user puts `ports: 80` it effectively becomes `ports: 45242:80`.  This way every container has the same port but just on different hosts.

msh100:
@ibuildthecloud ah thanks for the clarification - is there currently a way of exposing a port and having multiple containers of that service on a single host? As I am assuming this PR doesn't help out there then.

vincent99:
@msh100 Run the service without publishing ports with whatever scale you want, then a load balancer pointing to it and listening on the actual public port with scale <= number of hosts.

ibuildthecloud:
@alena1108 Can you remove: code/packaging/app/keys.properties.  You might as well put that in the `.gitignore`

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -239,6 +243,60 @@ public void setVIP(Service service) {
...
+    public void setPorts(Service service) {
+        List<PortSpec> ports = getServicePorts(service);
+        Map<String, Object> launchConfigData = DataAccessor.fields(service)
+                .withKey(ServiceDiscoveryConstants.FIELD_LAUNCH_CONFIG).withDefault(Collections.EMPTY_MAP)
+                .as(Map.class);
+        Account env = objectManager.loadResource(Account.class, service.getAccountId());
+        List<String> newPorts = new ArrayList<>();
+        for (PortSpec port : ports) {
+            if (port.getPublicPort() == null) {
+                PooledResource resource = poolManager.allocateOneResource(env, service,
```
ibuildthecloud: I don't think this will work if the service has multiple random ports.  The allocateOneResource will return the same port each time.  Can you write a test that has a service with more that one random port that needs to be assigned.

alena1108: sure, will do

PR #1176, opened 2015-12-11
Log message not stacktrace.

hibooboo2:
rancher/rancher#2998


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/LDAPIdentityProvider.java
```diff
@@ -159,7 +159,7 @@ protected Identity getObject(String distinguishedName, String scope) {
             return attributesToIdentity(object);
         }
         catch (NamingException e) {
-            getLogger().info("Failed to get object: " + distinguishedName, e);
+            getLogger().info("Failed to get object: " + distinguishedName, e.getExplanation());
```
cjellick: @hibooboo2 Did you hit this code and verify that it outputs what you expect? I don't think it is going to. I think the string  needs to be a formatted string with {} where you want to do variable replacement.

cjellick:
LGTM

PR #1178, opened 2015-12-11
Don't hardcode lb algorithm

alena1108:
And ensure that "frontend*" name is consistent.

https://github.com/rancher/rancher/issues/876


cjellick:
LGTM

PR #1184, opened 2015-12-14
Fixed bug when orphaned containers were left around

alena1108:
after corresponding service got removed.

https://github.com/rancher/rancher/issues/1832


ibuildthecloud:
I'd rather not... I think there is a more correct fix we can do in the process manager code.

alena1108:
@ibuildthecloud ok, gonna close it for now then.

PR #1185, opened 2015-12-14
Timeout faster also invalidate pool after new config made.

hibooboo2:
rancher/rancher#2987


ibuildthecloud:
@hibooboo2 can you rebase?

PR #1194, opened 2015-12-16
Metadata config refactoring: 1) from json to yml format 2) versioning 3) Services as nested objects to stack

alena1108:
and containers as nested objects for a service. It makes it easier to retrieve all services for stack and all containers for service w/o parsing /latest/containers latest/services. 

@ibuildthecloud @vincent99 has some breaking api changes for rancher-metadata. Before "service.containers" field used to contain list of containers' names (and "stack.services" list of services' names). Now it would be collections of objects.

@cloudnautique ^^ should make your life easier :) 

https://github.com/rancher/rancher/issues/3144


cloudnautique:
Woot!

alena1108:
Update: following bugs will be fixed:

https://github.com/rancher/rancher/issues/2091

ibuildthecloud:
@alena1108 What are we going to do about API compatibility?

alena1108:
@ibuildthecloud a subject to discuss with you and @vincent99. We have to start versioning our metadata, so there are several things to discuss here:

1) do we generate one config file per version, or do we keep it in the same config file
2) would version correlate to Rancher version or rancher-metadata version. 

Unfortunately even with the proper versioning, this commit is not going to be backwards compatible as today users call rancher-metadata/latest as "latest" is the only entry point available today.

vincent99:
One file, basically take the existing file and wrap it in one more layer of map to denote the version:

```
{
  '2015-07-25': { // stuff for old version },
  '2015-12-19': { // stuff for new version }
}
```

The `2015-07-25` version has always existed and we tell people to use it instead of `latest`.  So that part is not breaking compatibility or a concern.

alena1108:
@vincent99 @ibuildthecloud the PR is ready for review, but not merge - rancher-metadata should be fixed to switch from reading .json to .yml.

 Here is the example of generated .yml file:

https://gist.github.com/alena1108/1320c4e8e725c2753991

Also do you think it would make sense to make /containers, /stacks, self/stack/services, self/service/containers to be of Map type (with name as a key) instead of List? To make search easier.

alena1108:
Fixed some bugs in yml file structure; here is the updated format:

https://gist.github.com/alena1108/bea937efd7ba153d4da3

PR #1197, opened 2015-12-17
Service ip retain

alena1108:
https://github.com/rancher/rancher/issues/2494

API changes: 

* new parameter to the service "retainIp" (boolean, false by default). If parameter is set to true, service's instances ips are going to be retained. If service instance dies, or gets upgraded, its ip address will be re-used by the replacement instance. 

Ip gets assigned to serviceSuffix object; each service suffix is assigned to service instance (instance name suffix is derived from serviceSuffix.suffix field). Example of instance name:

* foo_bar_1 where foo=stack name, bar=service name and 1=suffix

MetaData changes:

* new parameter to container metadata - "service_suffix"


alena1108:
@vincent99 UI would need to expose "retainIp" field - boolean, false by default. When set to true, re-created service instances will be using retained ip addresses.

vincent99:
UI support will be in next PR

ibuildthecloud:
@alena1108 This is a field or a label?

alena1108:
@ibuildthecloud a field

ibuildthecloud:
I hate fields.  I hate labels.  I hate options.  I just want the API to read my mind.

PR #1198, opened 2015-12-18
Vendor in viz.js v1.3.0

cjellick:
Using: https://github.com/mdaines/viz.js/releases/download/v1.3.0/viz.js


ibuildthecloud:
@cjellick What's the license of graphviz?  I think I'd rather just put the js file on the CDN and then reference a CDN url.

cjellick:
BSD 3 - https://github.com/mdaines/viz.js/blob/master/LICENSE

The file is also hosted as a github release if that is sufficient

PR #1202, opened 2015-12-21
Use super create.

hibooboo2:


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectResourceManager.java
```diff
@@ -135,22 +135,28 @@ protected Object createInternal(String type, ApiRequest request) {
...
-        List<Map<String, String>> members = (ArrayList<Map<String, String>>) project.get("members");
-        projectMemberResourceManager.setMembers(newProject, members);
-        policy.grantObjectAccess(newProject);
-
-        return newProject;
+        Object object = super.createInternal(type, apiRequest);
+        if (object instanceof Account) {
+            Account newProject = (Account) object;
+            newProject.setKind(AccountConstants.PROJECT_KIND);
+            objectManager.persist(newProject);
```
alena1108: Is there a reason why we don't call super.createInternal (AbstractObjectResourceManager.createInternal) here? Should be similar how we call super.updateInternal in update method.

hibooboo2: I do call `createinternal` On line 148.

alena1108: sorry, missed that. thought line 152 does the object create, but it just updates the kind

alena1108:
LGTM

PR #1207, opened 2015-12-22
graphite.options: fix bug with import from archaius.

kaos:
Untested, but I bet this does the trick ;)


ibuildthecloud:
LOL, I'm going to guess you're right :)

PR #1211, opened 2015-12-22
Don't default protocol to tcp for lb service

alena1108:
as empty prorotol should signal that the port is http

https://github.com/rancher/rancher/issues/3128


alena1108:
@ibuildthecloud TFs seem to be unrelated. I've re-run the tests locally, they've passed

PR #1215, opened 2015-12-23
Health check actions

alena1108:
https://github.com/rancher/rancher/issues/3076

With this change, you can define action performed on a health check failure. Default action - destroy and recreate. 2 extra actions can be defined as a field on service.launchConfig.healthCheck parameter:

* noopAction - do nothing when health check fails. Unhealthy instance won't be destroyed and won't be recreated. 
* destroyOnQuorumAction (has one extra field - "quorum", integer) - destroy/recreate unhealthy instance only when quorum of healthy instances is reached; otherwise do nothing (noop). Example: with service of scale=4 and quorum=2, unhealthy instance will get recreated only when number of healthy instances in the service is >=2

Also moved a bunch of stuff around which explains 37 "Files changed"


alena1108:
@vincent99 @ibuildthecloud could you review?

alena1108:
API changes based on offline review:

instanceHealthCheck has 2 extra parameters added (in replacement of those from the first comment):
```
"strategy": {
"type": "enum",
"transform": "",
"create": true,
"update": false,
"required": false,
"readOnCreateOnly": false,
"options": [ 3 items
"Noop",
"Recreate",
"RecreateOnQuorum",
],
"default": "Recreate",
},
```

```
"recreateOnQuorumStrategyConfig": {
"type": "recreateOnQuorumStrategyConfig",
"transform": "",
"create": true,
"update": false,
"nullable": true,
"readOnCreateOnly": false,
},
```

vincent99:
@alena1108 lowercase the first letter of the enum options?  And @ibuildthecloud wants "none" instead of "noop"

alena1108:
@vincent99 ok

PR #1227, opened 2016-01-05
Don't remove instances from inactive hosts

alena1108:
Only instances for which health check failed, should be removed.

https://github.com/rancher/rancher/issues/3155

I don't know what I was thinking when I put the original code in


cusspvz:
Thanks for fixing this up!

PR #1228, opened 2016-01-05
volume delete

cjellick:
Allows volume delete from the API, but only allows it when the volume
has no mounts and is inactive.

Sets the agentId on convoy storagePools so that the convoy-agent can
handle volume lifecycle events (like volume deleting).

Refactors the mount and volume lifecycles a bit. Simplifies mount to
just requested->active->removed. When an instance is removed, the mount
will be removed. If an instance is restored from the removed state, new
mounts will be created when the instance is started. Changes volume
lifecycle sucht that mount.create and mount.remove causes them to be
activated and deactivated.

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/instance/InstanceVolumeCleanupStrategyValidationFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/instance/InstanceVolumeCleanupStrategyValidationFilter.java
@@ -1,0 +1,42 @@
+package io.cattle.platform.iaas.api.filter.instance;
+
+import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.util.SystemLabels;
+import io.cattle.platform.iaas.api.filter.common.AbstractDefaultResourceManagerFilter;
+import io.cattle.platform.object.util.DataUtils;
+import io.github.ibuildthecloud.gdapi.exception.ClientVisibleException;
+import io.github.ibuildthecloud.gdapi.request.ApiRequest;
+import io.github.ibuildthecloud.gdapi.request.resource.ResourceManager;
+import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
+import io.github.ibuildthecloud.gdapi.validation.ValidationErrorCodes;
+
+import java.util.Map;
+
+public class InstanceVolumeCleanupStrategyValidationFilter extends AbstractDefaultResourceManagerFilter {
+
+    @Override
+    public String[] getTypes() {
+        return InstanceConstants.CONTAINER_LIKE.toArray(new String[InstanceConstants.CONTAINER_LIKE.size()]);
+    }
+
+    @Override
+    public Class<?>[] getTypeClasses() {
+        return new Class<?>[] { Instance.class };
+    }
+
+    @Override
+    public Object create(String type, ApiRequest request, ResourceManager next) {
+        Map<?, ?> labels = DataUtils.getFieldFromRequest(request, InstanceConstants.FIELD_LABELS, Map.class);
+        if (labels != null) {
+            Object l = labels.get(SystemLabels.LABEL_VOLUME_CLEANUP_STRATEGY);
+            if (l != null && !InstanceConstants.VOLUME_REMOVE_STRATEGIES.contains(l)) {
+                throw new ClientVisibleException(ResponseCodes.BAD_REQUEST, ValidationErrorCodes.INVALID_OPTION, String.format(
+                        "%s is an invalid value for the %s label.", l.toString(), SystemLabels.LABEL_VOLUME_CLEANUP_STRATEGY), null);
+            }
+        }
+        
+        return super.create(type, request, next);
+    }
+
+}
--- a/code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/lock/MountVolumeLock.java
+++ b/code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/lock/MountVolumeLock.java
@@ -1,0 +1,11 @@
+package io.cattle.platform.process.common.lock;
+
+import io.cattle.platform.lock.definition.AbstractBlockingLockDefintion;
+import io.cattle.platform.lock.definition.BlockingLockDefinition;
+
+public class MountVolumeLock extends AbstractBlockingLockDefintion implements BlockingLockDefinition {
+
+    public MountVolumeLock(Long volumeId) {
+        super("MOUNT.VOLUME." + volumeId);
+    }
+}
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
@@ -1,5 +1,6 @@
 package io.cattle.platform.process.externalevent;
 
+import static io.cattle.platform.core.model.tables.AgentTable.*;
 import static io.cattle.platform.core.model.tables.EnvironmentTable.*;
 import static io.cattle.platform.core.model.tables.ServiceTable.*;
 import static io.cattle.platform.process.externalevent.ExternalEventConstants.*;
@@ -10,6 +11,7 @@ import io.cattle.platform.core.dao.GenericResourceDao;
 import io.cattle.platform.core.dao.HostDao;
 import io.cattle.platform.core.dao.StoragePoolDao;
 import io.cattle.platform.core.dao.VolumeDao;
+import io.cattle.platform.core.model.Agent;
 import io.cattle.platform.core.model.Environment;
 import io.cattle.platform.core.model.ExternalEvent;
 import io.cattle.platform.core.model.Host;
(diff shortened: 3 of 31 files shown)


code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/instance/InstanceVolumeCleanupStrategyValidationFilter.java
```diff
@@ -0,0 +1,42 @@
...
+import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
+import io.github.ibuildthecloud.gdapi.validation.ValidationErrorCodes;
+
+import java.util.Map;
+
+public class InstanceVolumeCleanupStrategyValidationFilter extends AbstractDefaultResourceManagerFilter {
+
+    @Override
+    public String[] getTypes() {
+        return new String[] { "container", "virtualMachine" };
```
ibuildthecloud: Can you use InstanceConstants.CONTAINER_LIKE?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/volume/VolumeActionFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    @Inject
+    GenericMapDao mapDao;
+
+    @Override
+    public Resource filter(ApiRequest request, Object original, Resource converted) {
+        if (!(original instanceof Volume)) {
+            return converted;
+        }
+        Volume vol = (Volume)original;
+        List<? extends Mount> mounts = mapDao.findNonRemoved(Mount.class, Volume.class, vol.getId());
```
ibuildthecloud: The performance of this will probably be bad.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/volume/VolumeActionFilter.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public Class<?>[] getTypeClasses() {
+        return new Class<?>[] { Volume.class };
+    }
+
+    @Override
+    public String[] getTypes() {
+        return new String[] { "volume" };
```
ibuildthecloud: The string type isn't required if it is the string equivalent of the getTypeClasses().  So you can just return empty array.

code/iaas/logic-common/src/main/java/io/cattle/platform/process/common/lock/MountVolumeLock.java
```diff
@@ -0,0 +1,17 @@
+package io.cattle.platform.process.common.lock;
+
+import io.cattle.platform.lock.definition.AbstractLockDefinition;
+import io.cattle.platform.lock.definition.BlockingLockDefinition;
+
+public class MountVolumeLock extends AbstractLockDefinition implements BlockingLockDefinition {
```
ibuildthecloud: You can extend AbstractBlockingLockDefintion and then getWait() isn't required and will default to `default.lock.wait.millis`

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -161,6 +163,10 @@ public void doWithLockNoResult() {
                     spData.put(ObjectMetaDataManager.ACCOUNT_FIELD, event.getAccountId());
                     spData.put(FIELD_ZONE_ID, 1L);
 
+                    Agent agent = objectManager.findOne(Agent.class, AGENT.ACCOUNT_ID, event.getReportedAccountId(),
```
ibuildthecloud: We should adjust the agentId when the pool is found to deal with new agents taking over ownership of the pool

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePurge.java
```diff
@@ -27,12 +40,29 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+        deleteVolumes(instance, state);
 
         deallocate(instance, null);
 
         return null;
     }
+
+    private void deleteVolumes(Instance instance, ProcessState state) {
+        Object b = DataAccessor.fieldMap(instance, FIELD_LABELS).get(SystemLabels.LABEL_VOLUME_CLEANUP_STRATEGY);
+        String behavior = b != null ? b.toString() : VOLUME_CLEANUP_STRATEGY_UNNAMED;
```
ibuildthecloud: Random comment, not needed to be done: We need a util method for get label that returns a string.  I have this same code in a bunch of places.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePurge.java
```diff
@@ -27,12 +40,29 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+
+        if (VOLUME_CLEANUP_STRATEGY_NONE.equals(behavior)
+                || (!VOLUME_CLEANUP_STRATEGY_UNNAMED.equals(behavior) && !VOLUME_CLEANUP_STRATEGY_ALL.equals(behavior))) {
+            return;
+        }
+
+        Set<? extends Volume> volumes = volumeDao.findNonremovedVolumesWithNoOtherMounts(instance.getId());
```
ibuildthecloud: `YouMayHaveWonTheContestForLongestMethodNameInCattle()` :smile:

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
```diff
@@ -207,9 +206,6 @@ protected void allocate(Instance instance) {
 
     protected void storage(Instance instance, ProcessState state) {
         List<Volume> volumes = getObjectManager().children(instance, Volume.class);
-
-        volumes.addAll(InstanceHelpers.extractVolumesFromMounts(instance, objectManager));
```
ibuildthecloud: Why is this removed?  I don't even know why it was there to begin with, honestly.

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
```diff
@@ -96,4 +98,20 @@ public void createVolumeInStoragePool(Map<String, Object> volumeData, String vol
...
     }
+
+    @Override
+    public Set<? extends Volume> findNonremovedVolumesWithNoOtherMounts(long instanceId) {
+        List<VolumeRecord> vols = create()
+            .selectDistinct(VOLUME.fields())
+            .from(MOUNT)
+            .join(VOLUME)
+                .on(VOLUME.ID.eq(MOUNT.VOLUME_ID).and(VOLUME.REMOVED.isNull()))
+            .where(MOUNT.VOLUME_ID.in(
```
ibuildthecloud: We had a similar subselect in @alena1108's code (can't remember where), but it killed mysql.  It was way faster to actually do the select(), get the result and then pass the result as an are to `in`.  This approach will work fine as long as you know the data set from the first select will typically be smallish (less than 100).

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
```diff
@@ -266,7 +266,7 @@ void setLogConfig(Instance instance, Map<String, Object> fromInspect) {
     void setLabels(Instance instance, Map<String, Object> fromInspect) {
         // Labels not yet implemented in docker-java. Need to use the raw map
         Object l = CollectionUtils.getNestedValue(fromInspect, "Config", "Labels");
-        Map<String, String> cleanedLabels = new HashMap<String, String>();
+        Map<String, Object> cleanedLabels = new HashMap<String, Object>();
```
ibuildthecloud: Why this change?  Is this related to this code or some other random bug?

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
alena1108: This is the file @ibuildthecloud referenced:

https://github.com/rancher/cattle/blob/275ff45c25f9a2064af33a21a4e6137b0bcabbd9/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/service/impl/NicPerVnetNicLookup.java

I'm scarred for life by that bug.

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
cjellick: @ibuildthecloud This is related to adding the `io.rancher.container.volume_cleanup_strategy=none` for native containers [in ContainerEventCreate](https://github.com/rancher/cattle/pull/1228/files#diff-c5617822afbbc645620f4bd13d6caff8R196).

When I add the label there, when this code finally gets called as part of back population, it will clobber any existing labels that didn't come back as part of the docker inspect.

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/VolumeDaoImpl.java
cjellick: I do expect the data set to be smallish. But now I'm scared!

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
cjellick: Yeah...after seeing it and thinking about what it did, I came to the conclusion that it should not be there.

ibuildthecloud:
LGTM

PR #1237, opened 2016-01-07
Validate lb service public port against reserved port 42

alena1108:
1) As thats the port that gets configured as a a lb listener

https://github.com/rancher/rancher/issues/2701

2) Validate lb service protocol 
https://github.com/rancher/rancher/issues/667


cjellick:
LGTM. Go ahead and merge when build is done

PR #1240, opened 2016-01-08
Fixed the bug in recreate on quorum

alena1108:
Recreate only when number of healthy instances is < quroum

https://github.com/rancher/rancher/issues/3076


alena1108:
Closing as the initial implementation was correct :-/

PR #1241, opened 2016-01-08
Remove instances in initialzing state after a configurable time

alena1108:
https://github.com/rancher/rancher/issues/3229


code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/HealthcheckCleanupMonitor.java
```diff
@@ -0,0 +1,5 @@
+package io.cattle.iaas.healthcheck.service;
+
+public interface HealthcheckCleanupMonitor {
+
```
ibuildthecloud: Hmmmm.... maybe we don't need this interface.

vincent99:
Described transitions and the field naming LGTM.. code LJavaTM

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/HealthcheckCleanupMonitor.java
alena1108: Added it by following the example for PurgeMonitor...

PR #1245, opened 2016-01-12
Fixed the bug in networkfrom

alena1108:
for the case when secondary config uses network from primary launch config

https://github.com/rancher/rancher/issues/3245


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DeploymentUnitService.java
```diff
@@ -115,19 +116,29 @@ public Service getService() {
...
-                    sidekicksLaunchConfigIds.add(sidekicksLaunchConfigObj.toString());
+                    configNames.add(sidekicksLaunchConfigObj.toString());
                 }
             }
         }
-        return sidekicksLaunchConfigIds;
+
+        List<String> toReturn = new ArrayList<>();
+        for (String name : configNames) {
+            if (name.equalsIgnoreCase(service.getName())) {
```
alena1108: With sidekicks, when do networkFrom/volumesFrom from sidekick to master, we accept service name. But internally service name gets translated to ServiceDiscoveryConstants.PRIMARY_LAUNCH_CONFIG_NAME to support service rename situation. There was one place where we didn't make this conversion.

cjellick:
LGTM

PR #1246, opened 2016-01-13
Metadata: service.create_index to 2015-12-19 only

alena1108:
Excluded it from the 2015-07-25 as it wasn't supported in that version


cjellick:
LGTM

PR #1257, opened 2016-01-15
Inject idformatter

hibooboo2:


alena1108:
Even with the class cast fix, parseId always returns null:

parseId(String.valueOf(ObjectUtils.getId(resource))) = null

code/framework/auditing/src/main/java/io/cattle/platform/iaas/api/auditing/AuditServiceImpl.java
```diff
@@ -140,8 +144,10 @@ private AuditEventType convertToAction(String method) {
     @Override
     public void logResourceModification(Object resource, Map<String, Object> data, AuditEventType eventType,
                                         String description, Long accountId, String clientIp) {
-        putInAsString(data, "resource", "Failed to convert resource to json.", resource);
-        publishEvent(auditLogDao.create(objectManager.getType(resource), parseId((String) ObjectUtils.getId(resource)),
+        if (data != null) {
+            putInAsString(data, "resource", "Failed to convert resource to json.", resource);
+        }
+        publishEvent(auditLogDao.create(objectManager.getType(resource), parseId(String.valueOf(ObjectUtils.getId(resource))),
                 data, null, accountId, null, eventType.toString(), null, null, description, clientIp));
```
alena1108: EventType that we put to publishEvent method, should be constructed similar to its API counterpart: resourceType.actionType. Example: service.update. So instead of eventType.toString(), we should use objectManager.getType(resource) + "." + eventType.toString()

alena1108:
LGTM

ibuildthecloud:
@hibooboo2 You should probably use the ApiContext idformatter first and if not available fall back to the injected one.

PR #1262, opened 2016-01-16
wait for volume state in tests

cjellick:
The relationship between the lifecycles of instances, mounts, and volumes have changed a bit. Particularly, we've gone from `executing` volume state changes to `scheduling` them. So, we can no longer expect a volume's state to be changed as soon as the instance's state has changed. This PR fixes the tests that were previously relying on that assumption.

### DIFF
--- a/tests/integration/cattletest/core/test_docker.py
+++ b/tests/integration/cattletest/core/test_docker.py
@@ -497,18 +497,21 @@ def test_docker_volumes(docker_client, super_client):
             baz_mount = mount
             baz_vol = mount.volume()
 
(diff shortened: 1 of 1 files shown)


cjellick:
bonus: fixed a flake8 error

alena1108:
LGTM

cjellick:
Failures are:
* [gw2] FAILED core/test_allocation.py::test_compute_free 
* [gw1] FAILED core/test_ha.py::test_container_ha_restart 
* [gw0] FAILED core/test_svc_discovery.py::test_validate_service_scaleup_scaledown 

all known/unrelated

PR #1265, opened 2016-01-19
Add identity field to accounts.

hibooboo2:
Works to address rancher/rancher#3324

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/AccountOutputFilter.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/AccountOutputFilter.java
@@ -1,0 +1,52 @@
+package io.cattle.platform.iaas.api.auth.identity;
+
+import io.cattle.platform.api.auth.Identity;
+import io.cattle.platform.core.constants.AccountConstants;
+import io.cattle.platform.core.constants.ProjectConstants;
+import io.cattle.platform.core.model.Account;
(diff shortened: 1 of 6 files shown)


ibuildthecloud:
@hibooboo2 Can you rebase

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/AccountOutputFilter.java
```diff
@@ -0,0 +1,60 @@
...
+
+public class AccountOutputFilter implements ResourceOutputFilter {
+
+    @Inject
+    IdentityManager identityManager;
+
+
+    @Override
+    public Resource filter(ApiRequest request, Object original, Resource converted) {
+        IdFormatter idFormatter = ApiContext.getContext().getIdFormatter();
```
ibuildthecloud: Does any of this code touch the DB?  If so this means listing accounts will be pretty slow.

hibooboo2: You are right it does hit the database for each account. I could instead just create the identity without verifying it and the id i return from the api may be invalid. Take your pick.

vincent99: `n+1` requests (plus round-trips to GitHub?) doesn't sound good.  In what case would the ID be invalid?  I don't like the idea of having a link that might 404 but it may be the only practical option here.

PR #1270, opened 2016-01-19
Changes to generate a random token per Service for public dns

prachidamle:
A cryptographically secure random token is generated during Service create and exposes in the rancher-metadata so that it can be consumed by public dns provider

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,8 @@ code/packaging/app/keys.properties
 .classpath
 .project
 .settings
-
 code/packaging/app/keys.properties
+code/framework/encryption/keys.properties
 dist
 docs/build
 resources/content/db/mysql/cattle_dump*.sql
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ServiceMetaData.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ServiceMetaData.java
@@ -35,6 +35,7 @@ public class ServiceMetaData {
     protected Integer scale;
     protected String fqdn;
     protected List<String> expose = new ArrayList<>();
+    protected String token;
 
     public ServiceMetaData(ServiceMetaData that) {
         this.name = that.name;
@@ -58,6 +59,7 @@ public class ServiceMetaData {
         this.isPrimaryConfig = that.isPrimaryConfig;
         this.launchConfigName = that.launchConfigName;
         this.stackId = that.stackId;
+        this.token = that.token;
     }
 
     public ServiceMetaData(Service service, String serviceName, Environment env, List<String> sidekicks,
@@ -82,6 +84,7 @@ public class ServiceMetaData {
         this.scale = DataAccessor.fieldInteger(service, ServiceDiscoveryConstants.FIELD_SCALE);
         this.fqdn = DataAccessor.fieldString(service, ServiceDiscoveryConstants.FIELD_FQDN);
         this.stackId = env.getId();
+        this.token = DataAccessor.fieldString(service, ServiceDiscoveryConstants.FIELD_TOKEN);
     }
 
     @SuppressWarnings("unchecked")
@@ -247,5 +250,13 @@ public class ServiceMetaData {
 
     public Long getStackId() {
         return stackId;
+    }
+
+    public void setToken(String token) {
+        this.token = token;
+    }
+
+    public String getToken() {
+        return token;
     }
 }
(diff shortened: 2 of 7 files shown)


resources/content/schema/base/service.json
```diff
@@ -68,6 +68,11 @@
             "type": "array[publicEndpoint]",
             "required": false,
             "nullable": true
+        },
+        "dnsToken": {
+            "type": "string",
+            "required": true,
```
alena1108: this field shouldn't be required as we generate it internally. This setting might break service.update for the service on upgraded setups

alena1108:
We need to add some integration test for this PR ;)

ibuildthecloud:
@prachidamle I thought we decided the UUID is good enough for this.  With UUID we get 112 bits of random data which should be sufficient for this thing.

prachidamle:
@ibuildthecloud no we dint decide on uuid, I guess the service uuid is exposed in the API. That wont make it secure for token exchange right?

.gitignore
```diff
@@ -10,8 +10,9 @@ code/packaging/app/keys.properties
 .classpath
 .project
 .settings
-
+.gitignore
```
vincent99: `.gitignore` shouldn't be... gitignored...

vincent99:
Depending on the version, many of the bits in a UUID are not random.  I thought we decided to use the same function that generates API secret keys (40 "base62" characters, ≈240 bits).

ibuildthecloud:
The Java impl specifically has 112 bits of randomness.  If we want to create a random token id rather we not call this "dnsToken" then.  Just call it token.

We were going to just use the uuid because we didn't have to program anything.  Now that we've gone through the effort and did this might as well use it.  Just change the name.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -470,4 +473,20 @@ public void doWithLockNoResult() {
...
+
+    @Override
+    public void setDnsToken(Service service) {
+        String dnsToken = generateNewDnsToken();
+        DataAccessor.fields(service).withKey(ServiceDiscoveryConstants.FIELD_DNSTOKEN).set(dnsToken);
+        objectManager.persist(service);
+    }
+
+    private String generateNewDnsToken() {
+        SecureRandom random = new SecureRandom();
```
ibuildthecloud: Just use `io.cattle.platform.iaas.api.filter.apikey.ApiKeyFilter.generateKeys()[1]` instead.  That is the same logic we use for API keys.

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ServiceMetaData.java
```diff
@@ -35,6 +35,7 @@
     protected Integer scale;
     protected String fqdn;
     protected List<String> expose = new ArrayList<>();
+    protected String dns_token;
```
ibuildthecloud: Use camel case, but really this should just be `token`, so moot point.

prachidamle: Yes was using camel case, but in the answers.yml it came up in camel case instead of _ separated. So changed it to using _ just as other fields here.

But anyway, moot

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
prachidamle: io.cattle.platform.iaas.api.filter.apikey.ApiKeyFilter.generateKeys()[1]
@ibuildthecloud @vincent99 so we would be using the AccessKey then and not the secretkey?

prachidamle:
@alena1108 @ibuildthecloud @vincent99 added new changes to the PR. please review...

prachidamle:
@alena1108 any idea what is wrong for the failed test

My test has passed though 
[gw4] PASSED core/test_svc_api_validation.py::test_validate_service_token

ibuildthecloud:
@prachidamle can you rebase?

prachidamle:
@ibuildthecloud rebased, but again my test passed and something else fails
[gw3] PASSED core/test_svc_api_validation.py::test_validate_service_token

prachidamle:
This test is failing "test_docker_mount_life_cycle"
@cjellick believes its a problem with the test itself, he will be fixing sometime later
@alena1108 can you kick off another tests to see if I get lucky this time for some random reason

prachidamle:
@ibuildthecloud @alena1108 woohoo I got lucky!

alena1108:
LGTM

PR #1279, opened 2016-01-21
Changes to haproxy monit script

alena1108:
* safe haproxy reload
* cleanup invalid process entires on every reload

https://github.com/rancher/rancher/issues/3355


resources/content/config-content/haproxy/content/etc/monit/conf.d/haproxy
```diff
@@ -1,3 +1,3 @@
 check process haproxy with pidfile /var/run/haproxy.pid
-   start program = "/etc/init.d/haproxy start"
+   start program = "./var/lib/cattle/haproxy_reload.sh"
```
ibuildthecloud: Why a `.`?

PR #1283, opened 2016-01-22
LB support for TLS 1.0

untoldone:
Fix for https://github.com/rancher/rancher/issues/3097

Any reason for TLS 1.0 to be disabled in the first place assuming HAProxy has a POODLE-safe implementation?

There's still many clients that don't support TLS above 1.0.

### DIFF
--- a/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
+++ b/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
@@ -46,7 +46,7 @@ defaults
 <#if listener.sourceProtocol == "https"><#assign protocol="http"></#if>
 <#if listener.sourceProtocol == "ssl"><#assign protocol="tcp"></#if>
 frontend ${listener.uuid}_frontend
-        bind ${publicIp}:${sourcePort}<#if (listener.sourceProtocol == "https" || listener.sourceProtocol == "ssl") && certs?has_content> ssl crt /etc/haproxy/certs/<#if !defaultCert??> strict-sni</#if>  no-sslv3 no-tlsv10</#if>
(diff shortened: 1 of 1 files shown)


vincent99:
TLS 1.1+ supported by all the major non-EOL browsers/aides and 1.0 was deprecated by NIST 2 years ago.

The suggestion in that issue of putting `ssl-default-bind-options no-sslv3 no-tlsv10` in global where it can be overridden seems like a much better idea than just re-enabling by default and now having no way to re-disable.

untoldone:
At the very least, .net 4.0 standard network lib does not have support for TLS 1.1 which shipped with Windows 7 and Server 2008 R2 (which means all windows machines through Aug 2012 wont have client lib support without updates which can't really be assumed as a desktop app developer).

If its not a security risk, why opt to disable it?

That said, at this point, my work around is to have my own LB docker image that isn't the built in Rancher one. I'd be fine either way, but there would be a segment of the app population that wouldn't be supported.

vincent99:
It is not secure.  See BEAST, among probably others, and I don't know for sure but don't think there is any usable cipher + perfect forward secrecy 1.0.

But more importantly it is already currently disabled and reopening deprecated protocols is contrary to user expectations.  A PR to move where it is disabled so that it can be overridden easily by the user is fine.  There is a mechanism to add to the global and defaults sections of the config already.  Simple re-enabling what was explicitly disabled, especially leaving no way to undo it, is not.

alena1108:
I agree with the suggestion that we should set no-sslv3 no-tlsv10 in "global" section (under ssl-default-bind-options), so it can be overridden by the user based on its applications settings.

alena1108:
https://github.com/rancher/cattle/pull/1285

alena1108:
@vincent99 support for "ssl-default-bind-options" param was added in 1.5.7 version of haproxy, while ours is 1.5.4 - need to upgrade agent-instance image to 1.5.14 before merging in #1285 .

alena1108:
Closing this PR as the parameter is defined in the globals section now, and should be overridden there.

PR #1285, opened 2016-01-22
haproxy: move ssl-default-bind-options to global

alena1108:
used to be defined on per every listener

https://github.com/rancher/rancher/issues/3097


alena1108:
don't merge it yet, have to test is putting this flag in global when no certs are present, wouldn't blow up.

alena1108:
@vincent99 @ibuildthecloud support for this param was added in 1.5.7 version of haproxy, while ours is 1.5.14 - need to upgrade agent-instance image before merging this fix in.

alena1108:
haproxy 1.6 is in, so this PR is good to be merged as well

PR #1287, opened 2016-01-22
Add all filter to user preferences

hibooboo2:
rancher/rancher#3007
rancher/rancher#1814


ibuildthecloud:
@hibooboo2 Can you add a test?

PR #1290, opened 2016-01-23
DataVolume bind mount for all agent instances

alena1108:
@cloudnautique @ibuildthecloud

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
@@ -1,0 +1,52 @@
+package io.cattle.platform.process.instance;
+
+import io.cattle.platform.core.constants.AgentConstants;
+import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.util.SystemLabels;
+import io.cattle.platform.engine.handler.HandlerResult;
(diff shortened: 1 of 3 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { "instance.create" };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        Instance instance = (Instance)state.getResource();
+        if (!InstanceConstants.CONTAINER_LIKE.contains(instance.getKind())) {
```
alena1108: Not 100% sure if we need this limitation... @ibuildthecloud ^^

cloudnautique:
Could we add the -v /var/lib/rancher/etc:/var/lib/rancher/etc:ro to the custom command for registering hosts. we will be directing users to place the ssl cert in that folder.

cloudnautique:
Load balancers are able to activate now with these changes.

ibuildthecloud:
@cloudnautique Do we really need the user to put `-v /var/lib...` in the register command being that that command will just launch the actual rancher-agent container?

cloudnautique:
Yes. The first container talks to rancher server, and the first container wires up everything for the agent.

cloudnautique:
@alena1108, @ibuildthecloud is ok with the bind mount being added to the run command.

PR #1292, opened 2016-01-25
Only deactivate non-removed mounts

cjellick:


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceRemove.java
```diff
@@ -48,7 +49,7 @@ protected void storage(Instance instance, Map<String, Object> data) {
             }
         }
 
-        List<Mount> mounts = getObjectManager().children(instance, Mount.class);
+        List<Mount> mounts = getObjectManager().find(Mount.class, MOUNT.REMOVED, null, MOUNT.INSTANCE_ID, instance.getId());
         for (Mount mount : mounts) {
```
alena1108: What if mount is in inactive/removing state? Do we want to exclude them as well?

cjellick: @alena1108 per our convo, I verified that deactivating an inactive resource has no ill effect. It just completes with "already done". Also, my thinking with REMOVING was that it would eventually transition to removed and then get excluded from the query and instance.remove would complete.

cjellick:
@alena1108 I went ahead and updated my query to be more selective of the states that it picks. fwiw, it _is_ safe to do deactivate on an inactive resource, but you point was well taken.

alena1108:
LGTM

cjellick:
Addresses https://github.com/rancher/rancher/issues/3392

PR #1293, opened 2016-01-26
Launch public dns system stack from catalog when project.publicdns=true

prachidamle:
Changes to launch the latest catalog public dns template on a system stack update event when project.publicdns flag is set to true.

A system stack "publicdns" gets added

### DIFF
--- a/code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
+++ b/code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
@@ -1,6 +1,7 @@
 package io.cattle.platform.systemstack.listener;
 
 import static io.cattle.platform.core.model.tables.EnvironmentTable.*;
+import io.cattle.platform.archaius.util.ArchaiusUtil;
 import io.cattle.platform.configitem.events.ConfigUpdate;
 import io.cattle.platform.configitem.model.Client;
 import io.cattle.platform.configitem.model.ItemVersion;
@@ -14,6 +15,7 @@ import io.cattle.platform.eventing.EventService;
 import io.cattle.platform.eventing.annotation.AnnotatedEventListener;
 import io.cattle.platform.eventing.annotation.EventHandler;
 import io.cattle.platform.eventing.model.EventVO;
+import io.cattle.platform.json.JsonMapper;
 import io.cattle.platform.lock.LockCallbackNoReturn;
 import io.cattle.platform.lock.LockManager;
 import io.cattle.platform.object.ObjectManager;
@@ -21,6 +23,7 @@ import io.cattle.platform.object.process.ObjectProcessManager;
 import io.cattle.platform.object.process.StandardProcess;
 import io.cattle.platform.object.util.DataAccessor;
 import io.cattle.platform.servicediscovery.api.constants.ServiceDiscoveryConstants;
+import io.cattle.platform.systemstack.listener.external.api.catalog.Template;
 import io.cattle.platform.systemstack.lock.SystemStackLock;
 import io.cattle.platform.systemstack.process.SystemStackTrigger;
 import io.cattle.platform.util.type.CollectionUtils;
@@ -34,15 +37,19 @@ import javax.inject.Inject;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.http.client.fluent.Request;
+import org.apache.http.client.fluent.Response;
+import org.apache.http.conn.HttpHostConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class SystemStackUpdate implements AnnotatedEventListener {
 
-    private static final String[] STACKS = new String[] {"swarm", "kubernetes"};
+    private static final String[] STACKS = new String[] {"swarm", "kubernetes", "publicDns"};
     private static final String STACK_EXTERNAL_ID = "system://%s";
     private static final String STACK_RESOURCE = "/config-content/system-stacks/%s/%s";
     private static final Logger log = LoggerFactory.getLogger(SystemStackUpdate.class);
+    private static final String CATALOG_RESOURCE_URL = ArchaiusUtil.getString("system.stack.catalog.url").get();
 
     @Inject
     ObjectProcessManager processManager;
@@ -64,6 +71,9 @@ public class SystemStackUpdate implements AnnotatedEventListener {
 
     @Inject
     GenericResourceDao resourceDao;
+
+    @Inject
+    JsonMapper jsonMapper;
 
     @EventHandler
     public void globalServiceUpdate(ConfigUpdate update) {
@@ -115,8 +125,22 @@ public class SystemStackUpdate implements AnnotatedEventListener {
             return true;
         }
 
-        String compose = getFile(stack, "docker-compose.yml");
-        String rancherCompose = getFile(stack, "rancher-compose.yml");
+        String compose, rancherCompose;
+
+        compose = getFile(stack, "docker-compose.yml");
+
+        if (compose == null) {
+            //not found in classpath, fetch from catalog
+            Template catalogEntry = getTemplateFromCatalog(stack);
+            if(catalogEntry == null) {
+                log.error("Failed to load template from catalog for stack [{}]", stack);
+                return true;
+            }
+            compose = catalogEntry.getDockerCompose();
+            rancherCompose = catalogEntry.getRancherCompose();
+        } else {
+            rancherCompose = getFile(stack, "rancher-compose.yml");
+        }
 
         if (compose == null) {
             log.error("Failed to find compose file for stack [{}]", stack);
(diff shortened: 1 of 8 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -470,4 +471,11 @@ public void doWithLockNoResult() {
             }
         });
     }
+
+    @Override
+    public void setToken(Service service) {
+        String token = ApiKeyFilter.generateKeys()[1];
+        DataAccessor.fields(service).withKey(ServiceDiscoveryConstants.FIELD_TOKEN).set(token);
+        objectManager.persist(service);
```
ibuildthecloud: Can you use objectManager.setFields() instead.  In general avoid persisting the whole object and just set the fields you need.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
```diff
@@ -34,15 +36,18 @@
...
 import org.apache.commons.lang3.StringUtils;
+import org.apache.http.client.fluent.Request;
+import org.apache.http.client.fluent.Response;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class SystemStackUpdate implements AnnotatedEventListener {
 
-    private static final String[] STACKS = new String[] {"swarm", "kubernetes"};
+    private static final String[] STACKS = new String[] {"swarm", "kubernetes", "publicdns"};
```
ibuildthecloud: Across the board can you switch to "publicDns"

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
```diff
@@ -137,13 +145,44 @@ protected boolean addStack(Account account, String stack) throws IOException {
...
+                        return templateVersion.getRancherCompose();
+                    }
+                }
+                log.error("Failed to load template version");
+            } else {
+                log.error("Failed to load template details");
+            }
             return null;
+        } else {
+            String resource = String.format(STACK_RESOURCE, stack, filename);
```
ibuildthecloud: We don't need fallback logic.  We will just always go to the catalog now.  The existing swarm, k8s functionality is not currently enabled so you don't need to worry about breaking things.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
```diff
@@ -34,15 +36,18 @@
...
 import org.slf4j.LoggerFactory;
 
 public class SystemStackUpdate implements AnnotatedEventListener {
 
-    private static final String[] STACKS = new String[] {"swarm", "kubernetes"};
+    private static final String[] STACKS = new String[] {"swarm", "kubernetes", "publicdns"};
     private static final String STACK_EXTERNAL_ID = "system://%s";
     private static final String STACK_RESOURCE = "/config-content/system-stacks/%s/%s";
     private static final Logger log = LoggerFactory.getLogger(SystemStackUpdate.class);
+    private static final String CATALOG_RESOURCE_URL = "http://localhost:8088/v1-catalog/templates/library:%s";
```
ibuildthecloud: We should probably make this a setting even though the web.xml hard codes it.  I'd call the setting something like `system.stack.catalog.url`.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
```diff
@@ -137,13 +145,44 @@ protected boolean addStack(Account account, String stack) throws IOException {
...
+                Template templateVersion = jsonMapper.readValue(versionRes.returnContent().asBytes(), Template.class);
+
+                if(templateVersion != null) {
+                    if(filename!= null && filename.startsWith("docker-")){
+                        return templateVersion.getDockerCompose();
+                    }else if(filename!= null && filename.startsWith("rancher-")){
+                        return templateVersion.getRancherCompose();
+                    }
+                }
+                log.error("Failed to load template version");
```
ibuildthecloud: I would not log error here, if a catalog does not have a `rancher-compose.yml`, that is not necessarily an error I would say.  Only `docker-compose.yml` is really required.  But I'd put that error handling logic in the caller of this function, not the function itself.  Just return null.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
```diff
@@ -137,13 +145,44 @@ protected boolean addStack(Account account, String stack) throws IOException {
     }
 
     protected String getFile(String stack, String filename) throws IOException {
```
ibuildthecloud: I would change the response type of this method to just `Template`.  No need to make two API calls to `rancher-catalog-service` for the same thing.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/external/api/catalog/Template.java
```diff
@@ -0,0 +1,376 @@
+package io.cattle.platform.systemstack.listener.external.api.catalog;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
```
ibuildthecloud: In general avoid jackson annotations if possible.  Here, none of them are required, so just remove all the `@JsonProperty`s

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/external/api/catalog/Template.java
```diff
@@ -0,0 +1,376 @@
...
+
+public class Template {
+    @JsonProperty("catalogId")
+    private String catalogId;
```
ibuildthecloud: There's no particular reason to make all the variables as private.  I would just leave them as the default visibility.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/external/api/catalog/Template.java
```diff
@@ -0,0 +1,376 @@
...
+    * 
+    * @return
+    * The catalogId
+    */
+    @JsonProperty("catalogId")
+    public String getCatalogId() {
+    return catalogId;
+    }
+
+    /**
```
ibuildthecloud: Remove all the javadocs for getters/setters.  They don't provide any value unless they have some meaningful info in them.

resources/content/schema/user/user-auth.json
```diff
@@ -264,6 +264,7 @@
         "project.members" : "cr",
         "project.swarm" : "cru",
         "project.kubernetes" : "cru",
+        "project.publicdns" : "cru",
```
ibuildthecloud: There needs to be a change somewhere in `test_authorization.py` to test this new field.

tests/integration/cattletest/core/test_projects.py
```diff
@@ -552,3 +552,17 @@ def _create_members(user_clients, members):
             'externalIdType': 'rancher_id'
         })
     return newMembers
+
+
+def test_update_project_publicdns(user_clients, project):
+    project = user_clients['Owner'].update(
+        project, name='Project Name', description='Some description',
+        publicdns=True)
+    assert project.publicdns is True
```
ibuildthecloud: `is True` is redundant.

ibuildthecloud:
@prachidamle In general I would not combine PRs like this.  The changes while both are for the same larger feature are completely independent and can be reviewed and merged independently.

resources/content/schema/user/user-auth.json
ibuildthecloud: I believe this is why the build is currently red.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
prachidamle: @ibuildthecloud I tried using objectManager.setFields(), but it does not set the token in the db.
objectManager.setFields(service, ServiceDiscoveryConstants.FIELD_TOKEN, token);
return objectManager.reload(service);

but this does not set the token in the db inside data[]
If I define the field in service.json and invoke the above, I get "Database record has changed" exception

prachidamle:
@ibuildthecloud the test failure is for  test_system_stack.py
this test expects that we load the compose files for swarm the old way - from classpath and not from catalog.

So either we need to fall back on classpath or just comment out this test until catalog is updated to have images for publicdns/swarm/kubernetes etc.

what do you suggest?

ibuildthecloud:
@prachidamle  id rather not run the catalog service in the integration tests.  I think the fallback logic should be added back but in a slightly different way.  For all catalog items check the classpath first, if not found, then check the catalog.

That should make the previous tests still work, is there anything in the new tests that require the catalog?

prachidamle:
@ibuildthecloud I did the changes to check the classpath first for compose files. If not found go to catalog. If catalog is not running, catch the error and just return null compose files.
This should make the previous tests work. New tests don't go to catalog yet.

prachidamle:
@ibuildthecloud @alena1108 please review, the changes are green

PR #1295, opened 2016-01-26
Added bindmount for conf/state dir to docker command.

cloudnautique:
In some circumstances, like adding a self signed CA pem, the first
instance of Rancher agent image needs to have access to the config
in the state directory to startup.


cloudnautique:
addresses my comments about the docker command in #1290

PR #1301, opened 2016-01-27
wait for mounts in docker tests

cjellick:
These TFs occurred because mount activation/deactivation was changed from an executed process to a scheduled on. The tests previously relied on the behavior of the executing being done before the parent container came to a resting state. Now that those mount processes are scheduled, I need to wait explicitly for the mounts to appear/disappear


ibuildthecloud:
@cjellick Seems like this failure may be legit?

cjellick:
whoops! latest nginx image dropped the built-in volume that i was relying upon. thats what i get for using the latest tag

PR #1311, opened 2016-01-28
Adds library name to Rancher catalog and add community.

cloudnautique:
Community catalog will be the location that our community members can
easily add templates.

This will set the defaults for new installations. The UI will make it easy to enable the community catalog. 

Having a separate catalog will allow organizations to enable / disable contributed templates on their infrastructure.


alena1108:
LGTM

PR #1322, opened 2016-01-30
Could fix three web.xml in cattle, and change org.mortbay.jetty to org.eclipse.jetty which is more official in maven repo.

kingsmiler:
1) change default db to mysql and comment others;
2) change org.mortbay.jetty to org.eclipse.jetty in code/packaging/app-parent/pom.xml;
3) add mysql dependency to code/packaging/app/pom.xml;
4) change jetty version and mysql version;
5) correct all web.xml's web-app header.


.gitignore
```diff
@@ -2,10 +2,9 @@
 /.gitconfig
 
 /.cache
-/.idea
+.idea
 /.m2
 
-code/packaging/app/keys.properties
```
ibuildthecloud: Is there a specific reason you removed this file?  That file gets generated a lot of the times when you launch the application through Eclipse and developers kept putting it in PRs, so we ignore it.

code/meta-parent/pom.xml
```diff
@@ -47,7 +47,7 @@
         </developer>
     </developers>
     <properties>
-        <jetty.version>9.2.11.v20150529</jetty.version>
+        <jetty.version>9.2.14.v20151106</jetty.version>
```
ibuildthecloud: @cjellick Is this a safe upgrade?  I know you had a lot of issues w/ Jetty and WebSockets.

code/meta-parent/pom.xml
```diff
@@ -290,7 +290,7 @@
             <dependency>
                 <groupId>mysql</groupId>
                 <artifactId>mysql-connector-java</artifactId>
-                <version>5.1.26</version>
```
ibuildthecloud: We don't actually use the MySQL driver.  Feel free to remove this dependency management reference.  We actually use the MariaDB driver.  This is because MariaDB is LGPL and MySQL is GPL.  We can't use the GPL driver for many legal reasons.

code/packaging/app-config/pom.xml
```diff
@@ -1,6 +1,28 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <artifactId>cattle-app-config</artifactId>
+    <dependencies>
```
ibuildthecloud: Out of curiosity, what does this solve by putting these here?  Does this help with IDE auto-completion?  This project is just config files, so there obviously is no real code dependencies.

code/packaging/app-parent/pom.xml
```diff
@@ -97,7 +97,7 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.mortbay.jetty</groupId>
+                <groupId>org.eclipse.jetty</groupId>
```
ibuildthecloud: :+1:

code/packaging/app/pom.xml
```diff
@@ -8,4 +8,11 @@
         <relativePath>../app-parent/pom.xml</relativePath>
     </parent>
     <packaging>war</packaging>
+
+    <dependencies>
```
ibuildthecloud: As mentioned before we don't use the MySQL driver but instead the MariaDB one.  Besides that why did you need to add a dependency here?

resources/content/cattle-global.properties
```diff
@@ -37,7 +37,7 @@ cattle.home=
 process.retry_max_wait.millis=120000
 
 # Choose mysql, hsqldb, or h2
-db.cattle.database=h2
+db.cattle.database=mysql
```
ibuildthecloud: The design of cattle is such that you can download the `cattle.jar` and just run it with zero dependencies.  We have other components that make this assumption.  As such we can't default to `mysql` and `h2` must be the default.  When running cattle in development just add `-Ddb.cattle.database=mysql` to the Java command line.

code/meta-parent/pom.xml
cjellick: I _think_ it _should_ be safe. The issue was that jetty 9.3 made java 8 a hard requirement.

deniseschannon:
https://github.com/rancher/rancher/issues/3870

alena1108:

PR #1331, opened 2016-02-01
Cleanup process execution table.

hibooboo2:
rancher/rancher#3136

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanUp.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanUp.java
@@ -4,6 +4,7 @@ import static io.cattle.platform.core.model.tables.AgentTable.AGENT;
 import static io.cattle.platform.core.model.tables.ContainerEventTable.CONTAINER_EVENT;
 import static io.cattle.platform.core.model.tables.ConfigItemStatusTable.CONFIG_ITEM_STATUS;
 import static io.cattle.platform.core.model.tables.ProcessInstanceTable.PROCESS_INSTANCE;
+import static io.cattle.platform.core.model.tables.ProcessExecutionTable.PROCESS_EXECUTION;
 import static io.cattle.platform.core.model.tables.ServiceEventTable.SERVICE_EVENT;
 import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
 import static io.cattle.platform.core.model.tables.AuditLogTable.AUDIT_LOG;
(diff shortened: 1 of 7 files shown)


deniseschannon:
https://github.com/rancher/rancher/issues/2381

hibooboo2:
@deniseschannon Thanks was gonna ref that issue.

code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanUp.java
```diff
@@ -37,6 +39,19 @@ public void cleanUp(){
                 .execute();
 
         long endProcessInstance = System.currentTimeMillis();
+
+        long startProcessExecution = System.currentTimeMillis();
+        long lowestProcessInstance = create()
```
ibuildthecloud: @hibooboo2 Can you explain the logic here?  Why do we consider the lowestProcessInstance, what are we trying to accomplish?

hibooboo2: Only way to determine which process executions to remove. as we don't want to delete the ones where there process instances are still alive.

ibuildthecloud: @hibooboo2 process_executions are already deleted as part of a cascading deleting from process_instance.  So the PR effectively does nothing new.  I think what we want is just to delete any process_execution record that is older than a day, just like we delete process_instance.

Because of the cascading delete we should delete process_instance first and then look for and delete process_execution.

ibuildthecloud:
@deniseschannon If their are a lot of process_execution records associated to one process_instance (this happens when a process is stuck for like a month), this will cause an OOM when the UI goes to view the process.  We should really get this fix in once it is proper.

code/packaging/app-config/src/main/resources/META-INF/cattle/process/spring-core-processes-context.xml
```diff
@@ -31,6 +31,7 @@
 
 
     <process:defaultProcesses resourceType="account" />
+    <process:defaultProcesses resourceType="processExecution" />
```
ibuildthecloud: We don't need orchestration on this type.

code/packaging/app-config/src/main/resources/META-INF/cattle/api-server/config-item-defaults.properties
```diff
@@ -41,4 +41,4 @@ default.services.domain=rancher.io
 
 item.context.healthcheck.info.items=healthcheck
 
-config.item.name.preload=reconcile
\ No newline at end of file
+config.item.name.preload=reconcile
```
ibuildthecloud: No need for this change

PR #1334, opened 2016-02-02
Removed appCookie stickiness support

alena1108:
As haproxy 1.6 no longer supports it. Should use lbCookie (create new
cookie) option instead

https://github.com/rancher/rancher/issues/3489


ibuildthecloud:
LGTM

PR #1338, opened 2016-02-02
Add cookie with value of pl setting.

hibooboo2:
This is used by ui to display generic or not ui.

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
@@ -20,6 +20,7 @@ import javax.servlet.FilterConfig;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
+import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
Is it possible to put this in a request handler?   It just seems dirty to put it here.  It's not core to the framework.

code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
```diff
@@ -177,4 +183,27 @@ public WrappedException(Throwable cause) {
...
+        if (httpRequest.getCookies() != null) {
+            for (Cookie c : httpRequest.getCookies()) {
+                if (PL.equals(c.getName()) && c.getName() != null) {
+                    plCookie = c;
+                    break;
+                }
+            }
+        }
+
+        if (plCookie == null) {
```
ibuildthecloud: Can this just be any or statement and not duplicate the blocks.

ibuildthecloud:
@hibooboo2 Nevermind about the request handler comment, you probably need to keep the code here.

code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
hibooboo2: yes

hibooboo2:
Tried to put it in handler needs to be here before the ui.

vincent99:
Results LGTM, Java LJTM

PR #1339, opened 2016-02-02
Only remap on event.

hibooboo2:
rancher/rancher#3131


ibuildthecloud:
@hibooboo2 Please add test

PR #1340, opened 2016-02-02
[WIP Don't review] Remove parsing of docker image

hibooboo2:
rancher/rancher#3122


ibuildthecloud:
@hibooboo2 While we are at it, can we get imageUuid=ubuntu to work (also support imageUuid=docker:ubuntu for backwards compatibiliy).   And then at some point we will drop the "docker:" completely.

PR #1355, opened 2016-02-11
Read lb service health check port from config

alena1108:
instead of hardcoding it.

https://github.com/rancher/rancher/issues/3375

### DIFF
--- a/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
+++ b/resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
@@ -97,6 +97,7 @@ backend ${listener.uuid}_${backend.uuid}_backend
 </#list>
 </#if>
 
-listen default 
-	bind 0.0.0.0:42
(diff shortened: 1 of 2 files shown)


resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
```diff
@@ -97,6 +97,7 @@ backend ${listener.uuid}_${backend.uuid}_backend
 </#list>
 </#if>
 
-listen default 
+listen default
+<#if lbHealthCheck??><#assign defaultPort = lbHealthCheck.port><#else><#assign defaultPort = 42></#if>
```
ibuildthecloud: I don't understand what your doing here?  You assign defaultPort but don't seem to use it.

alena1108: lol, I just forgot to remove the hardcoded value :) will fix

PR #1359, opened 2016-02-16
Bug fix in service reconcile transitioning message builder

alena1108:
@ibuildthecloud fixed that exception you saw often in the log (the bug was in the way we build the transitioningMessage for service reconcile). With the fix, the message would look like:

```
"Object [instance:1i15] failed to satisfy predicate [15000] millis. Instance status: Scheduling failed: No candidates available",
```


ibuildthecloud:
@alena1108 Do you know what these failures are?

PR #1360, opened 2016-02-16
Vendoring of GMS drivers with cattle.

hibooboo2:


code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/launch/MachineDriverLoader.java
```diff
@@ -0,0 +1,130 @@
...
+        CONFIG_FILE.addCallback(new Runnable() {
+            @Override
+            public void run() {
+                loadDrivers();
+            }
+        });
+        loadDrivers();
+    }
+
+    public void loadDrivers() {
```
ibuildthecloud: This should be synchronized

code/packaging/app-config/src/main/resources/META-INF/cattle/process/machine-defaults.properties
```diff
@@ -1,6 +1,8 @@
 machine.service.executable=go-machine-service
 machine.execute=false
 
+machine.driver.config=file:///var/lib/cattle/machinedrivers.json
```
ibuildthecloud: Change to `file://var/lib/rancher/etc/machinedrivers.json`  @cloudnautique Does that path sound good?  Does it align with the config you did for SSL?

code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/launch/MachineDriverLoader.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public void start() {
+        // do logic once
+        CONFIG_FILE.addCallback(new Runnable() {
+            @Override
+            public void run() {
+                loadDrivers();
+            }
+        });
+        loadDrivers();
```
ibuildthecloud: `loadDrivers` should really throw an exception if something fails.  On first start this will fail the start of rancher server.  But on reload (`addCallback`) the exception should be caught and ignored.

PR #1367, opened 2016-02-16
Internal dns changes

alena1108:
* services discoverable cross stacks by default
* support for k8s namespace
* setVip field on the service (boolean, false by default). Only when set to true, vip gets automatically allocated
* build dns json content in java code, so its easier to switch to yml later

Depends on the following host-api PR: https://github.com/rancher/host-api/pull/56

Also removed VIP provider code


alena1108:
Corresponding host-api PR: https://github.com/rancher/host-api/pull/58

PR #1384, opened 2016-02-25
host-api v0.26.0

cjellick:


cjellick:
Cant be merged bc of the dns fqdn issue. not sure if fixing that means a new hsot-api release, so leaving open

PR #1385, opened 2016-02-25
put non-fqdn self stack records to client sections

alena1108:
@ibuildthecloud @vincent99 

https://github.com/rancher/rancher/issues/3694

to support clients that do dns name search by always appending "." at the end of the dns-name-to-resolve, so configured search domains are not being used.


alena1108:
Closing as changes are gonna be done on internal dns side

PR #1392, opened 2016-02-27
If vm is in a service, use svc uuid in volume name

cjellick:
This allows for reuse of volumes if a vm goes down because it fails
healthchecks.

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
@@ -1,18 +1,19 @@
 package io.cattle.platform.configitem.context.dao.impl;
 
-import static io.cattle.platform.core.model.tables.HostIpAddressMapTable.HOST_IP_ADDRESS_MAP;
-import static io.cattle.platform.core.model.tables.HostTable.HOST;
-import static io.cattle.platform.core.model.tables.InstanceHostMapTable.INSTANCE_HOST_MAP;
-import static io.cattle.platform.core.model.tables.InstanceTable.INSTANCE;
-import static io.cattle.platform.core.model.tables.IpAddressNicMapTable.IP_ADDRESS_NIC_MAP;
-import static io.cattle.platform.core.model.tables.IpAddressTable.IP_ADDRESS;
-import static io.cattle.platform.core.model.tables.NicTable.NIC;
-import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.SERVICE_EXPOSE_MAP;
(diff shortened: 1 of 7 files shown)


code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
```diff
@@ -1,18 +1,19 @@
...
-import static io.cattle.platform.core.model.tables.ServiceIndexTable.SERVICE_INDEX;
+import static io.cattle.platform.core.model.tables.HostIpAddressMapTable.*;
+import static io.cattle.platform.core.model.tables.HostTable.*;
+import static io.cattle.platform.core.model.tables.InstanceHostMapTable.*;
+import static io.cattle.platform.core.model.tables.InstanceTable.*;
+import static io.cattle.platform.core.model.tables.IpAddressNicMapTable.*;
+import static io.cattle.platform.core.model.tables.IpAddressTable.*;
+import static io.cattle.platform.core.model.tables.NicTable.*;
+import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.*;
+import static io.cattle.platform.core.model.tables.ServiceIndexTable.*;
```
cjellick: blame cmd+shift+O

ibuildthecloud: I like `.*` in this situation.

code/implementation/vm/src/main/java/io/cattle/platform/vm/process/VirtualMachinePreCreate.java
```diff
@@ -98,10 +102,20 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
         Object objectDisks = DataAccessor.field(instance, InstanceConstants.FIELD_DISKS, Object.class);
         if (objectDisks instanceof List<?>) {
             String namePrefix = instance.getName();
+            Long svcIndexId = DataAccessor.fieldLong(instance, InstanceConstants.FIELD_SERVICE_INSTANCE_SERVICE_INDEX_ID);
+
+            String uuidPart = null;
+            if (svcIndexId != null) {
+                Service svc = serviceDao.getServiceByServiceIndexId(svcIndexId);
+                uuidPart = svc.getUuid().substring(0, 7);
```
ibuildthecloud: Can we add the service index to that volume name.  So if I scale the service to more than 1 I don't end up trying to use the same volume.

ibuildthecloud: Nevermind, the id is already there in the container name.

ibuildthecloud:
LGTM

PR #1395, opened 2016-02-28
Automatically retain ip for vm service.

cjellick:
If a vm is created as part of a service for VM HA reasons, we should automatically retain the IP.

This does not address a scenario where a VM is added as a sidekick to a service, but I don't think we should assume retain IP in that scenario. In that scenario, user should specify it explicitly themselves.


ibuildthecloud:
LGTM

PR #1404, opened 2016-03-03
Default secondary launch config mode to "container"

alena1108:
if networkMode parameter is not passed to the API


alena1108:
https://github.com/rancher/rancher/issues/3760

PR #1405, opened 2016-03-04
Recalculate health checks on host removal

alena1108:
https://github.com/rancher/rancher/issues/3763

@cjellick could you please review?

### DIFF
--- a/code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/HealthCheckReconcile.java
+++ b/code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/HealthCheckReconcile.java
@@ -1,0 +1,39 @@
+package io.cattle.iaas.healthcheck.process;
+
+import io.cattle.iaas.healthcheck.service.HealthcheckService;
+import io.cattle.platform.core.constants.HealthcheckConstants;
+import io.cattle.platform.core.model.HealthcheckInstanceHostMap;
+import io.cattle.platform.engine.handler.HandlerResult;
+import io.cattle.platform.engine.handler.ProcessPostListener;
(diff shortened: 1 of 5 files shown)


code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/HealthCheckReconcile.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public String[] getProcessNames() {
+        return new String[] { "healthcheckinstancehostmap.remove" };
+    }
+
+    @Override
+    public HandlerResult handle(ProcessState state, ProcessInstance process) {
+        HealthcheckInstanceHostMap hcihm = (HealthcheckInstanceHostMap) state.getResource();
+        hcSvc.healthCheckReconcile(hcihm, HealthcheckConstants.HEALTH_STATE_UNHEALTHY);
```
cjellick: Want to make sure I understand this logic:
If the healthcheckinstancehostmap is removed, you consider that as an "unhealthy" state.
Under what conditions does the healthcheckinstancehostmap get removed?

alena1108: @cjellick when the host gets removed/disconnected. So when it happens, I set the reported state as unhealthy. Then in underlying code, we check what non-removed hosts are saying. If they all say unhealthy - then host is considered as unhealthy. If one of them says healthy - host is considered as healthy.

I've picked unhealthy event over healthy because a single "healthy" event brings host to healthy even if the remaining hosts say its unhealthy. While "unhealthy" doesn't have this power.

cjellick:
LGTM

PR #1407, opened 2016-03-04
Scale policy

alena1108:
1) Scale policy parameter was added to the service object (updatable). Fields:

* minScale - minimum number of instances required to be running for the service
* maxScale - maximum allowed service scale. 
* increment - reflects a number of containers being deployed simultaneously as we move from minScale to scale.

Example: 

1) service created with scale=4, and policy {minScale:2, maxScale:5, increment:2}.
2) Durning activate, we try to deploy 2 (=minScale) containers. If it fails, service activation fails as we don't match minScale. If succeeds, move to 3)
3) We increment scale by 2 (=interval), and wait for these instances to deploy. If at least one of them fails to deploy, we roll back to 2, and reconcile finishes with currentScale=2. If deployment of 2 extra instances is successful, reconcile finishes with currentScale=scale=4

2) New field for the service (readable only) - "currentScale" - number of service's instances.

### DIFF
--- a/code/framework/object/src/main/java/io/cattle/platform/object/purge/impl/RemoveMonitorImpl.java
+++ b/code/framework/object/src/main/java/io/cattle/platform/object/purge/impl/RemoveMonitorImpl.java
@@ -19,8 +19,10 @@ import io.github.ibuildthecloud.gdapi.factory.SchemaFactory;
 import io.github.ibuildthecloud.gdapi.model.Schema;
 
 import java.util.Date;
+import java.util.HashMap;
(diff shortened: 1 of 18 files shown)


alena1108:
UPDATE: scale and scalePolicy parameters can no longer operate together. If scalePolicy is specified, scale will be ignored. This differs from the original design. Treating scalePolicy as a global scale and not let the user choose might be easier for round1 implementation. 

@vincent99 @ibuildthecloud ^^

alena1108:
API update: minScale/maxScale are replaced with min/max

PR #1424, opened 2016-03-08
Update metadata on account rename

alena1108:
https://github.com/rancher/rancher/issues/3742


alena1108:
Some test fixes as well - re-defined conflicting ports in lb test cases

PR #1425, opened 2016-03-08
Changes to add the missing Log driver/log opts to exported .yml file

prachidamle:
This will add the missing log_config details to the dockerCompose file while viewing the config back on UI.

https://github.com/rancher/rancher/issues/2388


prachidamle:
@alena1108 @ibuildthecloud I have tested is manually by verifying that the log_driver and log_opts are set in the exported config on the UI.

alena1108:
LGTM

PR #1426, opened 2016-03-08
Enable update operation on certificate.key

alena1108:
Made key a writable field. https://github.com/rancher/rancher/issues/3553

@ibuildthecloud tested is manually by verifying that the key was set in the DB as the key has ro permission, and returned to the API only on create.


alena1108:
Also added some test fixes

PR #1429, opened 2016-03-08
Suppress service reconcile exception

alena1108:
@ibuildthecloud

### DIFF
--- a/code/framework/utils/src/main/java/io/cattle/platform/util/exception/LoggableException.java
+++ b/code/framework/utils/src/main/java/io/cattle/platform/util/exception/LoggableException.java
@@ -1,0 +1,6 @@
+package io.cattle.platform.util.exception;
+
+
+public interface LoggableException {
+    void log();
+}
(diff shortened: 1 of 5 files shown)


code/framework/utils/src/main/java/io/cattle/platform/util/exception/LoggableException.java
```diff
@@ -0,0 +1,7 @@
+package io.cattle.platform.util.exception;
+
+import org.slf4j.Logger;
+
+public interface LoggableException {
+    void log(Logger log);
```
ibuildthecloud: Nah, this is too dirty to have slf4j in our interface.

alena1108: Should I just provide 2 methods - logLevel(INFO, DEBUG, etc), and logStackTrace (boolean) - instead in this interface?

ibuildthecloud: I think you can just have a `log()` message and the exception is responsible for logging itself with it's own logger.  Being that we use this in one place in NoExceptionRunnable, it doesn't seem like the logger we pass in is going to greatly enhance the information in the log

alena1108:
https://github.com/rancher/rancher/issues/3946

PR #1431, opened 2016-03-09
In case of a container with "host" net mode, set the primary_ip as the host_ip

prachidamle:
In case of a container with "host" net mode, set the primary_ip as the host_ip and not the agent_ip

https://github.com/rancher/rancher/issues/2750


prachidamle:
@alena1108 @ibuildthecloud can you please review and merge. No tests since this is a change in metadata.

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
```diff
@@ -30,17 +31,34 @@
...
+    public List<ContainerMetaData> getContainersData(final long accountId) {
         MultiRecordMapper<ContainerMetaData> mapper = new MultiRecordMapper<ContainerMetaData>() {
             @Override
             protected ContainerMetaData map(List<Object> input) {
                 ContainerMetaData data = new ContainerMetaData();
-                data.setIp((IpAddress) input.get(3));
+
+                Instance instance = (Instance) input.get(1);
+
+                final Map<Long, IpAddress> instanceIdToHostIpMap = dnsInfoDao
```
alena1108: @prachidamle , instanceIdToHostIpMap map should be fetched just once. If we place it here, it would be fetched every time we transform map to object, basically for every record. Move it up to the very beginning of the method.

prachidamle:
@alena1108 changed the code to grab the map once.

alena1108:
LGTM

PR #1435, opened 2016-03-09
Fix evaluation of io.rancher.scheduler.global value

janeczku:
This should fix https://github.com/rancher/rancher/issues/3881


alena1108:
LGTM based on code review and green tests :)

PR #1443, opened 2016-03-09
Log certificate parse errors in info

alena1108:
and no stack trace

@ibuildthecloud


alena1108:
Also fixed the TF in restart_stack

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/ssl/CertificateCreateValidationFilter.java
```diff
@@ -11,11 +11,11 @@
 import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
 import io.github.ibuildthecloud.gdapi.validation.ValidationErrorCodes;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
```
ibuildthecloud: YIKES!!.  We should a do a quick grep and see if commons logger is used anywhere else.

PR #1446, opened 2016-03-09
Admin accounts now recycled with localauth and first admin used if account not found for all auth types when turning auth on.

hibooboo2:
rancher/rancher#3780
rancher/rancher#2967
rancher/rancher#3892

### DIFF
--- a/resources/content/schema/base/credential.json
+++ b/resources/content/schema/base/credential.json
@@ -1,0 +1,7 @@
+{
+    "resourceFields" : {
+        "secretValue": {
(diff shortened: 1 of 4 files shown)


resources/content/schema/base/credential.json
```diff
@@ -0,0 +1,7 @@
+{
+    "resourceFields" : {
+        "secretValue": {
+            "type": "password"
```
ibuildthecloud: What is the impact of "password"?  Basically what does this change enable?

PR #1447, opened 2016-03-09
Update publicEndpoints on agent ip change and ports update support for standalone contaienrs

alena1108:
https://github.com/rancher/rancher/issues/3883

no test as agent ip is being updated via ping


alena1108:
Also added support for public point updates for standalone containers https://github.com/rancher/rancher/issues/1628

PR #1455, opened 2016-03-10
Fix volume create race condition

cjellick:
There are two ways volumes can be backpopulated into Rancher: from a
volume create event being sent by convoy-agent or from container
backpopulation. These two methods were not using the same lock to check
if the volume existed and thus could run into a race condition where
both created the volume.

Addresses https://github.com/rancher/rancher/issues/3446

I couldn't add a test for this because reproducing relies on hitting a race
condition that would be too hard to hit reliably. It also relies on an
active convoy stack, which we don't have or want in the cattle tests.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
@@ -19,6 +19,7 @@ import io.cattle.platform.core.model.Service;
 import io.cattle.platform.core.model.StoragePool;
 import io.cattle.platform.core.model.StoragePoolHostMap;
 import io.cattle.platform.core.model.Volume;
+import io.cattle.platform.docker.process.lock.DockerStoragePoolVolumeCreateLock;
 import io.cattle.platform.engine.handler.HandlerResult;
 import io.cattle.platform.engine.process.ProcessInstance;
 import io.cattle.platform.engine.process.ProcessState;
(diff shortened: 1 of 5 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalEventCreate.java
```diff
@@ -85,21 +86,23 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+            return;
+        }
+        String driverName = driver.toString();
+        final StoragePool storagePool = storagePoolDao.findStoragePoolByDriverName(event.getAccountId(), driverName);
+        if (storagePool == null) {
+            log.warn("Unknown storage pool. Returning. Driver name: {}", driverName);
+            return;
+        }
+
+        lockManager.lock(new DockerStoragePoolVolumeCreateLock(storagePool, event.getExternalId()), new LockCallbackNoReturn() {
```
cjellick: The important thing here is that I changed the lock from `ExternalEventLock` to `DockerStoragePoolVolumeCreateLock`. The other code moved so that I could get the storagePool before creating the lock, which is a parameter to the lock.

code/iaas/storage-service/pom.xml
```diff
@@ -31,7 +31,7 @@
         <dependency>
         	<groupId>io.cattle</groupId>
         	<artifactId>cattle-docker-common</artifactId>
-        	<version>0.5.0-SNAPSHOT</version>
+            <version>${project.version}</version>
```
cjellick: Unrelated to the bug, but stumbled upon this misuse of version and wanted to fix it.

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instancehostmap/DockerPostInstanceHostMapActivate.java
```diff
@@ -204,7 +204,7 @@ protected Volume createVolumeInStoragePool(final StoragePool storagePool, final
         if (volume != null)
             return volume;
 
-        return lockManager.lock(new DockerStoragePoolVolumeCreateLock(storagePool, dVol.getUri()), new LockCallback<Volume>() {
+        return lockManager.lock(new DockerStoragePoolVolumeCreateLock(storagePool, dVol.getExternalId()), new LockCallback<Volume>() {
```
cjellick: In all practicality, no, as it is the volume's name.

cjellick:
TF is test_healthcheck, which I see in builds from both master and other PRs. Thus, it is not related to this change.

alena1108:
LGTM. Health check TF is already addressed in master.

PR #1457, opened 2016-03-10
Exclude internal-dns ip from "recurse" section in config

alena1108:
@vincent99 

As we append rancher internal-dns ip to the list of "dns" docker param, we should exclude this value when build internal dns config, from the list defined in "recurse" (read from instance.dns param)


alena1108:
one of TFs - test_healthcheck- is addressed on master already; @hibooboo2 is looking at delete_project one.

PR #1458, opened 2016-03-10
Fixed invalid test causing TF on Drone

alena1108:
Test expected a newly added host to be registered as a health checker for the instance, but it didn't unles there was a network agent present on the host. Fixed by starting multi-port service triggering network agent start on the host.

@ibuildthecloud @cjellick


cjellick:
LGTM assuming green

alena1108:
TF is in delete_project, @hibooboo2 is looking at it

PR #1467, opened 2016-03-11
Changes to exclude the internal rancher lable 'io.rancher.service.hash'

prachidamle:
Changes to exclude the internal rancher lable 'io.rancher.service.hash' from the exported config yml files.
Also changed the py test to check if the label gets excluded.

https://github.com/rancher/rancher/issues/3937

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/service/impl/ServiceDiscoveryApiServiceImpl.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/service/impl/ServiceDiscoveryApiServiceImpl.java
@@ -127,6 +127,7 @@ public class ServiceDiscoveryApiServiceImpl implements ServiceDiscoveryApiServic
                 Map<String, Object> cattleServiceData = ServiceDiscoveryUtil.getServiceDataAsMap(service,
                         launchConfigName, allocatorService);
                 Map<String, Object> composeServiceData = new HashMap<>();
+                excludeLables(cattleServiceData);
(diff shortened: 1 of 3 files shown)


prachidamle:
@ibuildthecloud @alena1108 This is the fix you needed to exclude the rancher-compose label.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/service/impl/ServiceDiscoveryApiServiceImpl.java
```diff
@@ -151,6 +152,19 @@ private String convertToYml(Map<String, Object> dockerComposeData) {
     }
 
     @SuppressWarnings("unchecked")
+    private void excludeLables(Map<String, Object> composeServiceData) {
+        if (composeServiceData.get(InstanceConstants.FIELD_LABELS) != null) {
+            Map<String, String> labels = new HashMap<>();
```
alena1108: You can use collectionUtils "toMap" method here (casts the object to map)

prachidamle: Yes, but any particular advantage of doing so? I am not iterating and casting each object of the map.

alena1108: @prachidamle there is some - It does safe casting :) But again, its super minor, just FYI

alena1108:
LGTM

PR #1468, opened 2016-03-11
Fix in healthcheck reconcile

alena1108:
On healthcheckinstancehostmap.remove we do service health check reconcile. There was a bug in the logic - if no other hosts were present as health checkers, instance was reported as unhealthy. Re-worked the logic by sending a reconcile event, then calculate the report based on the health of the present health checkers. If none are present - don't send an update.


cjellick:
LGTM

alena1108:
https://github.com/rancher/rancher/issues/3933

PR #1470, opened 2016-03-12
Resolve to vip on dns when assignIpAddress flag is set to true

alena1108:
Instead of relying on a sole presence of the flag. It can be passed as
false to the API


cjellick:
LGTM

PR #1474, opened 2016-03-12
Skip inactivIsh states on stack.stopServices

alena1108:
@ibuildthecloud


alena1108:
https://github.com/rancher/rancher/issues/3944

PR #1478, opened 2016-03-13
Dont delete shared pools when an agent is removed

cjellick:
If a storage pool agent container goes unhealthy, it will get removed
and replaced. This causes the associated agent to get removed and
replaced. This change prevents the storage pool assocaited with the
agent from being removed so that when the new storage pool agent
container comes online, it can be associated with the existing pool.

The logic for associating a new agent container with an existing pool
is already in place in `ExternalEventCreate.java`

Addresses: https://github.com/rancher/rancher/issues/3333

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/agent/AgentRemove.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/agent/AgentRemove.java
@@ -1,8 +1,10 @@
 package io.cattle.platform.process.agent;
 
 import io.cattle.platform.agent.util.AgentUtils;
+import io.cattle.platform.core.constants.StoragePoolConstants;
 import io.cattle.platform.core.model.Account;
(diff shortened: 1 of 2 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/agent/AgentRemove.java
```diff
@@ -40,6 +42,13 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
             }
 
             for (Object obj : objectManager.children(agent, clz)) {
+                if (obj instanceof StoragePool) {
+                    StoragePool sp = (StoragePool)obj;
+                    if (StoragePoolConstants.TYPE.equals(sp.getKind())) {
```
ibuildthecloud: Is this looking for kind=="storagePool".  So basically anything that is not docker?

ibuildthecloud:
LGTM

PR #1494, opened 2016-03-17
Changes to API to obfuscate the references present in the embedded resourceFields 

prachidamle:
Changes to obfuscate the references present in the nested resourceFields of a resource where the resoureField is an object itself containing db references.

Tested the API UI manually as shown below.

Example: http://localhost:8080/v1/projects/1a5/services/1s2

Here in API response, the "service" object has resourceField "launchConfig" which is an object containing key "dataVolumesFrom" whose value is of type array[reference[container]]

So value of "dataVolumesFrom" should be obfuscated db container id's

"id": "1s2",
"type": "service",
"links": { … },
"actions": { … },
"name": "random-957277-1458161855",
"state": "inactive",
"accountId": "1a5",
"assignServiceIpAddress": false,
"createIndex": null,
"created": "2016-03-16T20:57:36Z",
"createdTS": 1458161856000,
"description": null,
"environmentId": "1e1",
"externalId": null,
"fqdn": null,
"healthState": null,
"kind": "service",
"launchConfig": {
"capAdd": [
"SYS_MODULE",
],
"capDrop": [
"SYS_MODULE",
],
"command": [ 2 items
"sleep",
"42",
],
"cpuSet": "2",
"cpuShares": 400,
"dataVolumes": [
"/foo",
],
"dataVolumesFrom": [
"1i1",
],
"dns": [ 2 items
"8.8.8.8",
"1.2.3.4",
],
"dnsSearch": [ 2 items
"foo",
"bar",
],
"domainName": "rancher.io",
"entryPoint": [ 2 items
"/bin/sh",
"-c",
],
"environment": {
"TEST_FILE": "/etc/testpath.conf",
},
"healthCheck": {
"healthyThreshold": 5,
"interval": 4,
"name": "check1",
"port": 200,
"requestLine": "index.html",
"responseTimeout": 3,
"unhealthyThreshold": 6,
"strategy": "recreate",
},
"hostname": "test",
"imageUuid": "sim:random-333122-1458161843",
"instanceLinks": {
"container2_link": "1i3",
},
"kind": "container",
"labels": {
"foo": "bar",
},
"memory": 8000000,
"networkMode": "managed",
"ports": [ 2 items
"64122:8681/tcp",
"58021:8082/tcp",
],
"privileged": true,
"publishAllPorts": false,
"readOnly": false,
"requestedHostId": "1h1",
"startOnCreate": true,
"stdinOpen": true,
"tty": true,
"user": "test",
"version": "0",
"workingDir": "/",
"vcpu": 1,
},

https://github.com/rancher/rancher/issues/594

### DIFF
--- a/code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/id/IdFormatterUtils.java
+++ b/code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/id/IdFormatterUtils.java
@@ -1,7 +1,9 @@
 package io.github.ibuildthecloud.gdapi.id;
 
+import io.github.ibuildthecloud.gdapi.factory.SchemaFactory;
 import io.github.ibuildthecloud.gdapi.model.Field;
 import io.github.ibuildthecloud.gdapi.model.FieldType;
+import io.github.ibuildthecloud.gdapi.model.Schema;
 
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
@@ -10,11 +12,12 @@ import java.util.Map;
 
 public class IdFormatterUtils {
 
-    public static Object formatReference(Field field, IdFormatter formatter, Object value) {
-        return formatReference(field.getTypeEnum(), field.getSubTypeEnums(), field.getSubTypes(), formatter, value);
+    public static Object formatReference(Field field, IdFormatter formatter, Object value, SchemaFactory schemaFactory) {
+        return formatReference(field, field.getTypeEnum(), field.getSubTypeEnums(), field.getSubTypes(), formatter, value, schemaFactory);
     }
 
-    private static Object formatReference(FieldType fieldType, List<FieldType> subTypeEnums, List<String> subTypes, IdFormatter formatter, Object value) {
+    private static Object formatReference(Field field, FieldType fieldType, List<FieldType> subTypeEnums,
+            List<String> subTypes, IdFormatter formatter, Object value, SchemaFactory schemaFactory) {
         if (value == null) {
             return value;
         }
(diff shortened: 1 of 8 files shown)


prachidamle:
@ibuildthecloud @alena1108 please review the fix.

How do you add tests for API UI? In Integration tests, the ids in reponses are already obfuscated.

alena1108:
@prachidamle could you add an integration test for it? you can extend existing test_svc_discovery/test_activate_single_service test case. Validating that would be enough as the ids are expected to be obfuscated by the test.

prachidamle:
@alena1108 changed the PR with integration tests additions

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -190,6 +190,7 @@ def test_activate_single_service(client, context, super_client):
     assert svc.launchConfig.healthCheck.port == 200
     assert svc.metadata == metadata
     assert svc.launchConfig.version == '0'
+    assert svc.launchConfig.requestedHostId == host.id
```
ibuildthecloud: @prachidamle @vincent99 very much wants the embedded types to have a type field on them.  So can you make it such that the below works.  Make sure that if the embedded type already has a field named "type" we don't override it.

```python
assert assert svc.launchConfig.type == 'launchConfig'
````

File: code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/id/IdFormatterUtils.java [hunk 2]
```diff
@@ -10,10 +12,60 @@
 
 public class IdFormatterUtils {
 
-    public static Object formatReference(Field field, IdFormatter formatter, Object value) {
-        return formatReference(field.getTypeEnum(), field.getSubTypeEnums(), field.getSubTypes(), formatter, value);
+    public static Object formatReference(Field field, IdFormatter formatter, Object value, SchemaFactory schemaFactory) {
+        return formatReference(field, field.getSubTypeEnums(), field.getSubTypes(), formatter, value, schemaFactory);
     }
 
+    private static Object formatReference(Field field, List<FieldType> subTypeEnums, List<String> subTypes, IdFormatter formatter, Object value, SchemaFactory schemaFactory) {
```
ibuildthecloud: Can you combine this method with the similar one below?  Seems like the only difference is that the TYPE check needs field and schemaFactory, so in the condition of TYPE just check `if (field == null || schemaFactory == null) { return value; }`

tests/integration/cattletest/core/test_svc_discovery.py
prachidamle: I have added the code to add type, hope it looks good.

code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/id/IdFormatterUtils.java
prachidamle: @ibuildthecloud refactored it

prachidamle:
@ibuildthecloud please review - took care of comments and checkstyle

ibuildthecloud:
@prachidamle There are quite a few test failures that seen related.

prachidamle:
@ibuildthecloud reverted the changes made to add "type" field to embedded objects. This is the reason for most test failures. We should add this change separate.

prachidamle:
@ibuildthecloud I have resolved all the prior test failures. The recent failure is not seen on my local setup, that tests passes for me.

prachidamle:
@ibuildthecloud woohoo passed passed passed

PR #1495, opened 2016-03-17
Addressing various performance issues

alena1108:
DB migration for serviceIndexId field is covered as a part of https://github.com/rancher/cattle/pull/1532


alena1108:
@ibuildthecloud let me know if any more performance fixes are needed. Load Balancer config item could use some more TLC, but as config generation is all done based on links there, so shouldn't be too crazy. And I've put some changes already around getting instances/nics for them.

PR #1497, opened 2016-03-17
Maintain the random public ports assigned to a service and sidekicks after upgrade

prachidamle:
During upgrade, when the upgraded launchconfig does not have the public port
mapping specified, but same private port is present then assign the same public port.

So if a svc has port mapping 50000:80 and it is upgraded with ports =
[:80], then the port mapping in the upgraded service should be same as before upgrade and the same public port should be assigned

Added tests in test_svc_upgrade

https://github.com/rancher/rancher/issues/3979

### DIFF
--- a/tests/integration/cattletest/core/test_svc_upgrade.py
+++ b/tests/integration/cattletest/core/test_svc_upgrade.py
@@ -863,3 +863,40 @@ def test_rollback_id(context, client, super_client):
     expose_map = maps[0]
     c2 = super_client.reload(expose_map.instance())
     assert c1.uuid == c2.uuid
+
+
+def test_in_service_upgrade_port_mapping(context, client, super_client):
+    env = client.create_environment(name=random_str())
+    env = client.wait_success(env)
+    image_uuid = context.image_uuid
+    launch_config = {"imageUuid": image_uuid, 'ports': ['80', '82/tcp']}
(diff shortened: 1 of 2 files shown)


tests/integration/cattletest/core/test_svc_upgrade.py
```diff
@@ -863,3 +863,35 @@ def test_rollback_id(context, client, super_client):
...
     c2 = super_client.reload(expose_map.instance())
     assert c1.uuid == c2.uuid
+
+
+def test_in_service_upgrade_port_mapping(context, client, super_client):
+    env = client.create_environment(name=random_str())
+    env = client.wait_success(env)
+    image_uuid = context.image_uuid
+    launch_config = {"imageUuid": image_uuid, 'ports': ['80']}
+    secondary1 = {"imageUuid": image_uuid, "name": "secondary1",
```
alena1108: Minor, but can we add couple more use cases to the test:

* port defined with a "/protocol"
* updated launch config having more ports that the present launch config

alena1108:
Other than my minor comment on test, LGTM

tests/integration/cattletest/core/test_svc_upgrade.py
prachidamle: Added test for port with protocol and also a new public:private port pair in upgrade.

PR #1514, opened 2016-03-18
Service container name is available globally via dns

alena1108:
https://github.com/rancher/rancher/issues/4004

Also changed alias - link - names representation on dns. Instead of just linkName. it would be
    linkName.stackName.rancher.internal. + linkName.rancher.internal.


alena1108:
Corresponding host-api PR removing service search domain: https://github.com/rancher/host-api/pull/65

alena1108:
bug causing test_restart_stack TF, is fixed in another PR

PR #1529, opened 2016-03-20
Set search domain on regular containers too

alena1108:
Also added authoritative section to default dns settings

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
@@ -2,8 +2,11 @@ package io.cattle.platform.process.instance;
 
 import io.cattle.platform.core.constants.AgentConstants;
 import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.constants.NetworkConstants;
 import io.cattle.platform.core.model.Instance;
 import io.cattle.platform.core.util.SystemLabels;
+import io.cattle.platform.docker.constants.DockerInstanceConstants;
+import io.cattle.platform.docker.constants.DockerNetworkConstants;
 import io.cattle.platform.engine.handler.HandlerResult;
 import io.cattle.platform.engine.handler.ProcessPreListener;
 import io.cattle.platform.engine.process.ProcessInstance;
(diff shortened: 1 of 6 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -33,12 +36,36 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
         Map<String, Object> labels = DataAccessor.fieldMap(instance, InstanceConstants.FIELD_LABELS);
+        Map<Object, Object> data = new HashMap<>();
         if (labels.containsKey(SystemLabels.LABEL_AGENT_CREATE)
                 && labels.get(SystemLabels.LABEL_AGENT_CREATE).equals("true")) {
-            Map<Object, Object> data = new HashMap<>();
             List<String> dataVolumes = DataAccessor.fieldStringList(instance, InstanceConstants.FIELD_DATA_VOLUMES);
             dataVolumes.add(AgentConstants.AGENT_INSTANCE_BIND_MOUNT);
             data.put(InstanceConstants.FIELD_DATA_VOLUMES, dataVolumes);
+        }
+        if (InstanceConstants.KIND_CONTAINER.equalsIgnoreCase(instance.getKind())) {
```
ibuildthecloud: Do `InstanceContants.CONTAINER_LIKE.contains(instance.getKind())`

tests/integration/cattletest/core/test_docker.py
```diff
@@ -637,9 +637,9 @@ def test_container_fields(docker_client, super_client):
     assert set(c.data['dockerInspect']['HostConfig']['CapAdd']) == set(caps)
     assert set(c.data['dockerInspect']['HostConfig']['CapDrop']) == set(caps)
     actual_dns = c.data['dockerInspect']['HostConfig']['Dns']
-    assert set(actual_dns) == set(['8.8.8.8', '1.2.3.4'])
+    assert set(actual_dns) == set(['8.8.8.8', '1.2.3.4', '169.254.169.250'])
```
ibuildthecloud: Why did this change?  I though if the user configures DNS we don't override it?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
alena1108: ok

tests/integration/cattletest/core/test_docker.py
alena1108: no...we still put our dns there, and its set on the container, and our dns will be used. We just don't propagate it to dns config yml

alena1108:
Also fixed the order of search domains in dns config

alena1108:
Here is how default section will look like:

```
"default": {
		"search": ["rancher.internal"],
		"recurse": ["192.168.1.1"], // this is read from host config
		"authoritative": ["rancher.internal"],
....
```

client section:

```
"10.42.114.203": {
		"search": ["foo.rancher.internal", "rancher.internal"],
		"recurse": [],
		"authoritative": [],
```

/etc/resolv.conf:

```
search foo.rancher.internal rancher.internal
nameserver 169.254.169.250
```

@vincent99 @ibuildthecloud confirm that everything above is correct

tests/integration/cattletest/core/test_docker.py
ibuildthecloud: omg, i can't keep all this straight

ibuildthecloud:
@alena1108 do you define `rancher-metadata.rancher.internal => 169.254.169.250`in the defaults section?

alena1108:
@ibuildthecloud yes I do

tests/integration/cattletest/core/test_docker.py
alena1108: lol :)

PR #1530, opened 2016-03-20
service and servicelink name validation

prachidamle:
service names and service link names become DNS labels so the name is heavily restricted to
what DNS allows. The rules are (or probably aren't, but should be):

1-63 characters long
Consisting only of A-Z, 0-9, and hyphen
Can't start of end with a hyphen
Can't contain 2 consecutive hyphens

https://github.com/rancher/rancher/issues/680


prachidamle:
@ibuildthecloud @alena1108 again I cannot reproduce this test failure on my setup.

PR #1532, opened 2016-03-20
DB data migration for serviceIndex

alena1108:
and trigger healthState update for service/stack

@ibuildthecloud for health state upgrade, decided to just call "update" on the service. Bumping up reconcile config item had some caveats:

1) reconcile item could have not been present for the service
2) reconcile gets triggered only for services in active state, and health state should be updated for inactive services as well.

If 1) was possible to workaround by creating a new config_item_status, for 2) there is no workaround


alena1108:
https://github.com/rancher/rancher/issues/3994
https://github.com/rancher/rancher/issues/4009

alena1108:
Update: this PR will update service index only; health state update will be triggered via config items, and addressed by https://github.com/rancher/cattle/pull/1543

PR #1537, opened 2016-03-21
Servicename space back to dns

alena1108:
https://github.com/rancher/rancher/issues/4060

Corresponding host-api: https://github.com/rancher/host-api/pull/66


alena1108:
Also fixed the bug causing json mapper to bail - it didn't like fixed size array

https://github.com/rancher/rancher/issues/4065

PR #1540, opened 2016-03-21
Support ip in lb ports

alena1108:
@cjellick could you review?

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/LoadBalancerDeploymentUnitInstance.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/LoadBalancerDeploymentUnitInstance.java
@@ -63,7 +63,11 @@ public class LoadBalancerDeploymentUnitInstance extends DefaultDeploymentUnitIns
                 if (spec.getPublicPort() == null) {
                     spec.setPublicPort(spec.getPrivatePort());
                 }
-                String fullPort = spec.getPublicPort().toString() + ":" + spec.getPublicPort().toString();
+                // private port of LB instance = public port. Original private port specified,
+                // is set on lb target in config backend section
+                // protocol is always tcp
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
No way... It can't be that easy.

vincent99:
Indeed, no way this addresses all the bits needed.

alena1108:
@vincent99 @ibuildthecloud I will work on fixing haproxy config part, changed to WIP

alena1108:
@vincent99 @ibuildthecloud fixed binding on haproxy; leaving as WIP till we decide on hostname routing rules support

resources/content/config-content/haproxy/content/etc/haproxy/haproxy.cfg.ftl
```diff
@@ -47,7 +47,7 @@ defaults
 <#if listener.sourceProtocol == "https"><#assign protocol="http"></#if>
 <#if listener.sourceProtocol == "ssl"><#assign protocol="tcp"></#if>
 frontend ${listener.uuid}_frontend
-        bind *:${sourcePort}<#if (listener.sourceProtocol == "https" || listener.sourceProtocol == "ssl") && certs?has_content> ssl crt /etc/haproxy/certs/<#if !defaultCert??> strict-sni</#if></#if>
+        bind ${listener.sourceIp}:${sourcePort}<#if (listener.sourceProtocol == "https" || listener.sourceProtocol == "ssl") && certs?has_content> ssl crt /etc/haproxy/certs/<#if !defaultCert??> strict-sni</#if></#if>
```
ibuildthecloud: I don't understand this change.  How can haproxy bind to an IP?

alena1108:
@vincent99 @ibuildthecloud ok, final resolution. Binding to specific host ip address on haproxy would work only in host networking mode. As our haproxy runs on the managed, this can't be done. I'm switching back to my original fix, where we just configure port on the host by binding it to specific ip. Then all the traffic coming to this ip/port, will get forwarded to the listening port of haproxy container.

PR #1552, opened 2016-03-23
Fix to service name and link name validation to allow single character

prachidamle:
Fix to service name and link name validation to allow single character in the name
Also, added the fix for cross stack links which can contain /
Added some tests too

https://github.com/rancher/rancher/issues/4090

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
@@ -8,7 +8,7 @@ import java.util.regex.Pattern;
 
 public class AbstractDefaultResourceManagerFilter extends AbstractResourceManagerFilter implements Priority {
 
-    private static final Pattern DNS_NAME_PATTERN = Pattern.compile("^[a-zA-Z0-9](?!.*--)[a-zA-Z0-9-]*[a-zA-Z0-9]$");
+    private static final Pattern DNS_NAME_PATTERN = Pattern.compile("^[a-zA-Z0-9](?!.*--)[a-zA-Z0-9-]*$");
 
     @Override
     public int getPriority() {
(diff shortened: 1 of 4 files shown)


prachidamle:
@ibuildthecloud @vincent99 this is to fix the single char bug in the regex used

vincent99:
I suppose this works, probably, but I've done a lot of regex and don't know exactly what `(?!.*--)` means without looking it up, and then you're adding basic string checks because the regex doesn't actually work.  So why not just make all the exceptions simple checks and us a human-readable regex:

```
if ( matches('[^a-zA-Z0-9-]') || startsWith('-') || endsWith('-') || contains('--') {
  // bad
}
else {
  // good
}
```

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
```diff
@@ -8,7 +8,7 @@
 
 public class AbstractDefaultResourceManagerFilter extends AbstractResourceManagerFilter implements Priority {
 
-    private static final Pattern DNS_NAME_PATTERN = Pattern.compile("^[a-zA-Z0-9](?!.*--)[a-zA-Z0-9-]*[a-zA-Z0-9]$");
+    private static final Pattern DNS_NAME_PATTERN = Pattern.compile("^[a-zA-Z0-9](?!.*--)[a-zA-Z0-9-]*[a-zA-Z0-9]*$");
```
ibuildthecloud: The end of this pattern is redundant.

prachidamle: @vincent99 (?!{regex}) is the negative lookahead supported by java to match "something not followed by something" pattern

prachidamle: @ibuildthecloud I have removed the redundant portion.

PR #1556, opened 2016-03-23
Fix bind ip address logic

cjellick:
Addresses https://github.com/rancher/rancher/issues/53


alena1108:
LGTM

cjellick:
TF is legit. do not merge

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/PortsConstraint.java
```diff
@@ -43,6 +45,16 @@ public boolean matches(AllocationAttempt attempt,
         return true;
     }
 
+    private boolean publicIpTheSame(Port requestedPort, Port portUsed) {
+        if (requestedPort.getPublicIpAddressId() != null) {
+            return requestedPort.getPublicIpAddressId().equals(portUsed.getPublicIpAddressId());
+        } else {
```
cjellick: @alena1108 Since your LGTM, I had to tweak. This line used to say 
```
        } else if (portUsed.getPublicIpAddressId() == null) {
```
but that did not account for the scenario where the requested port's IP was not yet assigned bc that doesn't happen until later. The pre-existing test `test_allocation.test_port_constraint` caught this.

PR #1564, opened 2016-03-24
Log instance name in service's transitioning message

alena1108:
instead of its id.

@ibuildthecloud


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceUpdateActivate.java
```diff
@@ -106,8 +107,19 @@ protected String obfuscateId(TimeoutException ex) {
...
+            if (Instance.class.getSimpleName().equalsIgnoreCase(resourceType)) {
+                // in most cases it's going to be instance
+                Instance instance = objectManager.loadResource(Instance.class, resourceId);
+                if (instance != null && instance.getName() != null) {
+                    obfuscatedId = instance.getName();
+                }
+            }
+            
+            if (obfuscatedId == null) {
+                idFormatter.formatId(resourceType, resourceId);
```
ibuildthecloud: You don't seem to be making an assignment here.

alena1108: ...thank you

PR #1579, opened 2016-03-28
Simplify allocation

cjellick:
- Remove ComputeConstraint
- Make AccountConstraint.matches always return true
- add a test that checks allocation time


cjellick:
Closing in favor of #1580

PR #1585, opened 2016-03-28
Add agent null check to event api filters

cjellick:
Addresses https://github.com/rancher/rancher/issues/4176


alena1108:
LGTM assuming the client knows how to handle client exception that is being thrown

PR #1589, opened 2016-03-29
NPE check in case service was already removed

alena1108:
@ibuildthecloud yet to debug whey service_consume_maps weren't cleaned up on consumed service removal


alena1108:
https://github.com/rancher/rancher/issues/4196

PR #1593, opened 2016-03-31
Changes to the ProxyFilter to use the encoded RequestURI 

prachidamle:
Fixes to ProxyFilter so that the url encoding is preserved in case the RequestURI have spaces

Also we should not be decoding the proxy uri, else we lose the encoded spaces

https://github.com/rancher/rancher/issues/3124


prachidamle:
@alena1108 @ibuildthecloud I don't see the test failure related to my fix. I see this fail on master branch on my local setup as well.

PR #1595, opened 2016-04-01
Reverting the changes for service linkName validation

prachidamle:
Now the linkName allows any characters, no length limits, does not follow the DNS label RFC standard


prachidamle:
@ibuildthecloud please review the reverted code

vincent99:
Okay we can argue about allowing dots but it *definitely* shouldn't just allow arbitrary garbage.

ibuildthecloud:
@vincent99 I think _ and . should be allowed.  That's it

prachidamle:
@vincent99 @ibuildthecloud opening another PR to allow _ and .  , closing this one

vincent99:
And 1-253 characters, not consecutive dots, not starting with a dot...

ibuildthecloud:
@vincent99 Yeah, sorry, I mean the only thing we should relax is _ and . but otherwise validate everything else.

PR #1596, opened 2016-04-01
Changes to service linkName validation rules.

prachidamle:
Rules now are:
"validChars": "a-zA-Z0-9-/.\_"
Cannot start or end with a "-" or _ or "."
Cannot contain "--" or ".." or "__"
MinLength 1 and MaxLength 63

https://github.com/rancher/rancher/issues/4252

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
@@ -27,4 +27,40 @@ public class AbstractDefaultResourceManagerFilter extends AbstractResourceManage
         }
     }
 
+    protected void validateLinkName(String linkName){
+        if(linkName != null && !linkName.isEmpty()){
+            if(linkName.startsWith(".") || linkName.endsWith(".") || linkName.contains("..")) {
+                ValidationErrorCodes.throwValidationError(ValidationErrorCodes.INVALID_CHARACTERS,
+                        "name");
+            }
+            
+            //split around a "."
+            String[] parts = linkName.split("\\.");
+            if(parts.length > 1) {
+                //check total length <= 253
+                if (linkName.length() > 253) {
+                    ValidationErrorCodes.throwValidationError(ValidationErrorCodes.MAX_LENGTH_EXCEEDED,
+                            "name");
+                }
+            } 
+
+            for (String linkPart : parts) {
+                if(linkPart.startsWith("-") || linkPart.endsWith("-") || linkPart.contains("--")) {
+                    ValidationErrorCodes.throwValidationError(ValidationErrorCodes.INVALID_CHARACTERS,
+                            "name");
+                }
+                //check length
+                if (linkPart.length() < 1) {
+                    ValidationErrorCodes.throwValidationError(ValidationErrorCodes.MIN_LENGTH_EXCEEDED,
+                            "name");
+                }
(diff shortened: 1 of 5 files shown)


prachidamle:
@vincent99 @ibuildthecloud please review
Could not get or conditions to work with ?!.*, so back to java checks instead of regex

vincent99:
- The regex check doesn't work.  Any input matches 0 or more of those characters.
- None of the tests have an actually invalid char to catch that.
- maxLength should be 253 since it's (potentially) FQDN now and not a single label.
- Each individual label (split on dots) within the input still needs to be 1-63 chars though, a single label of 100 chars isn't valid.
- Underscore shouldn't _really_ be allowed at all so if we're going to allow it there's not much point enforcing rules about starting/ending/containing consecutive.

prachidamle:
@vincent99 
- For regex, I see I need a + instead of *, do you mean the error is an empty string matches or any thing else?
- So incase a label has no dots, then the length should be 63 or 253?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
```diff
@@ -27,4 +29,17 @@ protected void validateDNSPatternForName(String name) {
         }
     }
 
+    protected void validateLinkPatternForName(String name) {
+        if (name != null) {
+            if(!LINK_NAME_PATTERN.matcher(name).matches()
```
ibuildthecloud: Is this rexexp redundant since the API layer checks for these characters?

prachidamle:
@vincent99 @ibuildthecloud can you comment regarding splitting around a "/" and a "."

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/common/AbstractDefaultResourceManagerFilter.java
prachidamle: @ibuildthecloud true, only difference being the above regex does not allow a "/" in the link part - is that important to check or should I just remove this check?

vincent99:
- This regex still doesn't work.  It needs to either test that *all characters match* the list or that *any don't match*.
- Please add tests that check empty string, a single valid/invalid char, multiple valid/invalid chars, etc.  tests should be catching this stuff.
- Each dot-delimited label is limited to 63 chars, and the whole string to 253.  If there are no dots then it is a single label and must still be <= 63 chars.
- I don't understand why slash is in the validChars.  It seems like there was confusion for mapping stack+service name, e.g. `Default/blahsvc -> some_as_name` for rancher-compose.  But the left-half becomes the serviceId in the API and the right-half is the name, which needs to go into DNS, where slash is definitely not allowed.

prachidamle:
@ibuildthecloud can you pls clarify the usecase of a "/" in the linkname

prachidamle:
@ibuildthecloud @vincent99 please review after rebase - 
The final rules added are:

1) "validChars": "a-zA-Z0-9-._"  - This is applied by validation api layer(removed redundant regex)
2) Cannot start or end with a "-" or  "."
3) Cannot contain "--" or ".."
4) Each dot-delimited label is limited to 63 chars, and the whole string to 253. If there are no dots then it is a single label and must still be <= 63 chars.
5) Added tests for checking empty string, a single valid/invalid char, multiple valid/invalid chars, length> 253, linkName having a "." and one part > 63

prachidamle:
@vincent99 It looks like empty service linkNames are allowed, so I will be removing that check

PR #1605, opened 2016-04-07
Proxy protocol support for Lb

alena1108:
@ibuildthecloud


code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/LoadBalancerInfoDaoImpl.java
```diff
@@ -106,17 +108,17 @@
                 }
             }
             listenersToReturn.add(new LoadBalancerListenerInfo(lbService, privatePort, sourcePort,
-                    sourceProtocol, targetPort));
+                    sourceProtocol, targetPort, proxyProtocolPorts.contains(privatePort.toString())));
```
ibuildthecloud: Can privatePort ever be null?

PR #1608, opened 2016-04-10
Don't set rancher.internal search domain for system containers

alena1108:


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -54,10 +54,13 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
 
     protected void setDns(Instance instance, Map<String, Object> labels, Map<Object, Object> data) {
+        if (instance.getSystemContainer() != null) {
+            return;
+        }
         if (InstanceConstants.KIND_CONTAINER.equals(instance.getKind())) {
             boolean addDns = true;
             if (labels.containsKey(SystemLabels.LABEL_USE_RANCHER_DNS)
-                    && !Boolean.valueOf(SystemLabels.LABEL_USE_RANCHER_DNS)) {
+                    && !Boolean.valueOf(labels.get(SystemLabels.LABEL_USE_RANCHER_DNS).toString())) {
```
alena1108: Also while testing, found this embarrassing bug. Might be one of my late-GA-nights- checkins

ibuildthecloud: Will this cause NPE

alena1108: Might be if someone passed label's key - which presence we check in the first condition - with the null value. Will add an extra check.

PR #1619, opened 2016-04-14
Load balancer should not have search domains set

jimengliu:
Make DNS search domain only available for regular service container, not Load Balancer and external service containers.
Fixes https://github.com/rancher/rancher/issues/4373

@alena1108


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/LoadBalancerDeploymentUnitInstance.java
```diff
@@ -73,4 +73,9 @@ protected void setPorts(Map<String, Object> launchConfigData) {
             launchConfigData.put(InstanceConstants.FIELD_PORTS, newPorts);
         }
     }
+
+    @Override
+    public List<String> getSearchDomains() {
+       return null;
+    }
 }
```
alena1108: new line is needed

alena1108:
LGTM. TF looks unrelated.

PR #1620, opened 2016-04-15
tests: fix session smashing (aka Method Not Allowed errors)

dx9:
image


dx9:
So it is reproducible... attached the run log also. Contains some scheduling failures that don't appear in the drone console.

https://gist.github.com/dx9/14ac516f6003585e402913802d365811

LMK if anyone knows how to squash it.

PR #1622, opened 2016-04-19
Add Mesos environment

LLParse:
[Corresponding UI changes in this pull request.](https://github.com/rancher/ui/pull/604)


alena1108:
@LLParse looks like test_authorization python test needs to be fixed - mesos param needs to be added to the list of project's fields.

PR #1623, opened 2016-04-21
Reduced number of healthcheck config items increments

alena1108:
On instance stop/start, reload health check config only for the hosts where network agents running health checks for this instance are located. 

@ibuildthecloud


cjellick:
LGTM

PR #1624, opened 2016-04-23
tests: fix test_machines by marking as nonparallel and added cleanup …

dx9:


dx9:
CI failed in test_ha_config_script

PR #1625, opened 2016-04-24
ServiceDiscoveryConstants: ServiceDiscoveryConstants.KIND.x.name() do…

dx9:


dx9:
test_ha_config_script & test_docker_ports_from_container failures

ibuildthecloud:
@dx9 Good catch.  I'm not terribly sure why the code was done this way but it is inconsistent with the rest of the code base.  I would prefer instead of adding a toString() method that the Enum was removed and the constants where changed to just be a string.  That would be consistent with the rest of the code base.  For example:

```java
      public static final String SERVICE = "service";
      public static final String LOADBALANCERSERVICE = "loadBalancerService";
  ```

alena1108:
LGTM. TFs are unrelated, and can be addressed outside of this PR context.

PR #1628, opened 2016-04-26
Non removed check against healthcheckinstancehostmap

alena1108:
https://github.com/rancher/rancher/issues/4553


cjellick:
LGTM pending completion of CI and if we don't see test_host_create_container_inactive fail again.

@alena1108 feel free to merge after that

PR #1629, opened 2016-04-26
Adding environment uuid to Stack metadata

prachidamle:
Changes to add the environment UUID under stack metadata, useful for consumers that need some unique key per environment.

https://github.com/rancher/rancher/issues/4557


prachidamle:
@alena1108 cattle changes for metadata to add environment UUID.

alena1108:
LGTM

PR #1640, opened 2016-04-29
Cleanup k8s stacks when k8s is disabled for environment

alena1108:
* kubernetes://
* kubernetes-loadbalancers://

stacks should be removed after k8s environment is disabled

https://github.com/rancher/kubernetes/pull/10


alena1108:
https://github.com/rancher/rancher/issues/4876

PR #1641, opened 2016-04-29
WIP - DNM Fix ipsec config race

dx9:
@ibuildthecloud 

This doesn't seem to cause direct test failures but shows up in the logs.

SimulatorConfigUpdateProcessor tries to download the ipsec config before AgentInstanceIpsecNetworkServiceCreate has defined the key. This causes the template to fail because ipsecKey is undefined.

Unsure if this is expected behaviour. If so can we change the key in the template to be optional to avoid the log error?

If it's a real issue, another idea would be to try multiple times to read the key in IpsecInfoFactory (seems lame).

Thoughts?

racy logs: https://gist.github.com/dx9/fd4c91b112919be7131751bade55ddd8

### DIFF
--- a/code/iaas/config-item/api/src/main/java/io/cattle/platform/configitem/api/manager/ConfigContentManager.java
+++ b/code/iaas/config-item/api/src/main/java/io/cattle/platform/configitem/api/manager/ConfigContentManager.java
@@ -82,7 +82,6 @@ public class ConfigContentManager extends AbstractNoOpResourceManager {
                 throw new ClientVisibleException(ResponseCodes.NOT_FOUND);
             }
         } catch (IOException e) {
-            log.error("Failed to download [{}] for agent [{}]", id, agentId, e);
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
@dx9 out of curiosity.  Are you running on a really fast machine?

ibuildthecloud:
Generally we just make things optional (check for null).  Being that most everything is eventually consistent it is possible we download config before something has been defined.  The issue occurs is if we don't end up triggering a config update after the previously set value becomes set.

So in sort, the first option seems good.

dx9:
yeppers, R710

Sounds good. Thanks for taking a look.

ibuildthecloud:
Sounds like we need to run our CI on faster machines 🙂

alena1108:
@ibuildthecloud @leodotcloud is this still relevant given that psk.txt.ftl file is gone from cattle code base?

PR #1653, opened 2016-05-04
Disk size scheduler

jimengliu:
1. add cache manager to load host disk information on demand. For now now resync from changes.
2. add disk size constraint and its provider to use disksize label to verify against
3. add post remove of host listener
4. add post purge of instance listener

@prachidamle @alena1108


alena1108:
@jimengliu having tests would help to understand the flow; could you add them?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraintProvider.java
```diff
@@ -0,0 +1,49 @@
...
+        @SuppressWarnings("unchecked")
+        Map<String, String> labels = DataAccessor.fields(instance).withKey(InstanceConstants.FIELD_LABELS)
+                .as(Map.class);
+        if (labels == null) {
+            return;
+        }
+
+        for (Map.Entry<String, String> labelEntry : labels.entrySet()) {
+            String labelKey = labelEntry.getKey();
+            if (labelKey.startsWith("io.rancher.scheduler.disksize.")) {
```
alena1108: we should define the label prefix in SystemLabels class, and reference it here

jimengliu: thanks, I was wondering where to put those constant strings.

jimengliu:
added tests: 
1. one container with a volume scheduling on a host with not enough disk space will fail
2. one container with a volume scheduling on a host wit enough disk space will succeed
3. one container with 2 volumes scheduling on a host that can only satisfy one volume but not the other will fail
4. one container with 2 volumes scheduling on a host that can satisfy one volume on one of the disks, the other on another disk with enough space, will succeed

alena1108:
@jimengliu there was something wrong with the tests, looks like it got stuck. When you run your tests locally, can you make sure that "-Didempotent.checks=true" is set as a jvm option (run->debug configuration->pick cattle app ->arguments). Thats how Drone runs them. See if its your test that gets stuck

jimengliu:
ok, let me try that locally to see if it is stuck.

jimengliu:
had another one sent out

PR #1655, opened 2016-05-06
Disk size scheduler

jimengliu:
add cache manager to load host disk information on demand. For now now resync from changes.
add disk size constraint and its provider to use disksize label to verify against
add post remove of host listener
add post purge of instance listener
add integration test cases
fix a timing issue in test_host.py

@alena1108 @prachidamle @cjellick


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
```diff
@@ -0,0 +1,63 @@
...
+        List<InstanceHostMap> instanceHostMappings = objectManager.find(InstanceHostMap.class,
+                INSTANCE_HOST_MAP.INSTANCE_ID,
+                instance.getId());
+
+        if (instanceHostMappings.size() == 0) {
+            return null;
+        }
+        InstanceHostMap mapping = instanceHostMappings.get(0);
+        CacheManager cm = CacheManager.getCacheManagerInstance(this.objectManager);
+        HostInfo hostInfo = cm.getHostInfo(mapping.getHostId());
```
alena1108: Just to be on the safe side, you should check if the hostInfo is not null here.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
```diff
@@ -0,0 +1,63 @@
+package io.cattle.platform.servicediscovery.process;
```
alena1108: Any reason why we do deallocation on Purge, not Remove? Also when allocation happens - on create or on Start? If on Start, then the deallocation should happen on Stop.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
jimengliu: currently, cache manager will always allocate a new HostInfo object if not existing. I am thinking to add additional param for this getHostInfo() method to let calling code decide it needs more info of host(such as disk info) or not to reduce some unnecessary code.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+        return true;
+    }
+
+    public boolean checkDiskSize (CacheManager cm, HostInfo hostInfo, Long instanceId) {
+
+        for (Entry<String, DiskInfo> entry : hostInfo.getAllDiskInfo()) {
+            DiskInfo diskInfo = entry.getValue();
+            Long capacity = entry.getValue().getCapacity();
```
prachidamle: nit pick: you can use the diskInfo object instead of evaluating entry.getValue() again

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
jimengliu: allocation happens on instance start for sure. But I do see instance releasing computing resource  are called during purge path and that is where I plugged disk space release in the purge path.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
jimengliu: true.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
alena1108: @prachidamle Prachi, what are the reasons for this behavior?

alena1108:
@jimengliu we should also have tests for the following:

* container lifecycle - to make sure disk is allocated when instance is created, and released when it's removed
* multiple containers and disk allocation. Example: start 2 containers fitting into current available disk size, try to start 3rd, make sure it fails. Remove one of the first 2 containers, try to start a new one, make sure it can start.

jimengliu:
Alena, I will add these 2 tests.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/CacheManager.java
```diff
@@ -0,0 +1,112 @@
...
+import io.cattle.platform.util.type.CollectionUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CacheManager
```
cjellick: Seems like the methods of this class that do the getting/setting of hosts and instances in the cache should be synchronized.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
```diff
@@ -0,0 +1,96 @@
...
+            // container case. That means free size will subtract all reserved
+            // size by other volumes as well beside allocated size.
+            InstanceDiskReserveInfo diskReserved = instanceInfo.getDiskReserveInfo(((DiskInfo)diskInfo).getDiskDevicePath());
+            if (diskReserved != null) {
+                freeSize -= diskReserved.getReservedSize();
+            }
+            if (freeSize >= this.reserveSize) {
+                // reserve disk for this instance
+                if (diskReserved != null) {
+                    diskReserved.addReservedSize(this.reserveSize);
```
prachidamle: The class InstanceDiskReserveInfo tracks for a given diskDevicePath, the size reserved for a particular volume by its VolumeName, as below:

+public class InstanceDiskReserveInfo {
 +
 +    private String diskDevicePath;
 +    private Long reservedSize;
 +    private String volumeName;
 +    private boolean allocated;
 +

So for multiple volumes, there should be multiple instances of this class created each tracking one volume

But here it seems like we just increased the reservedSize of an existing InstanceDiskReserveInfo() tracking some existing volume, instead of creating another InstanceDiskReserveInfo() ?

Wont this cause volumes V1 and V2 have a single InstanceDiskReserveInfo having volumenName = V1 or V2 but the reservedSize = addition of both volume sizes?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
```diff
@@ -0,0 +1,96 @@
...
+        this.reserveSize = Long.parseLong(sizeWithUnit.replaceAll("[^0-9]", ""));
+        this.objectManager = objMgr;
+    }
+
+    @Override
+    public boolean matches(AllocationAttempt attempt, AllocationCandidate candidate) {
+        Set<Long> hostIds = candidate.getHosts();
+
+        // if one of the host does not have enough free space then return false
+        for (Long hostId : hostIds) {
```
cjellick: Do you need the oneGood variable?
Seems like this logic can be simplified to 
```
           CacheManager cm = CacheManager.getCacheManagerInstance(this.objectManager);
            HostInfo hostInfo = cm.getHostInfo(hostId);

             // we will get a bunch of disks for that host and we need at least one disk with
             // free space large enough
             // if no disk with big enough free space for this host, then
             // candidate is no good
             if (!checkDiskSize(cm, hostInfo, attempt.getInstanceId())) {
                 log.debug("Scheduling instance [{}] to host [{}] rejected", attempt.getInstanceId(), hostId);
                 return false;
             }
```

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
```diff
@@ -0,0 +1,96 @@
...
+            boolean oneGood = false;
+
+            CacheManager cm = CacheManager.getCacheManagerInstance(this.objectManager);
```
cjellick: You can get the cm outside the loop

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
jimengliu: Right, volumeName should not be 1:1 to InstanceDiskReserveInfo mapping. I will remove this from InstanceDiskReserveInfo class but add another map for the InstanceInfo class and tracking 1:1 mapping for that volume.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
jimengliu: Sure.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
jimengliu: sure.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
```diff
@@ -0,0 +1,96 @@
...
+            if (diskReserved != null) {
+                freeSize -= diskReserved.getReservedSize();
+            }
+            if (freeSize >= this.reserveSize) {
+                // reserve disk for this instance
+                if (diskReserved != null) {
+                    diskReserved.addReservedSize(this.reserveSize);
+                } else {
+                    diskReserved = new InstanceDiskReserveInfo(((DiskInfo)diskInfo).getDiskDevicePath(), this.reserveSize, this.volumeName);
+                    instanceInfo.reserveDisk(diskReserved);
```
cjellick: I have a couple concerns about this logic:

##### Concern 1: You might have a memory leak
You are modifying the data in the cache here, right?
Let's say you have two candidate hosts: A and B. They both have enough free diskspae, so would they both get a cache entry reserving the disk space?

My understanding is that you use the reservation entry in the cache to then update the amount of available disk in allocatorDao.modifyDisk(). So, the fact that you have these bogus entries for every host in the environment doesn't affect the calculation, BUT you _only_ remove entries for where the host to which the instance was _actually_ scheduled (in InstancePurgePostListener), right?

So, you aren't cleaning up all entries in the cache for an instance when it is purged and you have a memory leak, right?


##### Concern 2: You are using the cache to temporarily persist state in memory so that it can be easily accessed later
The InstanceDiskReserveInfo entry's current purpose is to be accessible by allocatorDao.modifyDisk() so that that method can update its cached tally of disk space, right?

I don't think storing this data in memory is a good design. As evidence by concern #1, it can lead to things like memory leaks. We've tried to avoid this in cattle.

So, I think:
* AllocatorDao.modifyDisk() should just recalculate the disk needed when its called as opposed to pulling an entry from the cache manager. You could use a shared util method to ensure both places are calculating the same.
* The _only_ thing that should be stored in the cache is the running tally of available disk space for each host.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
prachidamle: To me, the deallocation should happen on Stop, since the InstanceStart is allocating. 
I will defer to @ibuildthecloud for why do we deallocate on Purge - but not on Instance Stop or maybe on Remove

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
prachidamle: While calculating the freeSize of the disk, we are considering the reservedSize:
+            InstanceDiskReserveInfo diskReserved = instanceInfo.getDiskReserveInfo(((DiskInfo)diskInfo).getDiskDevicePath());
 +            if (diskReserved != null) {
 +                freeSize -= diskReserved.getReservedSize();
 +            }

Here shouldn't we check whether the volume is "reserved" Vs "allocated" using the boolean InstanceDiskReserveInfo. allocated?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/CacheManager.java
jimengliu: Do you mean the map data structure or synchronize the block of access ?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -170,6 +176,39 @@ protected void modifyCompute(long hostId, Instance instance, boolean add) {
...
+
+        for (Entry<String, InstanceDiskReserveInfo> entry : instanceInfo.getAllReservedDisksInfo()) {
+            String diskDevicePath = entry.getKey();
+            InstanceDiskReserveInfo reserveInfo = entry.getValue();
+            Long reserveSize = reserveInfo.getReservedSize();
+
+            // figure out available size
+            DiskInfo diskInfo = cm.getDiskInfoForHost(hostId, diskDevicePath);
+            Long allocated = diskInfo.getAllocatedSize();
+            if (add && (diskInfo.getCapacity() - allocated >= reserveSize)) {
```
cjellick: What should happen when its add && capacity - allocated > reserved?

The disk is going to get used, so I don't think we can ignore it. I think you should raise an error.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -170,6 +176,39 @@ protected void modifyCompute(long hostId, Instance instance, boolean add) {
...
+                diskInfo.addAllocatedSize(reserveSize);
+                log.info("allocated disk space on disk [{}] with total = {}, {} {} {} = {} as used",
+                        diskInfo.getDiskDevicePath(), diskInfo.getCapacity(), allocated, "+", reserveSize,
+                        allocated + reserveSize);
+
+                // mark reserved info for this instance
+                reserveInfo.setAllocated(true);
+            } else if (!add && reserveInfo.isAllocated()
+                    && allocated >= reserveSize) {
```
cjellick: Same thing here: what happens when !add and allocated < reservedSize?

We should _probably_ raise an error, but I'd also consider just setting allocated back to zero. Thoughts on which is better?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
prachidamle: I have few more concerns around the same logic and also some questions for @cjellick or @ibuildthecloud about the allocation framework.

**Scenario 1) ** Consider we are allocating an Instance which has multiple constraints - one of the constraints is the diskConstraint. Say other constraint is affinityConstraint. Suppose the diskConstraint passes, but say the affinityConstraint fails. Now:

Issue 1)  What should happen to the "reserved" disksize? Is it cleared out? 
Issue 2)  Do we have retries of the whole allocate() logic when one allocation try fails due to some constraint failure? @cjellick @ibuildthecloud ?
If there are retries, will the DiskConstraint that had passed previously, reserve the disk space again?

**Scenario 2)** @cjellick @ibuildthecloud  Can we have multiple Instances getting allocated simultaneously? I see that we lock at the start of by using 
lockManager.lock(new AllocateResourceLock(request), new LockCallback<Boolean>() {..}
But what does it lock on? 

I am asking because the logic above only considers the multiple volumes of the same instance that is being allocated, to calculate the "freeSize" of a given disk.
Don't we need to consider volumes of other instances that may have "reserved" space on the same disk?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
jimengliu: Do you mean add && capacity - allocated < reserved ?If so, it won't happen because the reservation process ensure that to fail the constraint.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -170,6 +176,39 @@ protected void modifyCompute(long hostId, Instance instance, boolean add) {
         objectManager.persist(host);
     }
 
+    protected void modifyDisk(long hostId, Instance instance, boolean add) {
+        CacheManager cm = CacheManager.getCacheManagerInstance(this.objectManager);
+        InstanceInfo instanceInfo = cm.getInstanceInfoForHost(hostId, instance.getId());
+
+        for (Entry<String, InstanceDiskReserveInfo> entry : instanceInfo.getAllReservedDisksInfo()) {
```
cjellick: This comment applies to everything within this for loop. We do multiple mutations to the cache in this loop. Since this group of mutations as a whole is not atomic, we run the risk of data corruption in our cache.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/CacheManager.java
cjellick: Note to self: synchronizing on the methods herein would introduce a global block (across all environments).

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -170,6 +176,39 @@ protected void modifyCompute(long hostId, Instance instance, boolean add) {
...
+            Long allocated = diskInfo.getAllocatedSize();
+            if (add && (diskInfo.getCapacity() - allocated >= reserveSize)) {
+                diskInfo.addAllocatedSize(reserveSize);
+                log.info("allocated disk space on disk [{}] with total = {}, {} {} {} = {} as used",
+                        diskInfo.getDiskDevicePath(), diskInfo.getCapacity(), allocated, "+", reserveSize,
+                        allocated + reserveSize);
+
+                // mark reserved info for this instance
+                reserveInfo.setAllocated(true);
+            } else if (!add && reserveInfo.isAllocated()
```
cjellick: I see here that you rely on some state stored in the cache here. This will complicate my previous suggestion to basically get rid of storing the instance data in the cache.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
jimengliu: @cjellick I heard your concern. Currently the implementation of cache manager could have leak in the cases that when we reserved but due to other constraint failed as Prachi pointed out, then InstanceDiskReserveInfo may leak. However, since we need to preserve the reservation disk size for the same disk for the other volume in case of 2 labels are passed in for one instance. In which case, we have 2 disksize constraints for one instance. I think we can't just calculate on the spot in AllocatorDao.modifyDisk(), correct me if I am wrong. 

We need some information accumulated to this point to figure out the total disksize for this instance for this specific disk. That accumulated info could live in the cache. But as you guys suggested, we should have deleted them(leaks) if that instance is not eventually allocated to that host in order to save memory. That could be done by making InstanceInfo not part of HostInfo and after allocation succeeds(AllocatorDao.modifyDisk()), we clean up InstanceDiskReserveInfo in InstanceInfo object by replacing all disksize reservation entries using final allocated disksize for that disk on that host.

Would that work ? @cjellick @prachidamle

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
jimengliu: Is this concern the same as synchronize the Cache Manager ? Everything in CM I meant ?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
cjellick: We used to have the logic and removed ~7 weeks ago
https://github.com/rancher/cattle/pull/1513

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
cjellick: @prachidamle allocation logic is serialized so that one instance is allocated at a time. This _should_ be scoped to a per-project basis but is actually currently global.

cjellick: @prachidamle if allocation completely errors out because no valid candidate is found, then that instance is considered completely failed. However, if that instance is part of a service, I believe we will keep trying as part of the service logic.

cjellick: @jimengliu These allocation constraints should really be tied to the allocation of the volume and not the instance, but I think that is out of scope for this round of the implementation.

Why even have more than one constraint? What if in the constraint provider you summed up all the disk space labels and had a single constraint for them? If you like, you wouldn't even need to sum the constraints in the provider. You could create a  list that contains each disk.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
cjellick: No, not really. It was more along the lines of the fact that the mutations to the cache are not in a single transaction, so you could _potentially_ get into a state where not all of the changes completed. BUT I'm realizing now that really the only way for that to happen is if cattle crashed and if cattle crashed, then we'd be starting from scratch anyway.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
jimengliu: Yes, I am planning to do single constraint implementation and put all the volumes into that constraint. This way, there is   no cache for InstanceDiskReserveInfo during scheduling process. Only when modification of disk happens, we will cache only total disk size really allocated for each disk of that instance(no volume info needed then) for fast deallocation purpose.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/CacheManager.java
cjellick: The methods on the class, ie:
```
public synchronized HostInfo loadHostInfoToCache(Long hostId) {
...
}

public syncronized HostInfo getHostInfo(Long hostId) {
...
}
```

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
```diff
@@ -0,0 +1,63 @@
...
+        Instance instance = (Instance) state.getResource();
+
+        List<InstanceHostMap> instanceHostMappings = objectManager.find(InstanceHostMap.class,
+                INSTANCE_HOST_MAP.INSTANCE_ID,
+                instance.getId());
+
+        if (instanceHostMappings.size() == 0) {
+            return null;
+        }
+        InstanceHostMap mapping = instanceHostMappings.get(0);
```
cjellick: Why just the first one? I realize that logically there will only be one, but I think you should just throw this in a loop and do it for each InstanceHostMap that is returned.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/CacheManager.java
jimengliu: ok, using synchronized methods to do global lock of the critical section.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
```diff
@@ -0,0 +1,63 @@
...
+        CacheManager cm = CacheManager.getCacheManagerInstance(this.objectManager);
+        HostInfo hostInfo = cm.getHostInfo(mapping.getHostId());
+
+        hostInfo.removeInstance(mapping.getInstanceId());
+        log.debug("removed instance [{}] info from host [{}] info in cache manager", instance.getId(),
```
cjellick: Most of the time (when there is not entry), you will not have deleted anything. Can you only log if you actually removed something?

jimengliu: if not in a loop, it will remove the current instance got purged, so it only logs once, right ?

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/InstancePurgePostListener.java
cjellick: About that, I don't think that is good behavior. That puts an entry in the cache for every single container or vm that is ever allocated, regardless of whether or not it actually uses the disk space constraint.

We should contain this logic so that the only entries in the cache ones for VMs that are actually using the disk space constaint.

cjellick:
Note: further cattle changes are required to add the disk space label to an instance during VirtualMachinePreCreate. But I think we discussed this and it was agreed that I would do it.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/constraint/DiskSizeConstraint.java
cjellick: @jimeng: that plan sounds good.

jimengliu:
@cjellick @prachidamle @alena1108 Please take a look at updated algorithm using an util function that takes care of calculation of the allocation from multiple disks to multiple volumes, some previous code review fixes and add one more disk space life cycle check test case.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -1,28 +1,33 @@
 package io.cattle.platform.allocator.dao.impl;
 
-import static io.cattle.platform.core.model.tables.AgentTable.*;
```
cjellick: I dont want to merge with this formatting change. Can you change it back? There is a setting in eclipse that controls how many static imports will cause a .* can you change that setting to 1?

Preferences > Java > Organize Imports > Number of imports needed for .*

jimengliu: sure. I will push one more change.

PR #1658, opened 2016-05-09
Ingress-controller to k8s template

alena1108:
WIP till rancher/ingress-controller:0.1.0 is released


alena1108:
Image is released, so the PR is good to merge

PR #1669, opened 2016-05-13
rename scheduler disksize label prefix string

jimengliu:
@cjellick 
rename scheduler disksize label prefix string to "io.rancher.resource.disksize."


cjellick:
LGTM

PR #1682, opened 2016-05-17
Update startCount on instancehostmap.activate

alena1108:
it would ensure that the instance got started on the host

https://github.com/rancher/rancher/issues/4833


cjellick:
LGTM

PR #1696, opened 2016-05-20
Healthcheck config item: update on instance.stop/start

alena1108:
instead of nic.activate/deactivate. Because instance is reset to
initializing/reinitializing was moved from nic.activate/deactivate to instance.start/stop


alena1108:
https://github.com/rancher/rancher/issues/4872

PR #1697, opened 2016-05-20
WIP Deploy kubernetes from catalog

alena1108:
To make upgrades possible.

@ibuildthecloud tested stack upgrade for k8s (using rancher-compose), it works. WIP on this PR as corresponding catalog [item](https://github.com/rancher/rancher-catalog/pull/150) has to be merged + k8s 1.2.4 image released (branch 1.2.4 containing rancher changes is already pushed to rancher/kubernetes)


alena1108:
@vincent99 we might need a UI fix as k8s stack external_id is ```system-catalog://library:kubernetes:0``` instead of ```system://kubernetes```. On my local setup I'm seeing UI being stuck in "Creating Kubernetes system stack" although the stack is active and healthy. Also if we solely switch to new format, upgraded setups might get broken. So we should either consider doing db migration, or UI supporting both external ids formats for some time (first way is cleaner of course, but might require putting older version of k8s system catalog entry just for reference purpose)

vincent99:
Yes, system://{swarm,mesos,kubernetes} are used to find the right system stack to watch, I can update that to support both.

alena1108:
https://github.com/rancher/rancher/issues/4896

alena1108:
@vincent99 we should also let user override "library" and "community" repos' urls via UI to the private repos when Rancher is deployed in no internet mode: https://github.com/rancher/rancher/issues/4970

PR #1713, opened 2016-05-31
schedule instances using Go micro-service scheduler

jimengliu:
- This is the cattle side client code changes launching and calling Go scheduler in another process.
- It uses current allocating logic to call scheduler for allocating CPU/Memory/IOPS before all other constraints get matched.
- added resource allocation tests in python and modified existing tests to pass by giving each simulated host enough CPU and Memory.

@cjellick @ibuildthecloud


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AbstractAllocator.java
```diff
@@ -228,13 +239,58 @@ protected boolean doAllocate(final AllocationRequest request, final AllocationAt
...
+                    for (AllocationCandidate candidate : candidatesForCPUMemoryIops) {
+                        boolean good = true;
+                        
+                        // schedule cpu, memory and iops first
+                        try {
+                            for( Long hostId : candidate.getHosts() ){
+
+                                // in case we already tried to schedule instance for this hostId, then skip it
+                                if (alreadyScheduledHostIds.contains(hostId)) {
+                                    continue;
```
cjellick: I think this might need to change this logic.

I hit a bug where two candidates containing the same hostId were passed in. They should not have been allowed to schedule to that I was assigning them to because the host didnt have fio data and the iops labels were set.

For the first host, the scheduler reported no and we hit the continue on line 286. Then the next candidate went. good was set back to true. then we hit the alreadyScheduledHostIds.contains(hostId) logic and it broke out of the loop without ever setting good to false. So, the host was used

cjellick: I was able to easily reproduce this by creating a host that did not have fio ran on it and then going to the UI and creating a container on that host with the label `io.rancher.resource.read_iops.foo` with a value of `100` to it in the UI.

The container successfully deployed to that host but it should not have been allowed to. When I commented out this if/else block, it failed to schedule, as expected.

jimengliu: nice catch! I will fix it by preserving the the good variable of already scheduled hostId.

code/iaas/allocator/pom.xml
```diff
@@ -28,5 +28,10 @@
             <artifactId>cattle-framework-jooq</artifactId>
```
alena1108: I suspect that there might be circular reference dependencies. cattle-iaas-api-logic uses cattle-iaas-logic, and cattle-iaas-logic uses allocator:

```
code/iaas/api-logic/pom.xml:    <artifactId>cattle-iaas-api-logic</artifactId>
code/iaas/logic/pom.xml:            <artifactId>cattle-iaas-allocator</artifactId>
```
Can you please run the check in the eclipse

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -225,6 +236,35 @@ protected void modifyDisk(long hostId, Instance instance, boolean add) {
...
+            int statusCode = response.getStatusLine().getStatusCode();
+            if (statusCode != 200) {
+                log.error("statusCode: {}", statusCode);
+            }
+        } catch (HttpHostConnectException ex) {
+            log.error("Scheduler Service not reachable at [{}]", ALLOCATE_CPU_MEMORY_URL);
+        } catch (IOException e) {
+            log.error((e.getStackTrace()).toString(), e);
+        }
+
```
alena1108: Should we throw resource allocation exception when status code != null or when there it an http exception?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -225,6 +236,35 @@ protected void modifyDisk(long hostId, Instance instance, boolean add) {
 
     }
 
+    @Inject
+    IdFormatter idFormatter;
+
+    public void addReleaseSchedulerResource(String action, Long instanceId, boolean isVM, Long hostId, Long envId) {
```
alena1108: Can you check if we hold the lock in the caller stack? Or if the lock is held in your scheduler microservice, it is fine too

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/HostRemovePostListener.java
```diff
@@ -21,10 +41,39 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+                log.error("statusCode: {}", statusCode);
+            }
+
+        } catch (HttpHostConnectException ex) {
+            log.error("Scheduler Service not reachable at [{}]", ALLOCATE_CPU_MEMORY_URL);
+        } catch (IOException e) {
+            log.error((e.getStackTrace()).toString(), e);
+        }
+
+        return;
```
alena1108: Again, when status code != 200 or an exception thrown, should we error out here? Does the microservice return any particular error code when host no longer exists for the environment? if so, we should just ignore that error code, and for all the rest handle the exception

code/iaas/allocator/pom.xml
jimengliu: How? no idea how I can do that in Eclipse, can you show me? I checked the pom file, it does not seem so. code/iaas/allocator/pom.xml contains code/iaas/api-logic/pom.xml as dependency, but code/iaas/api-logic/pom.xml does not have any dependency that could lead back to code/iaas/allocator/pom.xml.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
jimengliu: Are you saying if statusCode != 200, we should throw exception ? Yes, if there is a http or I/O exception, we log them as well. I think this is best effort operation.

PR #1715, opened 2016-06-01
Cleanup only unehalthy instance, not the entire unit

alena1108:
Also cleanup unhealthy instaces first, before recreating them

https://github.com/rancher/rancher/issues/4311

https://github.com/rancher/rancher/issues/4894


alena1108:
Closing the PR as destroying the entire unit when one of the components fail, is the correct behavior.

PR #1724, opened 2016-06-07
Omit error state for network agent only

alena1108:
WIP as I want to examine test results to make sure nothing is broken


alena1108:
Can be merged

PR #1726, opened 2016-06-07
Changes to add missing docker-compose parameters in the exported config

prachidamle:
Changes to add missing docker-compose parameters pid, device, memswap_limit on exported config
Also added tests for the parameters

https://github.com/rancher/rancher/issues/4890
https://github.com/rancher/rancher/issues/4232
https://github.com/rancher/rancher/issues/5001

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/resource/ServiceDiscoveryConfigItem.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/resource/ServiceDiscoveryConfigItem.java
@@ -58,6 +58,11 @@ public class ServiceDiscoveryConfigItem {
 
     public static final ServiceDiscoveryConfigItem LABELS = new ServiceDiscoveryConfigItem(
             InstanceConstants.FIELD_LABELS, InstanceConstants.FIELD_LABELS);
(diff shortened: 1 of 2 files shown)


prachidamle:
My tests passed [gw0] PASSED core/test_svc_discovery.py::test_export_config

Seems to be some other failure, @alena1108 can you restart the tests

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/resource/ServiceDiscoveryConfigItem.java
```diff
@@ -58,6 +58,9 @@
 
     public static final ServiceDiscoveryConfigItem LABELS = new ServiceDiscoveryConfigItem(
             InstanceConstants.FIELD_LABELS, InstanceConstants.FIELD_LABELS);
+    public static final ServiceDiscoveryConfigItem MEMSWAPLIMIT = new ServiceDiscoveryConfigItem("memorySwap", "memswap_limit");
```
alena1108: Can you change this param to be read from DockerInstanceConstants.java? And the params below should also be read from constants. No need to convert to "_", underlying code will do it for you.

prachidamle: Moving the params to DockerInstanceConstants , but the "_" conversion is probably not applicable here.

PR #1728, opened 2016-06-08
iops API schedma change

jimengliu:
adding iops read/write API changes

@cjellick

### DIFF
--- a/resources/content/schema/base/container.json
+++ b/resources/content/schema/base/container.json
@@ -235,6 +235,14 @@
          "startCount" : {
             "type" : "int",
             "nullable" : true
+        },
+        "blkioDeviceReadIOps": {
+            "type": "array[map[string]]",
+            "nullable": true
+        },
+        "blkioDeviceWriteIOps": {
+            "type": "array[map[string]]",
+            "nullable": true
         }
     },
     "resourceActions": {
--- a/resources/content/schema/admin/admin-auth.json
+++ b/resources/content/schema/admin/admin-auth.json
@@ -15,6 +15,9 @@
         "account.projectId" : "r",
         "account.identity" : "r",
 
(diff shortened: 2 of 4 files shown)


resources/content/schema/base/container.json
```diff
@@ -235,6 +235,14 @@
          "startCount" : {
             "type" : "int",
             "nullable" : true
+        },
+        "blkioDeviceReadIOps": {
+            "type": "array[map[string]]",
```
cjellick: they type of the array shouldnt be a map. it should be an explicit object of type iops. 

You can follow the haConfig option as an example on how to do this. You need:
- an iops.josn file in the same directory as the haConfig.json file
- an entry in spring-core-model-context.xml, same place where haConfig is defined
- add permissions in user-auth.json
- change the type from map to this concrete type

jimengliu: Docker define this way array of map. Should we keep this way or create our own way and then convert it back to docker way at python agent ?

cjellick:
It would be best if you added a test for this here as well, to ensure the round trip from cattle to docker and back is working.

You can add it to test_docker.py's test_container_fields test.
Now, this test will fail in CI until we merge python-agent changes and update python-agent in cattle. We can do that as part of this PR so that the whole thing works when we merge it.

You should get the test_container_fields working locally before we merge the python-agent PR. Then we'll merge python-agent, I'll do a release of py-agent, then in this PR we will update python-agent.

cjellick:
To be complete, you actually need to add logic for this field in two other places as well:
- DockerTransformerImpl - this lets us backpopulate the field when we get it from docker
- ServiceDiscoveryConfigItem - this lets us translate from cattle syntax to docker-compose syntax

Hopefully the pattern in each is easy enough to follow.

PR #1731, opened 2016-06-09
Bug fixes around host removal and agent.deactivate/remove

alena1108:
This PR fixes a couple of bugs:

1) Host was removed, but agent was left around non-removed. Happened because we had agent.disconnect scheduled from 2 places: 

a) HostRemove handler
b) AgentResourceRemove post handler (triggered by host.remove process) - disconnect is scheduled with remove process chained. The bug was - disconnect/remove process was canceled at this point because a) already put agent to "disconnecting" state. So the agent was never removed as a result.

2) Agent reconnect process (AgentActivate handler) was waiting for the ping event when agent.disconnect was called. So we ended up having agent stuck in Disconnecting state because we were dropping events. It only timed out/got cancelled after ~5 mins, and Disconnect process was able to complete. The fix here would be - include deactivating/inactive/removing/removed/purged/purging to the list of states triggering AgentRemovedException, so Reconnecting process completes faster for the agent in these states, and Deactivate process can go through.

@ibuildthecloud let me know if the above sounds valid

https://github.com/rancher/rancher/issues/5047

### DIFF
--- a/code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/service/impl/AgentServiceImpl.java
+++ b/code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/service/impl/AgentServiceImpl.java
@@ -11,6 +11,7 @@ import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.model.Agent;
 import io.cattle.platform.eventing.EventProgress;
 import io.cattle.platform.eventing.EventService;
+import io.cattle.platform.eventing.exception.AgentInactiveException;
 import io.cattle.platform.eventing.exception.AgentRemovedException;
 import io.cattle.platform.eventing.exception.EventExecutionException;
 import io.cattle.platform.eventing.model.Event;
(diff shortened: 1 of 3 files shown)


code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/service/impl/AgentServiceImpl.java
```diff
@@ -39,6 +39,11 @@
     public static final Set<String> GOOD_AGENT_STATES = new HashSet<String>(Arrays.asList(CommonStatesConstants.ACTIVATING, CommonStatesConstants.ACTIVE,
             AgentConstants.STATE_RECONNECTING));
 
+    public static final Set<String> REMOVED_AGENT_STATES = new HashSet<String>(Arrays.asList(
+            CommonStatesConstants.REMOVED, CommonStatesConstants.REMOVING,
+            CommonStatesConstants.DEACTIVATING, CommonStatesConstants.INACTIVE, CommonStatesConstants.PURGED,
```
ibuildthecloud: DEACTIVATING and INACTIVE make me nervous.  Are those essential to this fix?

ibuildthecloud: What if for those two state we throw a different exception.  Like AgentInactiveException.

code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostRemove.java
```diff
@@ -31,11 +31,6 @@ public HandlerResult handle(final ProcessState state, ProcessInstance process) {
             return null;
         }
 
-        try {
```
ibuildthecloud: Sorry, I don't understand this change.

ibuildthecloud: Nevermind, I get it now.  It's redundant because of AgentResourceRemove

code/iaas/agent-server/src/main/java/io/cattle/platform/agent/server/service/impl/AgentServiceImpl.java
alena1108: Not fixing it will result in the host being stuck in deactivating state for longer as we are dropping agent events if the agent is in this state. I can introduce a new exception.

alena1108:
@ibuildthecloud addressed your comment by throwing AgentInactiveException when agent is in Deactivating/Inactive states

PR #1732, opened 2016-06-09
Bump up agent instance to v0.8.2

cloudnautique:
This is a rebuild of v0.8.1 with a new openssl package that fixes a vulnerability reported in: rancher/rancher#5038


alena1108:
LGTM

PR #1737, opened 2016-06-10
Stack healthState=active when there are 0 services

alena1108:
https://github.com/rancher/rancher/issues/5048


alena1108:
TF is in test_container_ha_default, seems unrelated.

PR #1742, opened 2016-06-14
Moved for volumes/networkFrom logic

alena1108:
from service code to on per instance code basis.

https://github.com/rancher/rancher/issues/5100

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
@@ -19,10 +19,14 @@ import io.cattle.platform.core.model.IpAddress;
 import io.cattle.platform.core.model.Nic;
 import io.cattle.platform.core.model.Port;
 import io.cattle.platform.core.model.Volume;
+import io.cattle.platform.core.util.SystemLabels;
+import io.cattle.platform.docker.constants.DockerInstanceConstants;
 import io.cattle.platform.engine.handler.HandlerResult;
 import io.cattle.platform.engine.process.ProcessInstance;
 import io.cattle.platform.engine.process.ProcessState;
 import io.cattle.platform.json.JsonMapper;
+import io.cattle.platform.object.resource.ResourceMonitor;
+import io.cattle.platform.object.resource.ResourcePredicate;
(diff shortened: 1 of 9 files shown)


ibuildthecloud:
@alena1108 How long do we wait and should we ever fail?  There are situations in which it will never finish like if the dependent container is stopped, removed, etc.   Seems like we should handle this in some way.  For example what if the dependent container is not transitioning and not running, we then fail start.

alena1108:
@ibuildthecloud the default wait it 15 seconds. For services containers, if the dependent container gets stopped/removed, reconcile will be triggered, and the situation will be fixed by either restarting the stopped dependent, or recreating the entire unit, by the deployment manager.

For standalone container, no such handling is being done. Before this PR, the dependent container being in stopped/removed state is just being ignored. To fix the situation - and handle it for both standalone and service containers in similar way - we can:

Throw an executionException when the dependent is in Removed/Stopped state. With that, thinking if I should move this code to the InstanceStart handler where ExecutionException is being handled (by putting an instance-to-start to either Removed/Stopped/or Error state)

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
```diff
@@ -117,6 +129,58 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+            instancesIds.add(networkFromId);
+        }
+        List<Instance> waitList = new ArrayList<>();
+        for (Long id : instancesIds) {
+            Instance i = objectManager.loadResource(Instance.class, id);
+            List<String> removedStates = Arrays.asList(CommonStatesConstants.REMOVED, CommonStatesConstants.REMOVING);
+            List<String> stoppedStates = Arrays.asList(InstanceConstants.STATE_STOPPED, InstanceConstants.STATE_STOPPING);
+            String type = networkFromId != null && networkFromId.equals(id) ? "networkFrom" : "volumeFrom";
+            if (removedStates.contains(i.getState())) {
+                throw new ExecutionException("Dependencies readiness error ", type + " instance is removed", i);
```
alena1108: @ibuildthecloud throwing executionException as situation when data/network container is not present == docker error when non-existing container id is passed in. Let me know if you think there should be some other exception.

PR #1746, opened 2016-06-14
Back populate labels for all containers

cjellick:


cjellick:
I'm afraid to do full back population for non-native containers on what I want to be a quick turnaround. So, I figured I'd start small.

PR #1749, opened 2016-06-14
Access Control: Changes to not allow a user to login when the user is not in "allowed" set

prachidamle:
Changes to not allow a user to login when the user is not in the list of the "allowed" users/groups. Even if the user has been invited to some environment, unless she is part of the "allowed" identities, she cannot
login to rancher.

This is a change in behavior for github primarily since only github had "restricted" access mode so far.

This will apply for "restricted" accessmode of all auth providers.

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
@@ -222,10 +222,9 @@ public abstract class AbstractTokenUtil implements TokenUtil {
 
     @Override
     public boolean isAllowed(List<String> idList, Set<Identity> identities) {
-        boolean hasAccessToAProject = authDao.hasAccessToAnyProject(identities, false, null);
         switch (accessMode()) {
             case "restricted":
-                if (hasAccessToAProject || isWhitelisted(idList)) {
+                if (isWhitelisted(idList)) {
                     break;
                 }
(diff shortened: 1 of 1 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
```diff
@@ -222,10 +222,9 @@ public boolean findAndSetJWT() {
 
     @Override
     public boolean isAllowed(List<String> idList, Set<Identity> identities) {
-        boolean hasAccessToAProject = authDao.hasAccessToAnyProject(identities, false, null);
         switch (accessMode()) {
             case "restricted":
-                if (hasAccessToAProject || isWhitelisted(idList)) {
+                if (isWhitelisted(idList)) {
```
alena1108: is whitelisted implies that the user has an access to any project? or should we do

```
 hasAccessToAProject && isWhitelisted(idList)
```

meaning that the user has to be whitelisted and has an access to at least one project in order to login

instead?

prachidamle: isWhitelisted() implies that the user can login to rancher. Originally even if the user was not in the whitelist, he was alllowed to login if he was invited to some project.
Now we want to let the user login only if he belongs to the whitelist, when "restricted" access is set.

Problem with using hasAccessToAProject && isWhitelisted(idList), would when the user is attempting to login for the very first time, his personal project is not created at this point I guess. So there many not be any project he has access to.

alena1108:
LGTM

PR #1751, opened 2016-06-14
fix issue4774 user input service name is not respected as anti-affini…

jimengliu:
https://github.com/rancher/rancher/issues/4775

original problem: Due to upper case service name without stack name, the stack_service name created inside affinity rule value part does not have the stack name prepended, then later on scheduler constraint match function will fail.

fix: when comparing user input service name with existing service names within the stack, it should not be case sensitive.

@alena1108


alena1108:
@jimengliu can you add a test for this fix?

jimengliu:
Sure.

alena1108:
LGTM

PR #1752, opened 2016-06-14
type=VirtualMachine to rancher-compose

alena1108:
```
foo:
  scale: 1
  type: virtualMachine
```

the param goes to rancher-compose file. Ping if you need it to be a part of docker-compose instead


ibuildthecloud:
@alena1108 what does this print for non-virtualMachine?  The type should not be included then.

PR #1756, opened 2016-06-15
Add event data necessary for voluem restore process

cjellick:
Pass the process id as the identifier for the restore process so that
it can be uniquely identified if the restore is retried. Also use an
explicit volumeName parameter for constructing the volume api URL. This
is necessary because the volume embedded inside backup is where the
backup originated but the target of the restore may be a different
volume.


ibuildthecloud:
LGTM

PR #1761, opened 2016-06-15
Minimum version to the catalog url request

alena1108:
@ibuildthecloud @vincent99 please review. Wasn't sure what min version to put given that we might release dev5, so put v1.0.2. Let me know if you think it should be changed to v1.1.0

### DIFF
--- a/code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
+++ b/code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
@@ -48,6 +48,8 @@ import org.apache.http.conn.HttpHostConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


alena1108:
Changed to read the value from "rancher.server.image" setting.

code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/listener/SystemStackUpdate.java
```diff
@@ -69,6 +69,7 @@
     private static final String STACK_RESOURCE = "/config-content/system-stacks/%s/%s";
     private static final Logger log = LoggerFactory.getLogger(SystemStackUpdate.class);
     private static final String CATALOG_RESOURCE_URL = ArchaiusUtil.getString("system.stack.catalog.url").get();
+    private static final String CATALOG_RESOURCE_VERSION = ArchaiusUtil.getString("rancher.server.image").get();
```
ibuildthecloud: Can you not store the value as a final.  Can you convert both to being properly dynamic

PR #1762, opened 2016-06-16
Delegate exception handling - fixes 500 to 409 error code translation

alena1108:
This PR is to test the fix for 500 Internal error caused by the DB not being translated to 409 conflict. 

ReadWriteApiHandler is registered as a ResourceManagerRequestHandler bean, and it delegates read/writes to ResourceManagerRequestHandler class: 

https://github.com/rancher/cattle/blob/dfab53f789b37217d34400e8143f75e0bcc03c30/code/packaging/app-config/src/main/resources/META-INF/cattle/iaas-api/spring-iaas-api-context.xml#L381

but exception handling wasn't delegated, therefore ResourceManagerRequestHandler.handleException translating 500 to 409 was never invoked. 

PR is WIP till the build finishes

@ibuildthecloud please comment

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java
@@ -35,7 +35,6 @@ import org.jooq.SelectQuery;
 import org.jooq.Table;
 import org.jooq.TableField;
 import org.jooq.exception.DataAccessException;
-import org.jooq.exception.DataChangedException;
 import org.jooq.impl.DSL;
 import org.jooq.impl.DefaultDSLContext;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java
```diff
@@ -377,7 +376,7 @@ public boolean handleException(Throwable t, ApiRequest apiRequest) {
         if (t instanceof ProcessCancelException) {
             log.info("Process cancel", t.getMessage());
             throw new ClientVisibleException(ResponseCodes.CONFLICT);
-        } else if (t instanceof DataAccessException || t instanceof DataChangedException) {
+        } else if (t instanceof DataAccessException) {
```
alena1108: DataChangedException already extends DataAccessException, so that fix of mine was fixing nothing. Reverted it.

PR #1763, opened 2016-06-16
Changes to support "required" accessMode for github and AD

prachidamle:
Adding a new "accessMode" value "required". 

New definitions of accessMode are:

When "accessMode" = "required", then a user can log in only when they are present in the "allowed" set of users/groups.

When "accessMode" = "restricted", then a user can log in if  (they are present in the "allowed" set of users/groups) OR (user has access to some project )

When "accessMode" = "unrestricted", then a valid user can log in anytime

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 import javax.inject.Inject;
 import javax.servlet.http.Cookie;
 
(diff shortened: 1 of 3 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
```diff
@@ -223,11 +223,17 @@ public boolean findAndSetJWT() {
     @Override
     public boolean isAllowed(List<String> idList, Set<Identity> identities) {
         switch (accessMode()) {
-            case "restricted":
+            case "required":
```
alena1108: would it be possible to define all these values as string constants?

prachidamle: yes I fixed that to use constants

alena1108:
LGTM

PR #1719, opened 2016-06-03
Changes to support Azure Active Directory Identity provider

prachidamle:
Cattle changes to add a new Identity Provider to cattle access control framework.

This adds a provider that talks with Microsoft Azure Graph API (REST) for auth and search groups/users functionality.

PR #1765, opened 2016-06-16
Changes to support "groups" in Azure.

prachidamle:
When a user logs in, the groups that she is memberOf are also pulled from Azure. This enables the users to create environments and invite the groups to be member of the environments. This action will give access to all azure users of that group, to that  rancher environment.

https://github.com/rancher/rancher/issues/5161

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureIdentityProvider.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureIdentityProvider.java
@@ -69,7 +69,7 @@ public class AzureIdentityProvider extends AzureConfigurable implements Identity
 
     @Override
     public Set<Identity> getIdentities(Account account) {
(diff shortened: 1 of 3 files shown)


prachidamle:
@alena1108 this one also addresses the comments you had for the original PR for Azure(https://github.com/rancher/cattle/pull/1719)

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureIdentityProvider.java
```diff
@@ -69,7 +69,7 @@
 
     @Override
     public Set<Identity> getIdentities(Account account) {
-        if (!isConfigured() || !AzureConstants.CONFIG.equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get()) ||
+        if (!isConfigured() ||
```
alena1108: could you explain why we removed the second condition?

prachidamle: isConfigured() already checks for this. This was your review comment on the last PR.

PR #1766, opened 2016-06-17
API Schema changes to support "required" access Mode

prachidamle:
To support the "required" accessMode, API schema has to include this new value "required" so that validation works.

https://github.com/rancher/rancher/issues/5166


prachidamle:
@ibuildthecloud @cjellick @alena1108 can someone merge this to unblock @tfiduccia

PR #1767, opened 2016-06-17
Better Error handling for Azure errors

prachidamle:
This is to handle the error codes and descriptions from Azure and present them in API response.

Also any error of type "invalid_grant" will be returned as 401(unauthorized) since Azure does not give a clear 401 error.

https://github.com/rancher/rancher/issues/5164


prachidamle:
@ibuildthecloud @cjellick @alena1108 can you review and merge

PR #1772, opened 2016-06-18
Exception is getting eaten up by the blanket catch(Exception ex)

prachidamle:
The unauthorized error is getting eaten up.

https://github.com/rancher/rancher/issues/5164


prachidamle:
@alena1108 @ibuildthecloud please review and merge

PR #1777, opened 2016-06-22
Dash instead of an underscore in service instance name

alena1108:
@vincent99 @ibuildthecloud 

https://github.com/rancher/rancher/issues/5195


alena1108:
we also have to add an upgrade path for the instances created prior to 1.1

alena1108:
Changed the code to handle instances created prior to upgrade. We won't update existing names, but upon the scale up, new names will be generated using new divider. So you will end up having mix of "-" and "_" in service's instances names:

```
mysql> select id, name, state from instance where removed is null;
+----+---------------+----------+
| id | name          | state    |
+----+---------------+----------+
|  1 | Default_foo_1 | running  |
|  2 | Default_foo_2 | running  |
|  3 | Network Agent | running  |
|  4 | Default_foo_3 | running  |
|  5 | Default-foo-4 | starting |
+----+---------------+----------+
```

nathanxu:
thumb up!

PR #1778, opened 2016-06-22
fix around duplicated internal dns being set

alena1108:
@ibuildthecloud

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
@@ -87,9 +87,14 @@ public class InstancePreCreate extends AbstractObjectProcessLogic implements Pro
                         .as(String.class);
                 if (DockerNetworkConstants.NETWORK_MODE_MANAGED.equals(networkMode)) {
                     List<String> dns = DataAccessor.fieldStringList(instance, DockerInstanceConstants.FIELD_DNS);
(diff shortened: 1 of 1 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -87,9 +87,14 @@ protected void setDns(Instance instance, Map<String, Object> labels, Map<Object,
...
                 if (DockerNetworkConstants.NETWORK_MODE_MANAGED.equals(networkMode)) {
                     List<String> dns = DataAccessor.fieldStringList(instance, DockerInstanceConstants.FIELD_DNS);
-                    dns.add(NetworkConstants.INTERNAL_DNS_IP);
-                    data.put(DockerInstanceConstants.FIELD_DNS, dns);
-                    data.put(InstanceConstants.FIELD_DNS_INTERNAL, Joiner.on(",").join(dns));
+                    if (!dns.contains(NetworkConstants.INTERNAL_DNS_IP)) {
+                        dns.add(NetworkConstants.INTERNAL_DNS_IP);
+                        data.put(DockerInstanceConstants.FIELD_DNS, dns);
+                    }
+                    String dnsInternal = DataAccessor.fieldString(instance, InstanceConstants.FIELD_DNS_INTERNAL);
```
ibuildthecloud: Why don't you just always set DNS internal to the same as DNS?

PR #1791, opened 2016-06-28
Do not propagate instances set for upgrade, to metadata

alena1108:
otherwise it is possible to have 2 instances correlating to the same service_index (old and a new one)

@ibuildthecloud let me know if you think it makes sense

@cloudnautique ^^


cloudnautique:
So this would exclude the stopped containers that were running previously? In the event of a rollback, what happens? The old containers/svc indexes would be used? What happens to create index?

alena1108:
@cloudnautique it would. On a rollback, old container will be propagated to a metadata.

cloudnautique:
@alena1108, The rollback would just show the running containers again. 

Just to be totally clear for my understanding though, the end state would be after an upgrade the "stopped" old containers would not be returned in metadata as part of `/self/service/containers` or anywhere the list of containers is called. The reason being, when they are included, looping through for leader election (done on create_index) we get a container that is not running. Then on a rollback, we delete the containers that were updated and start the previous ones which show up in metadata again.

Is that correct?

alena1108:
@cloudnautique this is correct

PR #1797, opened 2016-06-30
For Azure Group, use the displayName property for the accountName field

prachidamle:
For Azure Group, use the displayName property for the accountName field since the mailNickname property is a uuid.

https://github.com/rancher/rancher/issues/5163

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureRESTClient.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureRESTClient.java
@@ -237,8 +237,14 @@ public class AzureRESTClient extends AzureConfigurable{
 
     public AzureAccountInfo jsonToAzureAccountInfo(Map<String, Object> jsonData) {
         String objectId = ObjectUtils.toString(jsonData.get("objectId"));
+        String objectType = ObjectUtils.toString(jsonData.get("objectType"));
         String userPrincipalName = ObjectUtils.toString(jsonData.get("userPrincipalName"));
         String accountName = ObjectUtils.toString(jsonData.get("mailNickname"));
+
+        if("Group".equalsIgnoreCase(objectType)) {
+            accountName = ObjectUtils.toString(jsonData.get("displayName"));
+        }
+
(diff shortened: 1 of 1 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureRESTClient.java
```diff
@@ -237,16 +237,22 @@ public void refreshAccessToken() {
 
     public AzureAccountInfo jsonToAzureAccountInfo(Map<String, Object> jsonData) {
         String objectId = ObjectUtils.toString(jsonData.get("objectId"));
+        String objectType = ObjectUtils.toString(jsonData.get("objectType"));
```
ibuildthecloud: ObjectUtils.toString() will return an empty string if `obj` is `null`.  Is this really what you want to do?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureRESTClient.java
```diff
@@ -237,16 +237,22 @@ public void refreshAccessToken() {
...
         String userPrincipalName = ObjectUtils.toString(jsonData.get("userPrincipalName"));
         String accountName = ObjectUtils.toString(jsonData.get("mailNickname"));
+
+        if(objectType != null && objectType.equalsIgnoreCase("Group")) {
```
ibuildthecloud: Conditions like these can be simplified by flipping the equals, so
```java
"Group".equalsIgnoreCase(objectType)
```
Then you avoid the NPE or need for a null check.

prachidamle: Yep, knew about it, now changed to the optimum way

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureRESTClient.java
prachidamle: Yes that should be fine, in case objectType is empty, we will fall back to using the uuid property.

PR #1798, opened 2016-07-01
LB refactoring

alena1108:
Cattle changes for lb refactor project: convert all lb legacy rules to the metadata


alena1108:
https://github.com/rancher/lb-controller/pull/15

alena1108:
https://github.com/rancher/rancher/issues/2179

resources/content/cattle-global.properties
```diff
@@ -6,6 +6,7 @@ api.ui.index=//releases.rancher.com/ui/latest
 api.ui.js.url=https://releases.rancher.com/api-ui/1.0.4/ui.min.js
 api.ui.css.url=https://releases.rancher.com/api-ui/1.0.4/ui.min.css
 agent.instance.image=rancher/agent-instance:v0.8.3
+lb.instance.image=rancher/lb-service-haproxy:v0.3.0
```
alena1108: This is the default image if nothing is passed to the API.

alena1108:
Only one remaining question - how are we doing to distribute images for lb services. Don't want to hold it in the global-properties as it is going to tight rancher to lb-controller releases. Would have to find a way to distribute it similar way we do for our system stacks - via catalog. Only in this case we can't expect lb to run in its own dedicated stack, as other items deployed from catalog do.

alena1108:
@vincent99 update on per discussion with @ibuildthecloud:

1) we no longer have a special type for the balancersService; regular service type should be used instead. As identifier for LB/CLI/rancher-compose, label "io.rancher.balancer.service"=true will be used.
2) All lb data should be passed on service create/update a part of service.metadata field, under "lb" key.
3) For UI/CLI/rancher-compose, I've created a helper API type - "balancerServiceMetadata" where all lb fields and their types are defined.

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/LoadBalancerInfoDao.java
```diff
@@ -19,4 +20,8 @@
             List<? extends LoadBalancerListenerInfo> listeners);
 
     List<LoadBalancerTargetInput> getLoadBalancerTargets(Service lbService);
+
+    Map<String, Object> processLBMetadata(Service lbService, LoadBalancerInfoDao lbInfoDao, Map<String, Object> meta);
```
alena1108: Two methods added are not used today, but in the future can be used for converting LB service data to "lb" metadata used by the new microservice

resources/content/schema/base/balancerTargetMetadata.json
```diff
@@ -0,0 +1,11 @@
+{
```
alena1108: this type is for the lb service metadata defined on the target service (when service joins the LB based on selector)

alena1108:
Corresponding catalog PR: https://github.com/rancher/rancher-catalog/pull/208/files

vincent99:
@alena1108 @ibuildthecloud Boo?  Service with a magic label and metadata that happens to match a type isn't discoverable from the API.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/factory/impl/AgentInstanceFactoryImpl.java
```diff
@@ -155,6 +155,27 @@ public Agent createAgent(Instance instance) {
         return null;
     }
 
+    @SuppressWarnings("unchecked")
+    protected boolean isSystemService(Service service) {
+        Stack stack = objectManager.loadResource(Stack.class, service.getStackId());
+        if (DataAccessor.fieldBool(stack, "isSystem")) {
```
ibuildthecloud: Can you use ServiceConstants.isService()

ibuildthecloud: sorry isSystem()

PR #1815, opened 2016-07-14
Do not trim spaces in the ldap DN. 

prachidamle:
Will later add normalizing of DN and more fixes to handle existing saved DNs during upgrade later.


prachidamle:
@ibuildthecloud PR reverting the ldap DN trim change

PR #1818, opened 2016-07-15
allow retainIpUpdate

alena1108:


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServiceUpdatePreListener.java
```diff
@@ -0,0 +1,89 @@
...
+        Subnet subnet = ntwkDao.getSubnetForObject(service, NetworkConstants.KIND_HOSTONLY);
+        if (subnet == null) {
+            return null;
+        }
+
+        // get all service managed instances and their ips/indexes
+        List<IpAddressToServiceIndex> ipToIndex = instanceDao.getIpToIndex(service, subnet);
+        for (IpAddressToServiceIndex i : ipToIndex) {
+            // release from subnet
+            poolManager.releaseResource(subnet, i.getIpAddress());
```
alena1108: @ibuildthecloud race condition prone as released resource can be allocated by some other thread. Appreciate suggestion on how to avoid it

alena1108: In non-retain ip address case, the owner of the ip address is subnet. In retainIp case, the owner is the serviceIndex. Have to release from one to assign to another

PR #1823, opened 2016-07-18
Public endpoints updates via config item

alena1108:


code/iaas/config-item/common/src/main/java/io/cattle/platform/configitem/model/Client.java
```diff
@@ -34,27 +38,37 @@ public Client(ConfigItemStatus status) {
...
-    protected void assignEvent() {
+    protected void assignEvent(String configItemStatusName) {
         if (resourceType == Agent.class) {
             eventName = IaasEvents.CONFIG_UPDATE;
         } else if (resourceType == Account.class) {
             eventName = IaasEvents.GLOBAL_SERVICE_UPDATE;
         } else if (resourceType == Service.class) {
-            eventName = IaasEvents.SERVICE_UPDATE;
+
+            if (StringUtils.equals(SERVICE_ENDPOINTS_UPDATE_ITEM_NAME, configItemStatusName)) {
```
ibuildthecloud: slightly hacky

alena1108: @ibuildthecloud  we map the db field to iaas event today, may be we should assign an event on per config item name instead?

ibuildthecloud: It's fine. Don't worry.

PR #1826, opened 2016-07-20
Allow adding/removing sidekicks during inservice upgrade

alena1108:
yet to finalize the format for sidekick removal. The label aka "remove_me" is used for now


alena1108:
all sidekicks are coming as a part of "secondaryLaunchConfigs" parameter in the upgrade strategy. If sidekick exists on the service, it will get upgraded. If not - it will get added. If we want to remove the sidekick from the service, the sidekick should be defined with the label "io.rancher.launch_config.delete=true" (chatted with @vincent99 about it) and still be passed as a part of secondaryLaunchConfigs. 

We wanted to avoid adding more parameters in v1 api for upgrade. In v2, we will be smarter and do something like:

```
secondaryLaunchConfigs:
 toAdd: {}
 toRemove{}
 toModify{}
```

alena1108:
Also we have 0 API validation for upgrade now (if volumesFrom/networksFrom get broken with adding/removing sidekicks, it will get discovered only further down in the code). I will add the validation later unless you feel it should be added at this point @ibuildthecloud let me know

PR #1827, opened 2016-07-20
make sure dns name is set correctly for kubernetes services

wlan0:
@alena1108

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
@@ -360,11 +360,17 @@ public class DnsInfoDaoImpl extends AbstractJooqDao implements DnsInfoDao {
                 if (StringUtils.isEmpty(targetInstanceName)) {
                     ipToInstanceName.put(targetIp, null);
                 } else {
+                    String dnsName;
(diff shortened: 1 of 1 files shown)


wlan0:
Is required to fix - https://github.com/rancher/rancher/issues/5236

wlan0:
This PR makes the kubernetes service adhere to the format defined under "network identity" subsection of these docs - http://kubernetes.io/docs/user-guide/petset/

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
```diff
@@ -360,11 +360,17 @@ protected void addToResolve(ServiceInstanceData targetInstance, String aliasName
                 if (StringUtils.isEmpty(targetInstanceName)) {
                     ipToInstanceName.put(targetIp, null);
                 } else {
+                    String dnsName;
+                    if (targetInstance.getService().getKind().equalsIgnoreCase("kubernetesservice")) {
```
ibuildthecloud: @alena1108 Can we have an approach that doesn't hard code logic to k8s?

alena1108: @ibuildthecloud as this split just changes fqdn format from instanceName.globalNamespace (followed by regular service instances) to instanceName.serviceName.stackName.globalNamespace (followed by k8s service instances), so we can go with the label approach. The label would define whether svcName.stackName should be included to the individual container resolution.

PR #1828, opened 2016-07-21
add host.update to the default processes

wlan0:
@ibuildthecloud 

When we update host labels in the rancher UI, the change doesn't get updated in the metadata service immediately. This PR is meant to fix that. Tested and verified that `requested_version` and `applied_version` fields are updated immediately in the config_item_status table for these hosts, after changing the host label. Also tested that the metadata is updated immediately after updating labels. 

Fixes - https://github.com/rancher/rancher/issues/5493


alena1108:
LGTM

PR #1845, opened 2016-07-25
wait for startOnce to have startCount>0 on dependencies validation

alena1108:
When validate dependencies start, make sure that the instance with startOnce=true in Stopped state was at least started once


alena1108:
Closing, will resolve as a part of a different PR

PR #1849, opened 2016-07-26
Fix haproxy ssl behavior

cavemandaveman:
When choosing ssl for the load balancer in Rancher, `ssl` should be set on every backend `server` line. Currently it sets a mixed-mode ssl, using frontend ssl connections, but plain http calls to backend servers. See the discussion [here](https://github.com/rancher/rancher/issues/4977).


alena1108:
@cavemandaveman it does look like this fix would offset the ssl offload mode that is being explicitly set when "ssl" check box is set in the UI. In ssl offload mode: 

a) certs defined on the LB, will be uploaded them to haproxy
b) https will be ssl terminated by haproxy (using the uploaded cert)
c) so client to haproxy communication is https, haproxy to backend is http. 

Allowing ssl pass through shouldn't break existing default behavior...so changing the behavior of our "ssl" UI flag won't work unfortunately. The solution would be to add an extra flag to the service creation enabling ssl pass through mode, but at this point we are hesitant about changing our APIs because lb-refactor work is in progress. With the refactoring, you will be able to set any backend configuration of your choice.

cavemandaveman:
@alena1108 Even with the LB refactoring, it might be useful to have default options for SSL Termination as well as SSL Backend Passthrough. I might try to hack at the UI portion of Rancher to add this capability...

alena1108:
@cavemandaveman sure, we'd appreciate any PRs for the refactored lb once it is released. The repo will be: https://github.com/rancher/lb-controller. 

Closing this PR for now

PR #1851, opened 2016-07-27
add externalStackEvent

wlan0:
@alena1108


code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalStackEventCreate.java
```diff
@@ -0,0 +1,87 @@
...
+                    @Override
+                    public void doWithLockNoResult() {
+                        Map<String, Object> stackData = CollectionUtils
+                                .toMap(DataUtils.getFields(event).get(FIELD_STACK));
+                        if (stackData.isEmpty()) {
+                            log.warn("Empty stack for externalStackEvent: {}.", event);
+                            return;
+                        }
+
+                        String kind = stackData.get(ObjectMetaDataManager.KIND_FIELD) != null
```
alena1108: we don't need this dynamic schema type for stack; we have it for services as we used to have dynamic schema definition for k8s (and other external services), but we don't anynore

PR #1852, opened 2016-07-27
Fix for Global service rollback

alena1108:
On service upgrade, we do:

1) mark instance v0 for upgrade and stop it
2) wait till its v1 version is recreated by calling service.reconcile

On service rollback, we:

1) destroy one of the v1 instances
2) unmark v0 instance
3) call reconcile that would start v0 instance

and so on.

When we destroy new/unmark old instance for the rollback to pick it up, we do care to pick up a particular instance only in case when:

* the upgrade is partial - when not all launch configs get upgraded. Then we pick a deployment unit to rollback based on deployment unit uuid

But for cases like global service (and perhaps anti affinity constraint use case) it is also important that instance that is being destroyed, has the same hostId as the instance we rollback to. So the constraint never breaks. Otherwise all kinds of collision can possible happen. 

That correct fix would be too big to add at this point (it would impact the entire upgrade logic).
For now, to fix the Global service upgrade use case (and the bug for kubelet rollback found by QA) I've added cleanup logic to the globalDeploymentPlanner to kill the extra instances if the host already has the instance assigned to it. 

https://github.com/rancher/rancher/issues/5558


alena1108:
Submitted an issue for myself for unresolved part of the issue: https://github.com/rancher/rancher/issues/5561

PR #1853, opened 2016-07-27
Adding mac_address to metadata

leodotcloud:


alena1108:
@leodotcloud LGTM. Have you updated go-rancher-metadata go bindings? If so, could you link corresponding PR here?

PR #1859, opened 2016-08-04
Add instead of set the X-Forwarded-Port header

ractive:
The haproxy sets the X-Forwarded-Port header instead of adding it. This breaks http servers that rely on this header to create URLs if other loadbalancers or proxies are put in front of rancher.

Our setup is:
- Https terminated on load balancer, which sets the X-Forwarded-Port port to 443, forwards to rancher
- rancher haproxy listening on port 80, setting X-Forwarded-Port to 80
- The http server creates wrong URLs in the form https://example.com:80 because it honors the X-Forwarded-* headers to create the URLs.


ibuildthecloud:
LGTM.  Thanks!

willseward:
OH!!! That explains so much! Thanks for the fix, @ractive!

PR #1870, opened 2016-08-11
Changes configFile format

mrajashree:
Format of configFile is changed


mrajashree:
@vincent99 Changed the format for configFile

PR #1887, opened 2016-08-15
Improve docker tests

cjellick:
Make test_volume_restore fail faster if it is going to fail and print
more information so that we can understand why it fails randomly
sometimes.

Remove a volume from test_docker_volumes. This is an expensive test and
the third volume was not adding value.

Avoid pulling a new image in test_cleanup_volume_strategy and
test_service_links_with_no_ports.

Make test_delete_network_agent a nonparallel test since it impacts
the running time of other tests that rely on the network agent existing.

 of text is the title and the rest is description.


ibuildthecloud:
L(super swell)TM

PR #1900, opened 2016-08-19
Load Balancer: allow regex for request host 

AlexRRR:
## background
In our current setup we have several services using HTTPS all sharing a wildcard certificate `*.foo.com` so service names in our branch environments usually have the naming convention `servicename-branch.foo.com`.  Each branch environment is usually its own stack.

With such naming convention it was not possible to route`*-branch.foo.com` as the haproxy config uses `hdr_beg` or `hrd_end` part of the request host.

## usage
This change would recognize anything starting with `^` and ending with `$` as a regular expression and will use `hdr_reg` for the routing. This would unblock us as it allows us to use a pattern as `^.*branch.*$` for routing.

Idiomatically I would rather have used `/regex/` but that pattern is recognised as a path.


vincent99:
I don't think this is a good idea.. There is no explicit ordering of rules and no reasonable way to implicitly order regexes to do the "right" thing.  It's also very easy to provide an invalid regex and prevent startup(?), or an incorrect one that doesn't match what you wanted.

1.2 will have a new balancer API that will let you provide any custom configuration you want, as well as explicit ordering of the rules.  That would allow you to add your own matches that use regex or anything else.

AlexRRR:
> There is no explicit ordering of rules and no reasonable way to implicitly order regexes to do the "right" thing. It's also very easy to provide an invalid regex and prevent startup(?), or an incorrect one that doesn't match what you wanted.

This is the way with regexes, this would be more a _use with at own risk_ kind of feature, same risk as if one uses haproxy directly, same risk we will have with the new 1.2.

We do need this feature, but since the new release provides an way to do this, I will close the request.

Keep up the work, I really enjoy working with Rancher.

PR #1919, opened 2016-08-27
don't add dns record until container is running

bdentino:
Fixes https://github.com/rancher/rancher/issues/5792

Full disclosure, I took my first actual look at the rancher codebase this afternoon and I have minimal experience with Java, so it's fair to say [I have no idea what I'm doing](http://i.imgur.com/kNGKIvW.jpg) and I apologize if I'm violating some protocol or standard here. I might even just be missing something if there is a reason the decision was made to include "starting" as a state in which to allow dns exposure. 

Pretty straightforward change, but to be sure I've tested by setting up a service on a local deployment with 3 nodes launched via 10acre-ranch, and scaling it across all 3 hosts while monitoring the dns records provided by rancher. Without this patch, IPs are exposed as soon as each container is scheduled, while the image is downloading, extracting, etc which effectively directs some traffic away from healthy instances to containers that don't exist yet. After this change, the IP is not exposed until the container enters the 'running' state and its healthchecks, if any, are passing (unless there are no available records, in which case one is still exposed immediately upon being scheduled...this looks like it was the existing behavior and I didn't see a reason to change it).

Let me know if this requires any additional information or due diligence before a PR can be accepted; I'd be happy to oblige.

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java
@@ -132,8 +132,7 @@ public class DnsInfoDaoImpl extends AbstractJooqDao implements DnsInfoDao {
                         .and(targetNic.REMOVED.isNull())
                         .and(instanceLink.REMOVED.isNull())
                         .and(instanceLink.SERVICE_CONSUME_MAP_ID.isNull())
-                        .and(targetInstance.STATE.in(InstanceConstants.STATE_RUNNING,
-                                InstanceConstants.STATE_STARTING))
+                        .and(targetInstance.STATE.in(InstanceConstants.STATE_RUNNING))
                         .and(targetInstance.HEALTH_STATE.isNull().or(
                                 targetInstance.HEALTH_STATE.eq(HealthcheckConstants.HEALTH_STATE_HEALTHY))))
                 .fetch().map(mapper);
@@ -460,7 +459,7 @@ public class DnsInfoDaoImpl extends AbstractJooqDao implements DnsInfoDao {
             existingData.add(data);
             returnData.put(data.getService().getId(), existingData);
         }
-
+        
         // parse the health check
         Map<Long, List<ServiceInstanceData>> returnDataFiltered = new HashMap<>();
         if (client) {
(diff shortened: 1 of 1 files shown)


vincent99:
I believe the problem here is other use-cases require that the entry be available at the time the entrypoint is run... @alena1108 / @ibuildthecloud ?

vincent99:
(An alternative/workaround would be to use an internal load balancer and healthchecks on the services.. Backend won't get traffic from the balancer until they are healthy)

bdentino:
With the suggested workaround, the load balancer will still receive traffic
before its container is ready, no?

ibuildthecloud:
@bdentino Kudo's on navigating the Java code base and figuring this out.  I really need @alena1108 feedback on this.  My first reaction is this probably won't work but instead we need to be looking at the healthstate.  But there are really a lot of corner cases here to consider.  The other thing is we are just about to rewrite the DNS management code in Go.  If there's a quick simple change we can do in Java, that would be great, otherwise we might have to wait for the Go rewrite.

alena1108:
@ibuildthecloud we do look at the health state:

https://github.com/rancher/cattle/blob/master/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/DnsInfoDaoImpl.java#L137

and if the health check is configured, only instances in Healthy state will get propagated to DNS config.

Regarding Starting state. apply.config for DNS ("hosts" config item) is called on:

* nic.activate
* instance.updatehealthy 

As at the point nic.activate is called, the instance can still be in Starting state, I've included this state to the list of valid states. Unless we change the trigger from nic.activate to instance.start (or add instance.start to the list of triggers), the change introduced by this PR, can cause failures.

bdentino:
Thanks for the feedback, guys. I had my suspicions that if it really were such a simple fix it would've already been done. But I'm continually surprised by what sometimes slips through the cracks, especially in larger projects, so I figured it was worth taking a look.

@alena1108 I did notice that the health state was queried, but this issue does manifest even with services that have a health check configured. For such services, it looks an instance's DNS record is created immediatly, then removed once it starts until it transitions to a healthy state. Is it possible or expected that a DNS record could be created for a health-checked service before its health state is actually initialized (i.e. still null)? 

Also, what are the symptoms of some of the failures this could cause; under what circumstances? Will rancher auto-recover? I ask because I've been running a scaled-down version of our production stack on a release with this patch since Saturday just to kick the tires, and even after a reasonable amount of service deployments, upgrades, host removals and additions, etc I haven't seen anything unusual. I'd like to try and force a failure if I can just to see what happens.

alena1108:
@bdentino 

> I did notice that the health state was queried, but this issue does manifest even with services that have a health check configured. For such services, it looks an instance's DNS record is created immediatly, then removed once it starts until it transitions to a healthy state. Is it possible or expected that a DNS record could be created for a health-checked service before its health state is actually initialized (i.e. still null)? 

your observations are correct. It is quite possible for the instances to be propagated to DNS before health check state is actually set on it.

> Also, what are the symptoms of some of the failures this could cause; under what circumstances? 

Most likely you don't observe the failures because there are lots of dns config updates happening after your target instances has started. Every dns update triggers the entire config reload. So chances that instance will be missing from dns config, are really low. It can only happen when the instance start (w/o health check) is not followed immediately by any other operation triggering config reload.

alena1108:
Fix is applied on the refactored dns microservice as per https://github.com/rancher/rancher/issues/5792, closing this PR for now. @bdentino thank you for raising the issue.

bdentino:
Fantastic! Thanks everyone!

PR #1923, opened 2016-08-29
Upgrade fixes: state transition changes

alena1108:
This PR:

1) Check for the service state BEFORE marking instances for rollback
2) Removed cancel-rollback state
3) When finishUpgrade is called on a service in canceled-upgrade state, complete the upgrade first, then finish the upgrade.

@ibuildthecloud


alena1108:
test_authorization failed as looks like removed cancelrollback action is frozen for v1 APIs

alena1108:
there is one TF - test_bad_agent; I've seen it happening quite often, not sure if it's a known issue.

PR #1935, opened 2016-09-01
Make shouldStart logic less confusing

cjellick:
Is it just me or is the current implementation unnecessarily confusing? Did I miss something?


ibuildthecloud:
It does seem overly confusing but I think the following is the simplest
```java
 protected boolean shouldStart(Instance instance) {
    return DataAccessor.fields(instance)
        .withKey(InstanceConstants.FIELD_START_ON_CREATE)
        .withDefault(true)
        .as(Boolean.class);
}
```

The default is set to `true` so it can never be null.  If it's never null you just use the value of `startOnCreate`

ibuildthecloud:
@cjellick If you rebase on master the tests should pass.  There was an issue in the v1 tests that would always fail

PR #1941, opened 2016-09-03
look for stacks using account id as well

wlan0:
@alena1108


code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
```diff
@@ -131,7 +131,7 @@ Stack getStack(final ExternalEvent event) {
         }
         final String envExtId = eId.toString();
 
-        Stack stack = objectManager.findOne(Stack.class, STACK.EXTERNAL_ID, envExtId);              
+        Stack stack = stackDao.getStackByExternalId(event.getAccountId(), envExtId);
```
ibuildthecloud: whitespace

wlan0: oh right. Missed that!

PR #1958, opened 2016-09-08
Parallelize deployment unit alloction

cjellick:


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
```diff
@@ -86,10 +93,10 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
             try {
                 progress.checkPoint("Waiting for dependencies");
                 // wait till volumesFrom/networksFrom containers start up
```
ibuildthecloud: Poor orphaned comment.  All alone without it's matching method.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstanceStart.java
```diff
@@ -167,7 +171,9 @@ protected void waitForDependenciesStart(Instance instance) {
                     new ResourcePredicate<Instance>() {
                         @Override
                         public boolean evaluate(Instance obj) {
-                            return validateState(obj);
+                            InstanceHostMap ihm = objectManager.findAny(InstanceHostMap.class, INSTANCE_HOST_MAP.INSTANCE_ID, obj.getId(),
```
ibuildthecloud: I don't think is taking into account "startOnce" containers as it did before.  In this situation the ihm maybe be "inactive" yet valid.

cjellick: Ah, right.

tests/integration-v1/cattletest/core/test_allocation.py
```diff
@@ -606,7 +607,8 @@ def test_volumes_from_constraint(new_context):
         c2 = new_context.create_container_no_success(startOnCreate=False,
                                                      dataVolumesFrom=[c1.id])
 
-        c1 = c1.start()
+        c1 = client.wait_success(c1.start())
+        assert c1.state == 'running'
```
cjellick: @ibuildthecloud this needed to change because it is now feasible for c2 to get to allocation before c1 and thus fail one the fact that c1 isn't running.

Does this make sense to you?

PR #1968, opened 2016-09-09
Check allocation before checking instance state

alena1108:
https://github.com/rancher/rancher/issues/5944


alena1108:
test_bad_agent and test_virtual_machine_default_fields TFs

PR #1978, opened 2016-09-13
Renaming hostId to host_id to reflect in metadata correctly

leodotcloud:
@ibuildthecloud @alena1108


ibuildthecloud:
@leodotcloud We can add a field but we can't rename a field because it's a breaking change.  You can add `host_id`

ibuildthecloud:
@leodotcloud Is there any change that you can just use hostUuid for whatever you are doing.  HostId really should have never been in the metadata to begin with.

PR #1980, opened 2016-09-14
Remove scale from binding

mrajashree:
Scale is removed from binding


mrajashree:
@vincent99 can you review this?

alena1108:
LGTM

mrajashree:
The CI build failed because of two tests, the one in v1 was `test_validate_service_scaleup_scaledown`, that failed because of 409 conflict error.
For v2, `test_health_check_quorum` test failed which is a known error

PR #1982, opened 2016-09-15
Exclude os-specific python files from build and upgrade py.test

cjellick:


cjellick:
This test failure is full of LIES.

PR #1984, opened 2016-09-16
update metadata on instancehostmap.activate

alena1108:
As labels are being back populated from docker via this event

https://github.com/rancher/rancher/issues/5985

@ibuildthecloud we would need this change for 1.1.4, it fixes the bug in metadata that @cloudnautique hit


ibuildthecloud:
We are so close to just update metadata on *

alena1108:
@ibuildthecloud :)

PR #1989, opened 2016-09-19
Remove trailing newline on alloc logging

cjellick:


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AbstractAllocator.java
```diff
@@ -441,6 +441,7 @@ protected void logCandidate(String prefix, AllocationAttempt attempt, Allocation
         for (Map.Entry<Long, Long> entry : candidate.getSubnetIds().entrySet()) {
             candidateLog.append(String.format("  %s   nic [%s] subnet [%s]\n", prefix, entry.getKey(), entry.getValue()));
         }
+        candidateLog.deleteCharAt(candidateLog.length() - 1); // Remove trailing newline
         log.info(candidateLog.toString());
```
alena1108: Nit: we can also use log.info(candidateLog.toString().trim()) that trims whitespace/newlines.

cjellick: I realize that this is micro-optimization that will make virtual no difference, but I went with this to avoid the extra string copy that trim() would involve. I figured we have a stringBuilder, might as well use it.

But if you feel strongly about it, I'll change it.

alena1108: It's fine, LGTM

PR #1991, opened 2016-09-20
Do not return instances service when link to the service is gone

alena1108:
https://github.com/rancher/rancher/issues/6007


cjellick:
LGTM, assuming green build

PR #1992, opened 2016-09-20
Fix instance label logic

cjellick:
- In two places (AllocatorServiceImpl and ServiceDiscoveryServiceImpl),
  we were using the label table instead of the label field on
instances, which is the source of truth. This lead to a race condition
where labels that allocation needed were not present.
- Remove setlabels action. This was not backpopulating the
  instance.labels field, which is the source of truth and we decided we
did not want to support it anymore.


alena1108:
LGTM, unless there are TFs

alena1108:
Looks like test_set_container_labels should be removed from v1 as well.

PR #1993, opened 2016-09-20
Localization of China

orangedeng:
I am going use env variable to set default language and to active those machine driver which can be used in China.

About default language, I have already put a PR "https://github.com/rancher/ui/pull/843" .

PR #1994, opened 2016-09-20
Localization of china

orangedeng:
I am going use env variable to set default language and to active those machine driver which can be used in China.

Last PR is https://github.com/rancher/cattle/pull/1993 and I closed it because it could not pass the style check.
About default language, I have already put a PR "https://github.com/rancher/ui/pull/843" .


ibuildthecloud:
After looking at this it might be better if you look at extending the `rancher/server`.  If you do that then the following should be possible

```dockerfile
FROM rancher/server:v1.1.3
ENV DEFAULT_LOCALIZATION zh-Hans-CN
ENV DEFAULT_CATTLE_MACHINE_DRIVER_ALIYUNECS true,https://github.com/orangedeng/rancher-machine-driver/raw/master/Aliyun/docker-machine-driver-aliyunecs.tgz,
ENV DEFAULT_CATTLE_MACHINE_DRIVERS aliyunecs,amazonec2
```

With the above approach you need no code changes for machine drivers.  Only the changes needed for the cookie.  For the cookie it should be set regardless of the locale.  So it doesn't need to be chinese specific.  Just set the cookie to the default language if the cookie is not found and the default localization is set.

code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
```diff
@@ -37,6 +37,9 @@
     private static final DynamicStringListProperty IGNORE = ArchaiusUtil.getList("api.ignore.paths");
     private static final DynamicStringProperty PL_SETTING = ArchaiusUtil.getString("ui.pl");
     private static final String PL = "PL";
+    private static final String LOCALIZATION = System.getenv("CATTLE_LOCALIZATION");
```
ibuildthecloud: This should be
```java
      private static final DynamicStringProperty LOCALIZATION = ArchaiusUtil.getString("localization");
```

What this means is that the env variable `CATTLE_LOCALIZATION` will set it, or you can set `DEFAULT_CATTLE_LOCALIZATION` to set the default value but still let the user override it.  Typically we set `DEFAULT_CATTLE_*` in the dockerfile to prepopulate settings.

code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
```diff
@@ -37,6 +37,9 @@
...
+    private static final String LOCALIZATION_CHINESE = "zh-Hans-CN";
```
ibuildthecloud: This is unneeded per https://github.com/rancher/cattle/pull/1994#issuecomment-248626932

code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
```diff
@@ -201,5 +206,23 @@ private void addPLCookie(HttpServletRequest httpRequest, HttpServletResponse res
             response.addCookie(plCookie);
         } 
     }
-
+    
+    private void addDefaultLanguageCookie(HttpServletRequest httpRequest, HttpServletResponse response) {
+        Cookie languageCookie = null;
+        if(!LOCALIZATION_CHINESE.equals(LOCALIZATION))
```
ibuildthecloud: Just check `StringUtils.isNotBlank(LOCALIZATION.get())`

code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
```diff
@@ -201,5 +206,23 @@ private void addPLCookie(HttpServletRequest httpRequest, HttpServletResponse res
...
+        if(httpRequest.getCookies()!=null) {
+            for(Cookie c : httpRequest.getCookies()) {
+                if(DEFAULT_LANGUAGE.equals(c.getName()) && c.getName()!=null) {
+                    languageCookie = c;
+                    break;
+                    }
+                }
+        }
+        if(languageCookie == null) {
+            languageCookie = new Cookie(DEFAULT_LANGUAGE, LOCALIZATION_CHINESE);
```
ibuildthecloud: Set value to `LOCALIZATION.get()`.  This doesn't need to be chinese specific.

code/implementation/sample-setup/src/main/java/io/cattle/platform/sample/data/SampleDataStartupV10.java
```diff
@@ -29,6 +29,9 @@
     Configuration configuration;
```
ibuildthecloud: All changes to this file aren't needed per my comment https://github.com/rancher/cattle/pull/1994#issuecomment-248626932

resources/content/cattle-global.properties
```diff
@@ -38,6 +38,10 @@ machine.driver.amazonec2=true,local://,
 machine.driver.azure=true,local://,
```
ibuildthecloud: All changes unneeded due to https://github.com/rancher/cattle/pull/1994#issuecomment-248626932

PR #1997, opened 2016-09-20
Apply instance labels before config item update trigger

alena1108:
So the metadata is updated after the labels change

https://github.com/rancher/rancher/issues/5985


alena1108:
This fix should go to master and 1.1.4

cjellick:
LGTM

PR #2000, opened 2016-09-20
remove modifyCompute

cjellick:


ibuildthecloud:
One of the test failures seems fishy

code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/dao/impl/SimpleAllocatorDaoImpl.java
```diff
@@ -59,7 +55,7 @@
         Map<Long, String> hostIdsToUuids = new HashMap<>();
         if (orderedHostUuids == null) {
             Result<Record3<String, Long, Long>> result = getHostQuery(orderedHostUuids, options).fetch();
-
+            Collections.shuffle(result);
```
cjellick: Randomizing the results so that we don't always pick the same host.  Waddaya think?

PR #2001, opened 2016-09-20
no lifecycle for deployment unit

alena1108:


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DeploymentUnit.java
```diff
@@ -223,9 +226,34 @@ public void remove(String reason, String level) {
             instance.remove();
         }
 
-        // remove deployment unit object
         if (unit != null) {
-            context.objectProcessManager.scheduleStandardProcessAsync(StandardProcess.REMOVE, unit, null);
+            clenaupVolumes();
+            Map<String, Object> params = new HashMap<>();
+            params.put(ObjectMetaDataManager.REMOVED_FIELD, new Date());
+            params.put(ObjectMetaDataManager.REMOVE_TIME_FIELD, new Date());
```
ibuildthecloud: Since it's not an orchestrated resource can you just delete the row?  Otherwise, drop the REMOVE_TIME and STATE fields.

alena1108: @ibuildthecloud I can't delete the row as it is being referenced by the volume object

PR #2006, opened 2016-09-22
health state on environment

alena1108:
@ibuildthecloud


ibuildthecloud:
@alena1108 You're awesome!

PR #2011, opened 2016-09-26
Wait for deployment instances in InstanceStart

cjellick:


cjellick:
I believe this addressed https://github.com/rancher/rancher/issues/6005

PR #2012, opened 2016-09-26
Fix test that was relying on compute spread

cjellick:
We removed compute counting in cattle. Can no longer rely on that.


alena1108:
LGTM assuming the build is green

cjellick:
Ah shoot, have to update the v1 test.

PR #2018, opened 2016-09-28
deployment unit volume name

alena1108:


tests/integration/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,3 @@
+{
+  "cattletest/core/test_svc_volume_template.py::test_du_volume": true
+}
```
ibuildthecloud: ??

alena1108: fixing the tests, will remove this file. Not sure where it came from

PR #2048, opened 2016-10-07
container, stack, service name lowercased in metadata

alena1108:
@vincent99


vincent99:
LGTM, see also https://github.com/rancher/rancher-metadata/pull/22

deniseschannon:
https://github.com/rancher/rancher/issues/1908

ibuildthecloud:
@alena1108 I'm afraid of NPE's, are they definitely not possible?

alena1108:
@ibuildthecloud min length for service/stack name is 1; no such check for container. I'll add not null check just to be safe

PR #2051, opened 2016-10-07
Checks default registry

mrajashree:
https://github.com/rancher/rancher/issues/6195


mrajashree:
@alena1108 I had missed this if condition in the earlier registries code. All test cases need to have `index.docker.io` in the whitelist, else the other unrelated tests running in parallel fail. So cannot add a test for this specific condition

alena1108:
LGTM

PR #2052, opened 2016-10-08
Fix health_check_quorum and test_pagination_include tests

dx9:
I vote for consistency by default. But ya this could be pushed onto the consumer. I'm easy LMK.

https://github.com/rancher/rancher/issues/6110

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java
@@ -314,21 +314,19 @@ public abstract class AbstractJooqResourceManager extends AbstractObjectResource
     }
 
     protected void addSort(SchemaFactory schemaFactory, String type, Sort sort, SelectQuery<?> query) {
-        if (sort == null) {
-            return;
-        }
+        if (sort != null) {
+            TableField<?, Object> sortField = JooqUtils.getTableField(getMetaDataManager(), type, sort.getName());
+            if (sortField == null) {
+                return;
+            }
 
-        TableField<?, Object> sortField = JooqUtils.getTableField(getMetaDataManager(), type, sort.getName());
-        if (sortField == null) {
-            return;
-        }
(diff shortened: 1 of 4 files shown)


ibuildthecloud:
@ds9 MySQL does this by default, so I like this.  But this kind of conflicts with https://github.com/rancher/cattle/blob/master/code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java#L334.  This was a change done recently so it may not been present when you diid this change.  We automatically add "order by id" if a user specifies any sort at all.  So I think this change and the existing change need to be slightly refactored to work well together.  Basically regardless of the user input we alway add "order by id" at the end.

dx9:
@ibuildthecloud yeah I missed that. I'll clean it up. Also noticed sorting is related to the pagination_include failures. We need to sort by a unique id across the include. Looking into the best way to do that.

ibuildthecloud:
After merging the postgres stuff we are noticing a lot more test failures.  Not that it has anything really to do with postgres per say.  It seems its add more load uncovering more timing issues in our tests.  Any help would be appreciated as you seem to be really good at tracking these things down.  Thanks!

dx9:
@ibuildthecloud Yeah, it's why I wasn't bugging ya to merge. Didn't want it to mess up the workflow. Happy to try and help!

@alena1108 I think these fix the issues.

dx9:
I'm thinking this breaks user specified sorts in it's current form.

alena1108:
@dx9 just wanted to make sure the below refers to the previously failing test for pagination, and not AbstractJooqManager code changes causing some new failures:

>> I'm thinking this breaks user specified sorts in it's current form.

let me know if the PR is ready for review/merge

ibuildthecloud:
@dx9 Thanks!

dx9:
@alena1108 @ibuildthecloud Thanks guys

dx9:
@ibuildthecloud @alena1108 still ain't right... looking into it

https://drone.rancher.io/rancher/cattle/844/1

PR #2066, opened 2016-10-10
Added missing "Type" field for embedded list/map objects

alena1108:
https://github.com/rancher/rancher/issues/6207

WIP as trying to figure out why type is still not being set on PublicEndpoint object when viewed in API. I see it being set when debugging in formatType() code piece. UPDATE: figured out that one


alena1108:
WIP till the drone is fixed, and all the tests are executed

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/HostOutputFilter.java
```diff
@@ -1,53 +0,0 @@
-package io.cattle.platform.servicediscovery.api.filter;
```
ibuildthecloud: Not needed anymore?

PR #2067, opened 2016-10-11
metadata version 2016-07-29

alena1108:
Lowercase service/stack/container name only in the 2016-07-29/latest versions


alena1108:
Also dropped hostId in 2016-07-29 HostMetadata

alena1108:
https://github.com/rancher/rancher/issues/6226

PR #2070, opened 2016-10-11
Change condition to use default registry

mrajashree:
https://github.com/rancher/rancher/issues/5799

### DIFF
--- a/code/iaas/storage-service/src/main/java/io/cattle/platform/storage/api/filter/ExternalTemplateInstanceFilter.java
+++ b/code/iaas/storage-service/src/main/java/io/cattle/platform/storage/api/filter/ExternalTemplateInstanceFilter.java
@@ -20,6 +20,7 @@ import javax.inject.Inject;
 
 import org.apache.commons.lang3.StringUtils;
 
+import com.netflix.config.DynamicStringListProperty;
 import com.netflix.config.DynamicStringProperty;
 
 public class ExternalTemplateInstanceFilter extends AbstractResourceManagerFilter {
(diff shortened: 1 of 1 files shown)


mrajashree:
The `registry.whitelist` setting accepts comma separated values, and the code didn't trim the space between commas earlier that caused the error. There is a test already for this condition, which was passing since in the test there were no spaces between the whitelist registries. @alena1108

ibuildthecloud:
@mrajashree There is a type DynamicStringProperty that handles comma separated lists, you should just use that.

mrajashree:
@ibuildthecloud actually I'm using DynamicStringProperty for the whitelist over here: https://github.com/rancher/cattle/blob/409b60fe3cfe24d61ecb91355c3a6b102e9a59a7/code/iaas/storage-service/src/main/java/io/cattle/platform/storage/api/filter/ExternalTemplateInstanceFilter.java#L32

But if in the API the whitelist entered has space after comma, for example the input is `quay.io, index.docker.io`, the space after comma isn't getting trimmed, and that needs to be trimmed when the whitelist registry is being compared with the default or user provided registry.

ibuildthecloud:
@mrajashree Ooops, typo, DynamicStringListProperty

mrajashree:
@ibuildthecloud changed to `DynamicStringListProperty`

code/iaas/storage-service/src/main/java/io/cattle/platform/storage/api/filter/ExternalTemplateInstanceFilter.java
```diff
@@ -79,51 +83,33 @@ protected static void validateImageUuid(String image) {
...
 
         String server = dockerImage.getServer();
-        String qualifiedName = dockerImage.getQualifiedName();
-        String whitelist = WHITELIST_REGISTRIES.get();
+        List<String> whitelistRegistries = WHITELIST_REGISTRIES.get();
         String defaultRegistry = DEFAULT_REGISTRY.get();
         String userProvidedRegistry = dockerImage.getServer();
 
-        if (!qualifiedName.startsWith(server)) {
+        if (!image.contains(server)) {
```
alena1108: Nit Could you also use consistent names for registry? Like here, "server" is really a registry, but it takes time to figure it out :)

mrajashree: The qualifiedName field is a field from dockerImage, that is created by DockerImage.Parse
In the case when user explicitly prepended `index.docker.io` to the image, DockerImage.Parse was stripping `index.docker.io` off the image name, and then the validateImage function would never get to see that user had provided this registry. It would fail if the default registry had been set to something else. Say `quay.io` because in the absence of userProvidedRegistry, I'm prepending the default registry to image

mrajashree:
@ibuildthecloud @alena1108 The three TFs `test_du_removal`, `test_link_remove_restore` and `test_storage_driver_in_use` seem unrelated

PR #2080, opened 2016-10-13
networks to metadata

alena1108:
@ibuildthecloud


code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
```diff
@@ -229,4 +245,26 @@ protected HostMetaData map(List<Object> input) {
...
     }
 
+    @Override
+    @SuppressWarnings("unchecked")
+    public List<NetworkMetaData> getNetworksMetaData(long accountId) {
+        MultiRecordMapper<NetworkMetaData> mapper = new MultiRecordMapper<NetworkMetaData>() {
+            @Override
+            protected NetworkMetaData map(List<Object> input) {
+                Network ntwk = (Network) input.get(0);
+                Map<String, Object> meta = DataAccessor.fields(ntwk).withKey(ServiceConstants.FIELD_METADATA)
```
ibuildthecloud: I think you need to add NETWORK.DATA to the select.  Also there is DataAccessor.fieldMap() for this

alena1108: fixed

PR #2086, opened 2016-10-14
link external volume template to volume by name

alena1108:
@ibuildthecloud please let me know if it is a correct fix. if the volumeTemplate external, we just locate the volume having name=volumeTemplate.name. Otherwise user would have to:

* create volume template
* create external volume from the template

which is not the correct flow for external volume

@joshwget if the flow above is correct, then there is probably nothing you should change on the compose side. Volume existence validation is done in the services code.

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DeploymentUnit.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DeploymentUnit.java
@@ -358,8 +358,8 @@ public class DeploymentUnit {
         Volume volume = null;
         if (template.getExternal()) {
             // external volume should exist, otherwise fail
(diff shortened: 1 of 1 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DeploymentUnit.java
```diff
@@ -359,7 +359,7 @@ protected void createVolume(final Service service, String volumeName, Map<String
         if (template.getExternal()) {
             // external volume should exist, otherwise fail
             volume = context.objectManager.findOne(Volume.class, VOLUME.ACCOUNT_ID, service.getAccountId(),
```
ibuildthecloud: This should be findAny(), otherwise madness will ensue if the user manages to get two in there

alena1108: will fix

PR #2098, opened 2016-10-18
Schedule reconcile even when service is updating

alena1108:
Don't reject reconcile request when the service is in updating-active state. It makes it race condition prone for this scenario:

* service reconcile is finished, but the state of the service wasn't updated yet
* during this time, instance gets stopped/removed
* but the reconcile request gets rejected as service wasn't put into active state

So we end up having degraded service in active state.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ target
 tests/integration*/.idea/
 tests/integration*/.tox/
 tests/integration*/MANIFEST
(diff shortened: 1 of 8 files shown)


alena1108:
https://github.com/rancher/rancher/issues/5556
https://github.com/rancher/rancher/issues/6092
https://github.com/rancher/rancher/issues/5775

PR #2099, opened 2016-10-18
new state transition between agent.reconnecting and agent.active

alena1108:
https://github.com/rancher/rancher/issues/6013

@ibuildthecloud we use reconnecting state as a transition state when agent going down, and when it is coming back up. As we use agent.reconnect as one of the triggers for service reconcile, but hosts in "Reconnecting" state are not considered as available for allocation, so it is possible that the host might be skipped by global service allocator.

Added "follow up"/"delegate" state "finishUpgrade". "delegate" directive in spring-process-context.xml didn't seem to take affect, so setting this flag explicitly in one of the existing handlers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ target
 tests/integration*/.idea/
 tests/integration*/.tox/
 tests/integration*/MANIFEST
+tests/integration-v1/.cache/
+tests/integration/.cache/
 .venv
 .idea
 *.iml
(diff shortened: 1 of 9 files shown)


ibuildthecloud:
Hmm this is a fragile change.  Adding any state always has scary side effects.

ibuildthecloud:
What if reconnecting went to activating went to active?

alena1108:
@ibuildthecloud this will work too, I'll work on the change

alena1108:
@ibuildthecloud checked the code, have a couple of concerns about delegating to agent.activate:

* agent.reconnect and agent.activate share the same process logic. If I delegate from reconnect to activate, the same logic will be executed twice. 
* agents in activating state are not considered valid for allocation. 

adding an intermediate state with no logic attached to it, and making this intermediate state valid for allocator, seems like a safe fix from allocator point of view, even though I'm aware of the side affects you've brought up. Let me know if you still want to go with agent.activate as a delegate process.

PR #2109, opened 2016-10-21
 cleanup kubernetesStack(s) on system kubernetes removal

alena1108:
https://github.com/rancher/rancher/issues/5852


ibuildthecloud:
Hmmm.... This is all so hacky.  I wish there was a cleaner approach.

ibuildthecloud:
I'm inclined to say we shouldn't do this at all.  Instead the individual k8s components should look for invalid stacks on start.  So install, uninstall, and then next install of k8s will clean up orphans.

PR #2111, opened 2016-10-23
WIP: Call validator on Add account

mrajashree:
https://github.com/rancher/rancher/issues/6358
This PR fixes the bug that `api.auth.local.validate.url` was not being called on adding an account through Admin->Accounts->Add Account.
A timeout on the POST request has also been added as per the issue https://github.com/rancher/rancher/issues/6363

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/CredentialUniqueFilter.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/CredentialUniqueFilter.java
@@ -3,6 +3,9 @@ package io.cattle.platform.iaas.api.auth.dao.impl;
 import io.cattle.platform.core.constants.CredentialConstants;
 import io.cattle.platform.core.model.Credential;
 import io.cattle.platform.iaas.api.auth.dao.PasswordDao;
+import io.cattle.platform.iaas.api.auth.integration.local.LocalAuthPasswordValidator;
+import io.cattle.platform.json.JsonMapper;
+import io.github.ibuildthecloud.gdapi.context.ApiContext;
 import io.github.ibuildthecloud.gdapi.exception.ValidationErrorException;
 import io.github.ibuildthecloud.gdapi.factory.SchemaFactory;
 import io.github.ibuildthecloud.gdapi.request.ApiRequest;
@@ -20,6 +23,8 @@ public class CredentialUniqueFilter extends AbstractResourceManagerFilter {
     SchemaFactory schemaFactory;
     @Inject
     PasswordDao passwordDao;
+    @Inject
+    JsonMapper jsonMapper;
 
     @Override
     public Object create(String type, ApiRequest request, ResourceManager next) {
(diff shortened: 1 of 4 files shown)


mrajashree:
@prachidamle can you review this?

code/framework/encryption/src/main/java/io/cattle/platform/framework/encryption/Hasher.java
```diff
@@ -2,13 +2,21 @@
...
 import org.apache.commons.lang3.StringUtils;
 
 public abstract class Hasher implements Transformer, Named {
 
+    @Inject
+    JsonMapper jsonMapper;
+
     @Override
     public String transform(String value) {
+        LocalAuthPasswordValidator.validatePassword(value, jsonMapper);
```
ibuildthecloud: This seems like way too generic of a place to put this validation.  What use case are you trying to catch with this.

mrajashree: I wanted to call the validation code on adding an account from Admin->Accounts->Add Account.
The other functions I saw on debugging that were called on adding an account in that manner had the hashed value of password entered by the user

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/CredentialUniqueFilter.java
```diff
@@ -20,13 +22,16 @@
...
+    @Inject
+    JsonMapper jsonMapper;
 
     @Override
     public Object create(String type, ApiRequest request, ResourceManager next) {
         Credential credential = request.proxyRequestObject(Credential.class);
         if (StringUtils.isBlank(credential.getKind())) {
             credential.setKind(request.getType());
         }
+        LocalAuthPasswordValidator.validatePassword(credential.getSecretValue(), jsonMapper);
```
mrajashree: @ibuildthecloud yes sorry I was adding the call to TransformationService.
So along with calling the validatePassword only if credential is of kind password, even the transformationService will be called only if credential is password right?

mrajashree:
@ibuildthecloud I have done the changes including checking if credential is of kind password and calling the transformationService. I checked in the db and password is being hashed and stored. When I add account from Admin->Accounts->Add account, the password validation function is called now as expected, but on entering an acceptable password it seems to hang on `Saving` account although it has already been added in the db. I can even see it in the UI list of accounts

mrajashree:
The TFs `test_pagination_include`, `test_container_ha_restart`, `test_service_affinity_rules_w_policy` seem unrelated. And other TFs are docker related or have failed with `ClientApiError: Scheduling failed: No healthy hosts with sufficient resources available`

ibuildthecloud:
@mrajashree Have you resolved the issue you saw in the UI where it would get stuck in some state?

mrajashree:
@ibuildthecloud no I'm still figuring it out

mrajashree:
One test from `test_local_auth.py` is failing, `test_passwords_non_alpha_numeric_characters` and it fails only for v1 with an `Internal Server Error` as: `Exception in API for request [http://localhost:8081/v1/passwords].` 
Also the next line on the log says `java.lang.IllegalArgumentException: Transformer: the does not exist.` so is it because TransformationService handler was removed from default.properties?

mrajashree:
The failing test `test_passwords_non_alpha_numeric_characters` requires changes to v1-schemas

cjellick:
Closing in favor of https://github.com/rancher/cattle/pull/2130

PR #2120, opened 2016-10-26
Convert event exceptions in FailedToAllocate

cjellick:
When the scheduler replies with an error, that was being bubbled up as
an EventExecutionException. This type of exception prevented the
instance from ever failing to allocated, it just got stuck in starting.

This change catches those exceptions, logs their message (which
contains the error message sent by the scheduler) and rethrows
FailedToAllocate exceptions, which will cause the instance to go to an
error state.


cjellick:
Partially addresses https://github.com/rancher/rancher/issues/6253
(The underlying bug was fixed in the scheduler, this fixes the fact that the containers got stuck in "starting", instead of going to error).

PR #2122, opened 2016-10-26
HAProxy : Better security

Dremor:
Hi.

I changed the base HAProxy configuration in order to achieve a better security.

Why? As there is currently no way to control the whole configuration of Rancher Load-balancer/Reverse Proxy, we are stuck with a very bad grade on of the security check websites (F on Mozilla's Observatory, same for Security Header, etc.).

This configuration gives the following ratings:
-> [Mozilla Observatory](https://observatory.mozilla.org/) : B/B+, intermediate compatibility (Firefox 1, Chrome 1, IE 7, Opera 5, Safari 1, Windows XP IE8, Android 2.3, Java 7)
-> [Security Header](https://securityheaders.io/) : A
-> [SSLLab](https://www.ssllabs.com/ssltest/) : A+
-> [HTBridge](https://www.htbridge.com/ssl/) : A+, PCI DSS compliance
-> [tls.imirhil.fr](tls.imirhil.fr) : A+

Tested it with the following apps :
- Rancher
- Gitlab CE

Other apps should work, unless they needs mixed content (HTTP content on a HTTPS page).

Note : I used some acl in order to avoid overwriting the application own policies, if it has one.


ibuildthecloud:
https://github.com/rancher/cattle/pull/1798 removes the management of haproxy config from cattle to this microservice https://github.com/rancher/lb-controller in https://github.com/rancher/lb-controller/pull/15.  Basically I'm just saying we are in the middle of a transition (that should be finished this week), so it might be a bit tricky getting this in.

alena1108:
@Dremor as we no longer have lb code in Rancher, following can be done:

1) more secure haproxy config can be applied using custom haproxy configuration via loadBalancerService.create/update APIs to a particular LB service
2) PR for enhancing the default values should be put to https://github.com/rancher/lb-controller

Closing this PR for now

PR #2100, opened 2016-10-19
CI Containerize DB

dx9:

PR #2132, opened 2016-10-28
Revert "CI Containerize DB"

ibuildthecloud:
Reverts rancher/cattle#2100

@dx9 This PR has an unfortunate side effect that is uses the `docker-proxy` for DB communication.  And we all know the `docker-proxy` is a highly effective dual-ISDN line.  Basically it's killing the CPU on the drone boxes.  So we can probably reconfigure docker to `--userland-proxy=false` but that isn't the default and then just becomes the pain for anybody else.

The fix for this is really just to not use `localhost:XXXX`.  If you just lookup the containers IP from `docker inspect` and do CONTAINER_IP:3306 then it won't use the `docker-proxy`.  I think if you even use the hosts local IP (whatever is on eth0) it ends up bypassing the proxy too.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -6,23 +6,23 @@ ENV DAPPER_SOURCE /usr/src/cattle
 ENV DAPPER_OUTPUT dist
 WORKDIR ${DAPPER_SOURCE}
 
-# MySQL
-ENV MYSQL_VERSION 5.5
-ENV MYSQL_HOST 127.0.0.1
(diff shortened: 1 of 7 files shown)


ibuildthecloud:
@dx9 Just to be clear, I'd really like to re-merge this PR.  I just don't have time to fix it and it's causing a bit of hold up right now, so that's why I reverted it so quickly.

dx9:
@ibuildthecloud No problem, I'll fix it up.

PR #2142, opened 2016-10-29
CI Containerize DB Take 2

dx9:
Followup for https://github.com/rancher/cattle/pull/2132


ibuildthecloud:
@dx9 I'm pretty sure this is going to conflict with https://github.com/rancher/cattle/pull/2088 which will be merged before this one.  One thing that I changed in that PR (besides the million other things) is that on mysql start we preseed the database with https://github.com/rancher/rancher/blob/master/server/artifacts/mysql-dump.sql.  It is just to speed up the startup because the full liquibase migration on mysql takes a long time.  We do the exact same thing when you do `docker run rancher/server` at https://github.com/rancher/rancher/blob/master/server/artifacts/mysql.sh#L91

dx9:
@ibuildthecloud ok, I'll wait for the madness to drop and then rebase this to keep the seeding.

ibuildthecloud:
@dx9 1.2 is finally going out tomorrow so I want to start merging all these pending PRs.  Can you rebase this one just to make sure it's still good.

dx9:
yep, rebased

PR #2144, opened 2016-10-30
Make JwtTokenServiceImplTest work on Java 8

dx9:

### DIFF
--- a/code/framework/token/src/main/java/io/cattle/platform/token/impl/JwtTokenServiceImpl.java
+++ b/code/framework/token/src/main/java/io/cattle/platform/token/impl/JwtTokenServiceImpl.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import javax.inject.Inject;
 
 import org.apache.commons.lang.StringUtils;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
 
 import com.netflix.config.DynamicLongProperty;
 import com.netflix.config.DynamicStringProperty;
@@ -26,6 +27,7 @@ import com.nimbusds.jose.JWSHeader;
 import com.nimbusds.jose.JWSObject;
 import com.nimbusds.jose.JWSSigner;
 import com.nimbusds.jose.JWSVerifier;
+import com.nimbusds.jose.jca.JCASupport;
 import com.nimbusds.jose.crypto.RSADecrypter;
 import com.nimbusds.jose.crypto.RSAEncrypter;
 import com.nimbusds.jose.crypto.RSASSASigner;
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
Thanks for finding this one, been bugging developers for a bit, but nobody ever looked into it.  Do you know a bit more background on it?  We use bouncycastle at runtime which leads me to believe to we need to do something elsewhere if we plan to switch to Java 8 (which I very much want to do).  The thing that is odd is if the JDK provides an implementation, why isn't the output compatible?

dx9:
@ibuildthecloud Ya, I don't know why they vary, would need more digging to understand it better. 

This does currently default to J8 at runtime. So I think SshKeyGen would need to be updated as well to default to Bouncycastle on J8 at runtime. Maybe the security provider init could be pulled up higher since multiple things depend on it?

FWIW, I've mashed together a J8 build with all the maven dependencies updated. Still checking it over but I think it is pretty much there. Need to break it into logical commits.

This PR needs to go in first so the maven caching in dapper doesn't break on J8.

dx9:
@ibuildthecloud This is no good. I messed up testing it...

ibuildthecloud:
@dx9 So the thing that has actually held us up from going to Java 8 is the rancher/server image.  To use Java 8 we need to go to Ubuntu 16.04 (or alpine?).  We tried that but that means we went to a newer MySQL version too which caused upgrade issues.  Users should be able to stop old rancher server and then create a new container with `--volume-from old-server` and it should work.  But it didn't and we didn't have time to look into why and just rolled back.

TBH I really want to switch to Java 8 because the code could be much less verbose if we could use the lambda and stream API.

dx9:
@ibuildthecloud ok, cool. I'm gonna work on this line if it's open. 

I'll look at getting the database upgrade going. I wonder if we could containerize the db usage in rancher/server.

Ya lots of interesting things to play with in J8 land. Jooq and Jetty are pretty locked in there too.

I'm working on a PR for switching Cattle to Alpine J8 env. I can then rebase the Alpine stuff over in rancher.

LMK if any this thinking is off.

ibuildthecloud:
For rancher/server we probably wouldn't pull the mysql out of that container.  It's purely a user experience thing.  We really like the simplicity of `docker run -p 8080:8080 rancher/server` is really all you need to do to get started.  Then afterwards you can migrate to an external DB, containerize DB, or just stick with the embedded one.

dx9:
Ya that makes sense. Mysql is just big for some reason compared to postgres (~115MB more on alpine). Automatically managing the db of choice outside of the server container or switching to pg would be optimal from a minimal size perspective. I don't think the lazy db install approach I was playing with earlier is really a great option for gapped setups.

No worries though if things need to stay the way they are. Just kicking it around.

cjellick:
@dx9 can you rebase this PR? I'd like to pursue merging it.

cjellick:
@dx9 I submitted a rebased version here: https://github.com/rancher/cattle/pull/2304
You don't need to rebase. It still has you as the author.

dx9:
@cjellick give'r. thanks

PR #2147, opened 2016-10-31
image is required for lb service + image uuid is a public setting

alena1108:
@vincent99


alena1108:
https://github.com/rancher/rancher/issues/6460

PR #2149, opened 2016-11-01
Add setting for timeout

mrajashree:
https://github.com/rancher/rancher/issues/6363

@cjellick This PR adds a global setting for auth validation service timeout, set to 10000 milliseconds by default.


cjellick:
LGTM, assuming green build

mrajashree:
`test_resource_based_scheduler` has failed in both v2 runs, and one more TF is `test_project_from_template`. These two seem unrelated @cjellick

PR #2154, opened 2016-11-02
Fetch system flag from table field

LLParse:


deniseschannon:
https://github.com/rancher/rancher/issues/6158

cjellick:
@LLParse how is what you are doing different from the the io.cattle.platform.core.constants.ServiceConstants.isSystem(Service) method itself?

PR #2155, opened 2016-11-02
The externalIdType in cattle schema for github org was wrong. 

prachidamle:
Fixed the value to be "github_org"

https://github.com/rancher/rancher/issues/6491


alena1108:
LGTM assuming the build is green

PR #2158, opened 2016-11-03
Generate service links for lbservice

alena1108:


alena1108:
https://github.com/rancher/rancher/issues/7747

PR #2161, opened 2016-11-03
WIP Java 8 Wedding

dx9:
Depends on https://github.com/rancher/cattle/pull/2153 and https://github.com/rancher/rancher/pull/6627.


ibuildthecloud:
@dx9 For a bit I've been wondering if http://www.azul.com/downloads/zulu/ Is a good option for a JDK.  It's just a "certified" build of openjdk from a real company that does crazy JVM things.  I started looking at it because the disk footprint was much smaller than an equivalent `apt-get` of openjdk Ubuntu.  But now that we are going to alpine that might not matter, but then I get a bit iffy about using an alpine "community" package.  Like I'd like a little more mainstream build of JDK than alpine.  My guess is the cross section of alpine users and Java deployments is a bit small.

dx9:
@ibuildthecloud Neat. There is the glibc issue with zulu on Alpine. Maybe they'd support a musl libc version.

I didn't think about the community thing at all, well didn't even notice that it is a community package. Wonder why it lives in the community repo. I'll look for an answer. FWIW There are Alpine images on the official image hub. https://hub.docker.com/r/library/openjdk/tags Pulls by tag would be interesting...

The work was done, so to me it's, what was changed, is musl libc support moving upstream, is support growing. Is anything broken. http://git.alpinelinux.org/cgit/aports/tree/community/openjdk8

I'll look through the patch set to try and understand the details. Whatever you guys are comfortable with works for me.

dx9:
@ibuildthecloud Started to setup a shadow PR based on zulu-openjdk so we can eval it. The first thing I've hit is having to install http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html. Well that's what the internet is telling me.

Could you look at the errors in https://drone.rancher.io/rancher/cattle/1269/2 and let me know if it's obvious to you. Picked them up with the madness and it's not obvious to me yet. Thanks!

PR #2163, opened 2016-11-04
Allow host resource fields on create

cjellick:
This lets the user set the milliCpu, localStorageMb, and memory fields
when creating a host via our machine integration.

The current agent backpopulate logic copies ALL fields over on the
first sync. So, this change also makes sure the relevant fields are removed
from the update map if they are already set on the host.


alena1108:
@cjellick just wanted to double check that fields updates/removals trigger metadata update - given that we propagate these fields to meta.

cjellick:
@alena1108 yes, by way of `MetadataProcessHandler`, which is on activate, update, remove of host.

alena1108:
LGTM

PR #2176, opened 2016-11-06
check for nonempty instead of null for link name on export

alena1108:
https://github.com/rancher/rancher/issues/6445


ibuildthecloud:
@alena1108 I feel like this PR should go green :)

PR #2179, opened 2016-11-07
WIP CI Zulu J8

dx9:
Depends on #2142 #2144


scripts/ci
```diff
@@ -135,13 +135,13 @@ test_failed()
...
 ./test-warm >/dev/null &
 
-CI=true MAVEN_ARGS='-B -q' run ./build
-
 threads=$(($(nproc) + 1))
 if [ $threads -gt 8 ]; then
     threads=8
 fi
 
+CI=true MAVEN_ARGS="-B -q -T $threads" run ./build
```
ibuildthecloud: This isn't stable.  I've tried it in the past and sometimes maven just randomly fails.  Maybe it's better in newer versions of maven?  I tried with one in ubuntu:14.04

dx9: Ya, seems to be better now. The multi threading warnings are gone during the build. It hasn't tripped up yet. It does deadlock nicely with checkstyle though.

dx9:
Using this image would let cattle move up to Java 8 and stay on trusty for a bit longer (delaying the db headaches). The tradeoff though is picking up an extra 130MB on disk in the rancher/server image.

I think the only way to avoid bloat in rancher/server is to migrate to Alpine. Which means db migration and other testing. I kinda like this until all the other work gets sorted out.

ibuildthecloud:
@dx9 Any chance you can do a corresponding PR to github.com/rancher/rancher/server?  Once we release Rancher 1.2 I can merge this.  As we are heavy into the test cycle of Rancher 1.2 we don't really want to upset anything at the moment.

dx9:
@ibuildthecloud yep https://github.com/rancher/rancher/pull/6627

ibuildthecloud:
@dx9 I finally merged https://github.com/rancher/cattle/pull/2142 so I'd like to merge this one next.  For maven installation though I was wondering if we could just use https://github.com/takari/maven-wrapper.  If we add the mvnw artifacts then the `mvnw` will handle installing maven.  Which is cool for local dev too.

dx9:
@ibuildthecloud Sure. I'll tidy this up and put the maven changes into a separate PR.

PR #2195, opened 2016-11-14
fix ulimits and isolation fields

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/6614

StrongMonkey:
https://github.com/rancher/rancher/issues/6587

StrongMonkey:
@alena1108  Could you please review this PR? tested locally, it should fix the issues.

cjellick:
@StrongMonkey please add corresponding test logic to test_transform_inspect.py

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
```diff
@@ -388,6 +390,38 @@ void setLogConfig(Instance instance, Map<String, Object> fromInspect) {
...
+        List<Ulimit> ret = new ArrayList<>();
+        if (ulimits instanceof List) {
+            for (Object ulimit : (List<Object>) ulimits) {
+                if (ulimit instanceof Map) {
+                    Ulimit l = new Ulimit();
+                    Map<String, Object> temp = (Map<String, Object>) ulimit;
+                    if (temp.get("Name") instanceof String) {
+                        l.setName(temp.get("Name").toString());
+                    }
+                    if (temp.get("Hard") instanceof Integer) {
```
cjellick: Can you use Number instead of integer here and below?
So, you'd check if it is an instance of Number and then cast it to a number and use Number's intValue() function to convert it to an int.

code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
```diff
@@ -388,6 +390,38 @@ void setLogConfig(Instance instance, Map<String, Object> fromInspect) {
...
+                    if (temp.get("Soft") instanceof Integer) {
+                        l.setSoft((Integer) temp.get("Soft"));
+                    }
+                    if (!l.getName().isEmpty()) {
+                        ret.add(l);
+                    }
+                }
+            }
+        }
+        setField(instance, FIELD_ULIMITS, ret);
```
cjellick: I think you should only set this field if ret is not empty. This is just my preference for not setting fields on the returned instance when they are empty lists.

StrongMonkey:
@cjellick PR updated

StrongMonkey:
@cjellick flake8 error corrected.

PR #2204, opened 2016-11-17
go-agent v0.2.5

cjellick:


cjellick:
Failure unrelated but addressed in https://github.com/rancher/cattle/pull/2205

PR #2211, opened 2016-11-18
Fix volume scheudling fixes

cjellick:
addresses https://github.com/rancher/rancher/issues/6661

### DIFF
--- a/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AbstractAllocator.java
+++ b/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AbstractAllocator.java
@@ -2,7 +2,6 @@ package io.cattle.platform.allocator.service;
 
 import static io.cattle.platform.core.model.tables.InstanceHostMapTable.*;
 import static io.cattle.platform.core.model.tables.PortTable.*;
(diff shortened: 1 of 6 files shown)


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AbstractAllocator.java
```diff
@@ -146,8 +149,12 @@ protected void deallocateInstance(final AllocationRequest request) {
     protected void allocateInstance(final AllocationRequest request) {
         final Instance origInstance = objectManager.loadResource(Instance.class, request.getResourceId());
         final List<Instance> instances = getInstancesToAllocate(origInstance);
+        final List<Long> volumeIds = new ArrayList<>();
+        for (Instance instance : instances) {
+            volumeIds.addAll(InstanceHelpers.extractVolumeIdsFromMounts(instance));
```
alena1108: Nit: would it make sense to make volumeIds a set? If the same volumeId can be referenced by diff instances, duplicates are possible

alena1108:
LGTM

cjellick:
Failures unrelated

PR #2213, opened 2016-11-18
Fix flaky test

cjellick:


tests/integration/cattletest/core/test_allocation.py
```diff
@@ -215,10 +215,13 @@ def test_simultaneous_port_allocation(new_context):
...
-        lambda: len(client.list_container(name_like=name, state='error')) > 0
-    )
-    return client.list_container(name_like=name, state='error')[0]
+
+    def check():
+        containers = client.list_container(name_like=name, state='error')
+        if len(containers) > 0:
+            return containers[0]
+    container = wait_for(check)
+    return container
```
cjellick: The old way created a gap between when the wait_for finished and actually getting the container. In the gap, it appears that the container could get to  `removed` and thus not shown in the last list_container call. So, this change just eliminates the gap.

alena1108:
LGTM

PR #2215, opened 2016-11-18
Webhook service launcher

mrajashree:
This PR adds launcher for Webhook Service

### DIFF
--- a/resources/content/schema/admin/admin-auth.json
+++ b/resources/content/schema/admin/admin-auth.json
@@ -254,7 +254,9 @@
         "task" : "r",
 
         "projectTemplate.isPublic" : "cru",
-
+        
+        "genericObject" : "crd",
+		
         "end" : ""
     }
 }
--- a/resources/content/schema/project/project-auth.json
+++ b/resources/content/schema/project/project-auth.json
@@ -84,6 +84,7 @@
         "balancerServiceTargetConfig" : "cr",
         "networkDriverService" : "crud",
         "storageDriverService" : "crud",
+        "genericObject" : "crd",
         "end" : ""
     }
 }
--- a/resources/content/schema/user/user-auth.json
+++ b/resources/content/schema/user/user-auth.json
@@ -802,6 +802,11 @@
         "serviceBinding" : "r",
         "serviceBinding.labels" : "cr",
         "serviceBinding.ports" : "cr",
+        
+        "genericObject" : "r",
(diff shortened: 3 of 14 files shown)


cjellick:
We still need to figure out how to make the receive URL unauthenticated and the generate authenticated.

mrajashree:
@cjellick will something similar to this work: https://github.com/rancher/cattle/blob/88724c28750e87ed9fb0a05dbf1e6360d246fa7d/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/request/handler/GenericWhitelistedProxy.java#L287

And for the construct handler in the service we can check for authorization header?

cjellick:
@mrajashree to allow the receiver to be unauthenticated, add it as a path here:
https://github.com/rancher/cattle/blob/eccd9c23c9fd6d2588ea7f9a0295066536b585bf/code/packaging/app-config/src/main/resources/META-INF/cattle/iaas-api/spring-iaas-api-context.xml#L529

resources/content/schema/user/user-auth.json
```diff
@@ -799,6 +799,10 @@
         "binding" : "r",
         "binding.services" : "cr",
 
+        "webhook" : "cr",
```
cjellick: webhook should be just "r" for user-ath

resources/content/schema/project/project-auth.json
```diff
@@ -84,6 +84,7 @@
         "balancerServiceTargetConfig" : "cr",
         "networkDriverService" : "crud",
         "storageDriverService" : "crud",
+        "webhook" : "crud",
```
cjellick: no "u", just "crd". No one should be able to update a webhook

resources/content/schema/base/webhook.json
```diff
@@ -0,0 +1,10 @@
+{
```
cjellick: So it doesn't work unless you add this? I expected it to work without this file.

resources/content/schema/admin/admin-auth.json
```diff
@@ -254,6 +254,7 @@
         "task" : "r",
 
         "projectTemplate.isPublic" : "cru",
+        "webhook" : "cru",
```
cjellick: no "u", just "crd". No one should be able to update a webhook

cjellick:
LGTM

PR #2216, opened 2016-11-19
External IdentityProvider should not throw client visible errors if searching/get identities fails

prachidamle:
External IdentityProvider should not throw client visible errors if searching/get identities fails, it should return empty/null response.

https://github.com/rancher/rancher/issues/6615


prachidamle:
@alena1108 can you review/merge

alena1108:
LGTM

PR #2223, opened 2016-11-19
Handle anonymous volumes with drivers

cjellick:
https://github.com/rancher/rancher/issues/6672

If a datavolume is added of the form /dir or /dir:ro and volumeDriver
is set, we need to create volume objects and add those volumes to
dataVolumeMounts.


ibuildthecloud:
@cjellick seems like legit failures

cjellick:
@ibuildthecloud test failures fixed

PR #2224, opened 2016-11-20
upgrading state as active state in account dao

alena1108:
@ibuildthecloud @prachidamle added "upgrading" to a bunch of places where account state is validated against "active"


alena1108:
that doesn't seem to fix the lb problem though..just tested it. May be there are some other places in rancher/microservice validating against the state being active?

alena1108:
ok, found one last place where the upgrading state was needed, added the check in, and lbs start fine for upgrading accounts.

PR #2226, opened 2016-11-20
project upgrade action should be available for project owners/admins only

prachidamle:
Added the capability to do project upgrade for owner only. 
Admins can also upgrade any project.

https://github.com/rancher/rancher/issues/6684


alena1108:
LGTM

PR #2228, opened 2016-11-21
When using api keys, the auth token is not present in the request, we should regenerate the token

prachidamle:
When using api keys, the auth token is not present in the request, need to get new token and read the identities. Fixed the refreshToken API to create the token correctly.

Verified this issue is not seen with new auth code:
https://github.com/rancher/rancher/issues/6568


prachidamle:
@alena1108 can you merge

ibuildthecloud:
@prachidamle How often do we get a new token?  Does this mean every API request will contact github?  If so that is too slow.

prachidamle:
@ibuildthecloud We will create a token that lasts for 16 hours, stored against the accountid of the given API keys.

The token created during login to UI is stored against the account of kind 'token' so that is not being useful when API keys are used. Hence need a separate token for keys. Again it will be reused for 16 hours.

PR #2246, opened 2016-11-23
Inject env lables and health check on upgrade for lb

alena1108:
just the way we do on create

https://github.com/rancher/rancher/issues/6744


alena1108:
TF is unrelated

PR #2251, opened 2016-11-26
Fix cleanup logic

cjellick:
For the specialized logic previously added, keep deleting after the first limit is hit.
Add the tables that have specialized logic back to the 'other' list because the framework uses this list to check for FK constraints.


cjellick:
adresses https://github.com/rancher/rancher/issues/6659

PR #2257, opened 2016-11-29
Remove service instance on failed start

alena1108:
https://github.com/rancher/rancher/issues/6803

"stop" container.events are coming for managed containers from host that is going down (wasn't the case before, but it's ok). So the containers are marked as stopped, then service.reconcile tries to start them again, on the same host, and it is stuck in Starting state forever. 

Changed instance.start to remove the container that fails to start, if it is a service container. That is to give service.reconcile a chance to recreate this container on another available non-disconnected host.

@ibuildthecloud let me know if you think it makes sense


cjellick:
LGTM

PR #2260, opened 2016-11-29
disallow changing the scale from fixed to global

alena1108:
https://github.com/rancher/rancher/issues/6812


alena1108:
TF is unrelated.

PR #2276, opened 2016-12-01
validate scale on service.update

alena1108:
allowing to update service.launchConfig.ports lets user update other launch config fields like labels (there seem to be no way to disable update for embedded fields of launchConfig). For now, added a check for scale validation on the update (global scale is passed via service.launchConfig.labels). We should think on how to disable it in the future.

https://github.com/rancher/rancher/issues/6812


cjellick:
LGTM
@alena1108 the bug is already assigned over to @soumyalj

PR #2280, opened 2016-12-03
exclude stack name from portRule.serviceId

alena1108:
when lb and target service are part of the same stack

https://github.com/rancher/rancher/issues/6829


cjellick:
LGTM

PR #2281, opened 2016-12-03
Add requestedHostId allocation query logic and remove volume allocation/deallocation

cjellick:


cjellick:
Addresses:
https://github.com/rancher/rancher/issues/6427

PR #2285, opened 2016-12-05
Fix v1 local-auth test

cjellick:
Fixed this in v2 tests, but not v1
see d099b0af440dbdb8359e947fdddc5b698cf8a272 for details on original
fix.


cjellick:
Implied LGTM from previous merge of sister change in https://github.com/rancher/cattle/pull/2271

PR #2290, opened 2016-12-07
network.hostPorts to false by default

alena1108:
@ibuildthecloud @joshwget 

I've checked that it is being explicitly passed as "true" in both ipsec and vxlan templates. @ibuildthecloud point if there are any other templates where this flag has to be defined.


deniseschannon:
https://github.com/rancher/rancher/issues/6958

PR #2292, opened 2016-12-09
Add windows as an orchestration type

cjellick:


cjellick:
offline LGTM from @ibuildthecloud

PR #2294, opened 2016-12-09
Fix a potential blocking issue if executor under water

GregoireW:
I got a issue if a docker service on a host restart or if some services (rancher/scheduler  ? other rancher service ) die / restart.

UI is then not responding ( the endpoint /v2-beta/identities  is not responding).  

I detected that it come from IdentityManager linked to the fact that the executor is under water (process 10 task when it receives 20 more...).  
To be able to access the UI, I fixed the IdentityManager not to wait for refresh issue.

This is related to https://github.com/rancher/rancher/issues/6995

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/IdentityManager.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/IdentityManager.java
@@ -26,6 +26,8 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
@GregoireW Thanks for this, but 10 seconds is two short.  It would have to be like 3 minutes.  Amazing to say, but some users have AD systems that literally takes minutes to iterate all the groups.  It's painfully slow, but hey, that's "enterprise" :).

Can you just increase the timeout?  The real root cause of this issue was fixed in #2295 but this PR is good defensive programming regardless.

GregoireW:
Hello,

The identities endpoint use the method, and UI is blocked until we got the
answer.  waiting 3 minutes to process 1 uid is too much.

If I understand correctly, this method is used by background process to
update user cache and also real time by UI.
Perhaps we should do something like
if only one user then synchronous call
 else asynchronous call with 3 minutes timeout per uid ?


2016-12-13 7:26 GMT+01:00 Darren Shepherd <notifications@github.com>:

PR #2296, opened 2016-12-09
rancher#6995

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/5885

PR #2304, opened 2016-12-12
Fix jwt token issues for java 8

cjellick:
A rebased version of https://github.com/rancher/cattle/pull/2144


cjellick:
A rebased version of https://github.com/rancher/cattle/pull/2144

PR #2308, opened 2016-12-13
Subtype genericObject as webhook

mrajashree:
This PR creates separate resource for webhook and adds fields for url and payload


cjellick:
LGTM

PR #2321, opened 2016-12-15
do not translate bind mounts to v2 volumes

alena1108:
https://github.com/rancher/rancher/issues/7070

@deniseschannon @joshwget


joshwget:
It looks like we actually do this in libcompose by checking if the left string is an absolute path (https://github.com/docker/libcompose/blob/master/lookup/file.go#L61). Can we do something like that for consistency? There seems to be a similar function in Java http://stackoverflow.com/a/28526097.

alena1108:
@joshwget sure, let me test it

alena1108:
@joshwget updated the PR + added extra verification to the test

joshwget:
LGTM!

PR #2322, opened 2016-12-15
Noop deployment unit when image is rancher/none

alena1108:
disregard container selector presence

https://github.com/rancher/rancher/issues/7001


cjellick:
LGTM

PR #2344, opened 2016-12-22
Fix whitelist proxy

cjellick:
Make GenericWhitelistedProxy name aware so that when ApiRequestFilterDelegate loads 
the list of handlers and orders it based on the api.request.handler.list property, the
NoAuthenticationProxy will be properly recognized.

Also, sneak in:
- A tweak to the url pattern for webhooks to make it catch everything under and including /v1-webhooks
- Some seemingly unneeded sets and checks of apiRequest.version

### DIFF
--- a/code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/servlet/ApiRequestFilterDelegate.java
+++ b/code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/servlet/ApiRequestFilterDelegate.java
@@ -84,7 +84,6 @@ public class ApiRequestFilterDelegate {
                 }
                 apiRequest.setSchemaFactory(defaultFactory);
                 schemaUrl = ApiContext.getUrlBuilder().resourceCollection(Schema.class);
-                apiRequest.setVersion(null);
             } else {
                 schemaUrl = ApiContext.getUrlBuilder().resourceCollection(Schema.class);
             }
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/request/handler/GenericWhitelistedProxy.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/request/handler/GenericWhitelistedProxy.java
@@ -178,7 +178,7 @@ public class GenericWhitelistedProxy extends AbstractResponseGenerator implement
     @SuppressWarnings("unchecked")
     @Override
     protected void generate(final ApiRequest request) throws IOException {
-        if (request.getVersion() == null || !ALLOW_PROXY.get())
+        if (!ALLOW_PROXY.get())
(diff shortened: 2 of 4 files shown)


code/framework/java-server/src/main/java/io/github/ibuildthecloud/gdapi/servlet/ApiRequestFilterDelegate.java
```diff
@@ -84,7 +84,6 @@ public ApiContext doFilter(ServletRequest request, ServletResponse response, Fil
                 }
                 apiRequest.setSchemaFactory(defaultFactory);
                 schemaUrl = ApiContext.getUrlBuilder().resourceCollection(Schema.class);
-                apiRequest.setVersion(null);
```
cjellick: Lines 80-84 set this field, why would we unset it here?

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/request/handler/GenericWhitelistedProxy.java
```diff
@@ -174,7 +174,7 @@ protected boolean isAllowed(HttpServletRequest servletRequest, String host) {
     @SuppressWarnings("unchecked")
     @Override
     protected void generate(final ApiRequest request) throws IOException {
-        if (request.getVersion() == null || !ALLOW_PROXY.get())
+        if (!ALLOW_PROXY.get())
```
cjellick: Assuming the above change to remove `apiRequest.setVersion(null);` is cool, request.getVersion() will never not be set.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/request/handler/GenericWhitelistedProxy.java
```diff
@@ -63,14 +64,15 @@
...
+import org.springframework.beans.factory.BeanNameAware;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.netflix.config.DynamicBooleanProperty;
 import com.netflix.config.DynamicStringListProperty;
 
 
-public class GenericWhitelistedProxy extends AbstractResponseGenerator {
+public class GenericWhitelistedProxy extends AbstractResponseGenerator implements BeanNameAware, Named {
```
cjellick: `BeanNameAware` will make spring call setBeanName when creating this bean and `Named` will make [CollectionUtils.orderList()](https://github.com/rancher/cattle/blob/9f9fdd94034fa51c5d1dceaf0fade90d7d7ff88f/code/framework/utils/src/main/java/io/cattle/platform/util/type/CollectionUtils.java#L134) consider the name instead of just the class name.

PR #2356, opened 2016-12-29
Metadata refactor sequel

alena1108:
Built on top of environment service links


alena1108:
On Rancher upgrade, metadata has to be upgraded prior to the rest of infrastructure services, and the upgrade should be mandatory. @deniseschannon we should think whether to document it, or put a note in the UI

alena1108:
WIP - merge only after metadata service is released! :)

PR #2365, opened 2017-01-10
WIP: scale policy db upgrade

alena1108:
DB migration part for https://github.com/rancher/rancher/issues/6832

As we've finally decided to get rid of scale policy from RAncher completely (including API and backend logic), we had to ensure that:

1) existing services having scale policy will continue to work
2) It is possible to upgrade from scale policy to scale (example - etcd upgrade as a part of k8s upgrade)

The cleaner way for it will be - transform lockedScale set by scalePolicy code, to scale on the upgrade

https://github.com/rancher/rancher/issues/7264


alena1108:
Closing this PR for now as scale policy will remain in cattle, it just has to be refactored

PR #2392, opened 2017-01-16
Deactivate service before deleting instance

cjellick:
This fixes a race condition where the backend is trying to put the service in a state of 'updating-active' and the api is tring to put it in a state of 'deactivating'. As previously decided, getting a non-200 response in this scenario is ok, so the cattle code doesn't need to change, just the test to avoid the scenario.

See this build failure for an example of the situation: https://drone.rancher.io/rancher/cattle/1980/1

### DIFF
--- a/tests/integration/cattletest/core/test_svc_discovery.py
+++ b/tests/integration/cattletest/core/test_svc_discovery.py
@@ -1457,10 +1457,10 @@ def test_global_service_update_label(new_context):
     instance2_host = instance2.hosts()[0].id
     assert instance2_host == host2.id
 
-    # destroy the instance, reactivate the service and check
-    # both hosts got instances
-    _instance_remove(instance1, client)
+    # deactivate the service, destroy the instance,
+    # reactivate the service, and check both hosts got instances
     service = wait_state(client, service.deactivate(), 'inactive')
+    _instance_remove(instance1, client)
     service = wait_state(client, service.activate(), 'active')
     instance1 = _validate_compose_instance_start(client, service, env, "1")
     instance2 = _validate_compose_instance_start(client, service, env, "2")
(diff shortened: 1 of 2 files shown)


tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -1457,10 +1457,10 @@ def test_global_service_update_label(new_context):
     instance2_host = instance2.hosts()[0].id
     assert instance2_host == host2.id
 
-    # destroy the instance, reactivate the service and check
-    # both hosts got instances
-    _instance_remove(instance1, client)
+    # deactivate the service, destroy the instance,
+    # reactivate the service, and check both hosts got instances
     service = wait_state(client, service.deactivate(), 'inactive')
+    _instance_remove(instance1, client)
```
alena1108: @cjellick this test seems is failing on drone. service.deactivate just schedules instances removal, but doesn't wait for stop to finish. So calling instance.remove on instance that is not stopped, results in error.

cjellick: @alena1108  The failure in this test (where it just times out waiting) is due to a bug that @ibuildthecloud wants to fix for 1.4.0.

Heres's the problem:
When moving a resource from a transitioning state to a resting state, we don't check what the transitioning state is.

So, the instance is in stopping, this code submits  a DELETE. The API call succeeeds and resource is moved to removing. BUT then stop finishes and just stomps on the state and moves it from removing to stopped.

Apparently this has been a bug for awhile but @ibuildthecloud did something to make it worse or make it happen more often. So, I'd like to keep this PR open until he fixes the bug so that we can retest after the fix.

cjellick:
Unneeded

PR #2395, opened 2017-01-18
Reworked db upgrade for volume.uri field

alena1108:
@ibuildthecloud long story.. so volume.uri field was shortened from 512 to 255 as a part of this [pr](https://github.com/rancher/cattle/pull/2209), but instead of creating new core file, the old one core-019 was modified. Then looks like on your branch you already had an old version of this file in DATABASE_CHANGE_LOG, so running codegen reset the field back to 512 in VolumeTable.

The fix is: proper upgrade + regenerated jooq

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/model/Volume.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/model/Volume.java
@@ -242,7 +242,7 @@ public interface Volume extends java.io.Serializable {
 	/**
 	 * Getter for <code>cattle.volume.uri</code>.
 	 */
-	@javax.persistence.Column(name = "uri", length = 512)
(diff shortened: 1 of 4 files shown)


alena1108:
on a second thought, thinking if shortening the uri as a part of the upgrade is a good idea as some users might have this value longer. Let me know, and I'll just include the jooq fixes where field is reset to 255.

ibuildthecloud:
https://github.com/rancher/cattle/blob/master/resources/content/db/mysql/mysql-dump.sql  that's probably the real culprit here.

alena1108:
@ibuildthecloud reworked the PR by changing the field's length in mysql-dump.sql and Jooq classes

alena1108:
Closing as it is going to be addressed as a part of different PR

PR #2398, opened 2017-01-18
WIP Support calling more than one external scheduler

cjellick:
Chain the schedulers one after another based on a priority label and
add a context property to the even sent to them that contains all the
instances being scheduled so that the event can be generic.

We need to change the logic for prioritizing schedulers. Because the value is a string, the sorting wont behave as expected for numbers. Also, the value field is not indexed, so it is not an optimal query.

I have some work started on a jooq result handler and comparator.

### DIFF
--- a/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/impl/AgentInstanceDaoImpl.java
+++ b/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/impl/AgentInstanceDaoImpl.java
@@ -14,6 +14,7 @@ import io.cattle.platform.core.dao.GenericResourceDao;
 import io.cattle.platform.core.model.Agent;
 import io.cattle.platform.core.model.Credential;
 import io.cattle.platform.core.model.Instance;
+import io.cattle.platform.core.model.tables.LabelTable;
 import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
 import io.cattle.platform.object.ObjectManager;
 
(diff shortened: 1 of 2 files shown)


cjellick:
@StrongMonkey i pushed an extra commit to this that was my work-in-progress for fixing the sorting. Its not finished and doesnt work, but it is how I want to do the sorting fix. Please make it work :-)

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/impl/AgentInstanceDaoImpl.java
```diff
@@ -63,19 +75,51 @@ public Instance getInstanceByAgent(Agent agent) {
...
                 .join(INSTANCE_LABEL_MAP)
                     .on(INSTANCE_LABEL_MAP.INSTANCE_ID.eq(INSTANCE.ID))
                 .join(LABEL)
                     .on(LABEL.ID.eq(INSTANCE_LABEL_MAP.LABEL_ID).and(LABEL.KEY.eq(providedServiceLabel)))
+                .leftOuterJoin(priority)
+                    .on(LABEL.ID.eq(INSTANCE_LABEL_MAP.LABEL_ID).and(LABEL.KEY.eq(priorityLabel)))
                 .where(INSTANCE.ACCOUNT_ID.eq(accountId)
                     .and(INSTANCE.AGENT_ID.isNotNull())
                         .and(INSTANCE.STATE.eq(InstanceConstants.STATE_RUNNING))
                         .and(INSTANCE.HEALTH_STATE.in(HealthcheckConstants.HEALTH_STATE_HEALTHY,
```
LLParse: @cjellick This imposes the requirement of a health check on any external scheduler. Is that OK?

I ask because the windows scheduler is running in host networking mode without a health check.

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/impl/AgentInstanceDaoImpl.java
```diff
@@ -63,19 +75,51 @@ public Instance getInstanceByAgent(Agent agent) {
...
                                 HealthcheckConstants.HEALTH_STATE_UPDATING_HEALTHY)))
-                .orderBy(INSTANCE.AGENT_ID.asc())
-                .fetch().intoArray(INSTANCE.AGENT_ID));
+                .orderBy(priority.VALUE.asc(), INSTANCE.AGENT_ID.asc())
```
LLParse: The way this is written, schedulers with a lower priority value with be given higher priority. Would it be more idiomatic to flip it?

deniseschannon:
Closing in favor of https://github.com/rancher/cattle/pull/2415

PR #2400, opened 2017-01-19
Network policy changes

leodotcloud:


alena1108:
checkstyle errors should be fixed + make sure test_authorization passes for both v1 and v2 api

alena1108:
We also need to add a python test. For create/update cases (test should cover all the fields including embedded)

vincent99:
- [ ] `groupBy` should be an enum
- [ ] There should be validation to require exactly one of {`from && to`, `within`, `between`} on a given `networkPolicyRule` so you can't create invalid rules.

code/iaas/logic/src/main/java/io/cattle/platform/process/network/DefaultNetworkProcess.java
```diff
@@ -61,6 +61,8 @@ protected void setDefaultNetwork(Long accountId) {
 
             if (CommonStatesConstants.ACTIVATING.equals(network.getState()) && newDefaultNetworkId == null) {
```
alena1108: looks like lines 62 and 64 can be merged

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/NetworkMetaData.java
```diff
@@ -1,22 +1,31 @@
...
+import io.cattle.platform.core.addon.NetworkPolicyRule;
+import io.cattle.platform.core.addon.NetworkPolicyRule.NetworkPolicyRuleAction;
+
 public class NetworkMetaData {
     String name;
     String uuid;
     boolean is_default;
     boolean host_ports;
     protected Map<String, Object> metadata;
+    NetworkPolicyRuleAction default_policy_action;
```
ibuildthecloud: Just change to string, and List<Object>.  Using actual types just add unneeded overhead in marshaller/unmarshaller I believe.

code/iaas/model/src/main/java/io/cattle/platform/core/addon/NetworkPolicyRule.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public enum NetworkPolicyRuleWithin {
+        stack, service, linked
+    }
+
+    NetworkPolicyRuleMember from, to, between;
+    NetworkPolicyRuleAction action;
+    NetworkPolicyRuleWithin within;
+    String[] ports;
```
alena1108: what is the format for the port value? Given it is string, not integer

alena1108: We should document it

tests/integration/cattletest/core/test_network_policy.py
```diff
@@ -0,0 +1,54 @@
...
+def test_default_network_policy(super_client):
+    nd = super_client.wait_success(super_client.create_network_driver())
+    network = super_client.create_network(isPublic=True,
+                                          networkDriverId=nd.id,
+                                          defaultPolicyAction='deny',
+                                          policy=['{"within": "stack", "action": "allow"}'])
+    network = super_client.wait_success(network)
+    assert network.state == 'active'
+    assert network.defaultPolicyAction == 'deny'
+    assert len(network.policy) == 1
```
alena1108: can we also add a validation for network.policy fields? Just to make sure its every field/value was set

alena1108:
Other than my minor comment on the test, LGTM. Assuming all the tests pass + we plan to add more validation handling in cattle later.

leodotcloud:
@alena1108 I have a lot of tests for the validation of policy on the micro service side. Will add the ones in cattle later.

leodotcloud:
@alena1108 @ibuildthecloud Once the PR is good to go, I would like to rebase, squash and resubmit the PR.

PR #2414, opened 2017-01-24
Filter out Windows hosts from healthcheck candidates

LLParse:


code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/impl/HealthcheckServiceImpl.java
```diff
@@ -259,17 +260,33 @@ protected void createHealthCheckHostMaps(HealthcheckInstanceType instanceType, l
                 HealthcheckInstance.class, healthInstance.getId());
         List<? extends Host> availableActiveHosts = allocatorDao.getActiveHosts(healthInstance.getAccountId());
         
+        // skip non-linux hosts
+        Iterator<? extends Host> it = availableActiveHosts.iterator();
+        while (it.hasNext()) {
+            Host host = it.next();
+            Map<String, Object> data = (Map<String, Object>) host.getData();
```
alena1108: @LLParse there is a slightly better way to retrieve field from the "data
 field of the object: https://github.com/rancher/cattle/blob/ffd037c0fd465ba3f5fa15d6dad0567f346ae5e2/code/iaas/logic/src/main/java/io/cattle/platform/process/host/HostLabelReconcile.java#L46

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/impl/HealthcheckServiceImpl.java
```diff
@@ -259,17 +260,33 @@ protected void createHealthCheckHostMaps(HealthcheckInstanceType instanceType, l
...
+            if (data.containsKey("fields")) {
+                Map<String, Object> fields = (Map<String, Object>) data.get("fields");
+                if (fields.containsKey("labels")) {
+                    Map<String, String> labels =(Map<String, String>) fields.get("labels");
+                    if (labels.containsKey("io.rancher.host.os") && !labels.get("io.rancher.host.os").equals("linux")) {
```
alena1108: can we declare "io.rancher.host.os" label in the SystemLabels class? and reference it from there

alena1108: also I'd rather define some special label for "skip from health checker" rather than implicitly deriving it from the io.rancher.host.os...And - that would be a good improvement - put a neq affinity to the health check service in the template so health check container doesn't get deployed on the host marked with "io.rancher.skip.healthcheck"

LLParse: Here are the associated changes:
https://github.com/LLParse/agent/blob/windows/core/hostInfo/os_c.go#L56-L58
https://github.com/LLParse/rancher-catalog/blob/windows/infra-templates/windows-healthcheck/0/docker-compose.yml#L7

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/impl/HealthcheckServiceImpl.java
```diff
@@ -259,17 +261,33 @@ protected void createHealthCheckHostMaps(HealthcheckInstanceType instanceType, l
                 HealthcheckInstance.class, healthInstance.getId());
         List<? extends Host> availableActiveHosts = allocatorDao.getActiveHosts(healthInstance.getAccountId());
         
+        // skip hosts labeled accordingly
+        Iterator<? extends Host> it = availableActiveHosts.iterator();
+        while (it.hasNext()) {
```
ibuildthecloud: io.cattle.platform.util.type.CollectionUtils.getNestedValue(Object, String...)` would make this a bit more straight forward.

alena1108:
LGTM

PR #2417, opened 2017-01-24
Changes to Cattle to pass apifilterproxy config parameters to websocket-proxy 

prachidamle:
Changes to Cattle to pass config parameters to websocket-proxy to configure the api filter proxy. Also fixed the reload method to call correct port.

https://github.com/rancher/rancher/issues/7134

### DIFF
--- a/code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/launch/WebsocketProxyLauncher.java
+++ b/code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/launch/WebsocketProxyLauncher.java
@@ -107,6 +107,7 @@ public class WebsocketProxyLauncher extends GenericServiceLauncher {
         env.put("PROXY_MASTER_FILE", MASTER_CONF);
         env.put("PROXY_CATTLE_ADDRESS", cattleProxyAddress);
         env.put("PROXY_HTTPS_PROXY_PROTOCOL_PORTS", getProxyProtocolHttpsPorts());
+        env.put("PROXY_API_FILTER_CONFIG_FILE", API_FILTER_PROXY_CONFIG_FILE.get());
 
         String processName = ManagementFactory.getRuntimeMXBean().getName();
         if (processName != null) {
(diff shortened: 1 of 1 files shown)


prachidamle:
@alena1108 please review

alena1108:
LGTM

code/implementation/docker/machine/src/main/java/io/cattle/platform/docker/machine/launch/WebsocketProxyLauncher.java
```diff
@@ -51,12 +51,14 @@
 
     protected void prepareConfigFile() throws IOException {
         String config = API_FILTER_PROXY_CONFIG.get();
+        String configContent;
         if (StringUtils.isBlank(config)) {
-            new File(API_FILTER_PROXY_CONFIG_FILE.get()).delete();
+            configContent = "{}";
```
ibuildthecloud: Your service shouldn't assume a file exists.  I'd rather we delete the file if the config is blank.

prachidamle: @ibuildthecloud what should the service do on Reload, when initially the config is not blank, but later the admin comes in and updates the config to be blank? Should the service continue with the original config since it could not reload any config file, or should it erase its state whenever the config file is not found?

ibuildthecloud: The service should treat a missing file the same as an empty config.  In go check for error is not found.

prachidamle: @ibuildthecloud  @alena1108 Ok, I updated the PR to not write the empty file. Instead reverted to deleting file as before
please review

prachidamle:
@ibuildthecloud @alena1108 please review again. Reverted the change to write empty config.

cjellick:
LGTM

PR #2418, opened 2017-01-24
do not ignore removed/removing stacks when calculate healthState

alena1108:
https://github.com/rancher/rancher/issues/7341


cjellick:
LGTM

PR #2427, opened 2017-01-25
Chain instance.purge to instance.remove

cjellick:


StrongMonkey:
https://github.com/rancher/rancher/issues/7484

ibuildthecloud:
LGTM

cjellick:
@vincent99 doesnt like that he doesnt get an event where the instance's state is removed (that event is skipped and you only get ones with state as purging or purged)

PR #2429, opened 2017-01-26
lb with default image can act as environment admin account

alena1108:
@cjellick could you review? looks like isLbSystemService check gone missing with multi role agent. We have this special handling just for lb service as it is not being deployed as a part of "system" catalog services, yet it requires a system role - it needs to fetch the certs.

https://github.com/rancher/rancher/issues/7555


cjellick:
@alena1108 yep, I may have lost that during a messy rebase.

cjellick:
LGTM

PR #2449, opened 2017-01-31
use hostip if 0.0.0.0 is returned

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/7600

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ContainerMetaData.java
```diff
@@ -178,7 +178,11 @@ public void setInstanceAndHostMetadata(Instance instance, HostMetaData hostMetaD
                     if (StringUtils.isEmpty(port.getIpAddress())) {
                         newPorts.add(hostIp + ":" + portObj);
                     } else {
```
cjellick: stupid nitpick: I think an `} else if (port.getIpAddress().equals("0.0.0.0")) {`
would be better.

Another (though slightly less stupid) nitpick: "0.0.0.0".equals(port.getIpAddress()) is safer as it protects against a null pointer when getIPAddress() returns null

StrongMonkey: roger roger!🤔

cjellick:
LGTM

PR #2450, opened 2017-02-01
Fix endpoint IP calculation

cjellick:
If the bind address on a port is '0.0.0.0', use the host's
IP address for public endpoints. This bug always existed if a user
were to explicitly set their IP on the port to 0.0.0.0, but no one
does that in real life. This is an issue now becauseexternal scheduler
explicity sets it to 0.0.0.0 for all ports that haven't explicity set
it.

for https://github.com/rancher/rancher/issues/7556
for https://github.com/rancher/rancher/issues/7605


alena1108:
LGTM

PR #2451, opened 2017-02-01
Fail svc create if out of environment ports

cjellick:
When a service does not specify a public port in its port map, cattle
will allocate an environment scoped port to it. Then every container
created for that service will get the same port exposed on the host.

Environment ports have a defined range and the range can be exhuasted.
The previous behavior when the range was exhausted was to simply allow
the service to be created with undefined behavior for what would happen
to the ports. With the new port allocation logic introduced by migrating
that logic to the external scheduler, the service containers were
getting a randomly assigned port. This is the correct behavior if a
container doesn't have a port, but it doesn't make sense for the
environment scoped ports of services, where consistency across
containers in a port is necessary.

This fixes the problem by failing to create the service if the port
range is exhausted. The service will remain in registering but will
have a transitioning message indicating the error.

for https://github.com/rancher/rancher/issues/7568


cjellick:
whoops. ahve to fix the v1 tests

alena1108:
@cjellick I wonder on allocate (n) ports failures, in addition to throwing an exception, should we also release the resources that were allocated? Or let them be allocated till the service is removed? If so, would it be easy to add a test checking that the ports are released on service removal?

cjellick:
Got a verbal LGTM from @alena1108. Going to merge

PR #2452, opened 2017-02-01
DONT-MERGE exploring broken test

cjellick:
Don't merge this, if the fix makes sense, we can merge https://github.com/rancher/cattle/pull/2451, as it has the same fix.


cjellick:
Closing in favor of https://github.com/rancher/cattle/pull/2451

PR #2456, opened 2017-02-02
Fix scheduler spreading

cjellick:
This fix addresses a bug where containers were not spreading at all
when the external scheduler was enabled. All containers would go to the
first host that the scheduler knew about.

The cause of this bug was that we were not sending the "phase" to the
external scheduler for the prioritize event. The prioritize event is
sent by cattle to ask for a prioritized list of hosts where the fist
item in the list is the best fit.

We added the phase logic to accomodate releasing ports on instance.stop.
The scheduler makes a decision on what resources to prioritize,
allocate, and deallocate based on the phase. If the phase is
instance.allocate or intance.deallocate, it will consider all possible
resources (container count, cpu, ports). If the phase is
instance.start|stop, it will just consider ports.

For the fix, I've refactor the logic to guarantee that a phase is
always required and set in a single location.

### DIFF
--- a/code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/SimpleAllocator.java
+++ b/code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/SimpleAllocator.java
@@ -149,11 +149,11 @@ public class SimpleAllocator extends AbstractAllocator implements Allocator, Nam
         Long agentId = getAgentResource(instance.getAccountId(), instances);
         String hostUuid = getHostUuid(instance);
         if (agentId != null && hostUuid != null) {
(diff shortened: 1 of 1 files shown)


cjellick:
for https://github.com/rancher/rancher/issues/7656

alena1108:
Is there a way to write an integration test for that? or external scheduler can be tested only on a real docker setup as a part of validation test? If the latter, LGTM

cjellick:
Yeah, I had a super cool integration test that mocked out the scheduler in python, but I had to disable it because it became to difficult to have the python based scheduler do the "right" thing when idempotency was turned on.

cjellick:
Merging per the above LGTM

PR #2457, opened 2017-02-02
parsing fix in default lb image image

alena1108:
https://github.com/rancher/rancher/issues/7655

to support docker:registry/repo/image:tag format in addition to just docker:repo/image:tag


cjellick:
LGTM

PR #2458, opened 2017-02-02
Fix 'secrets' volume allocation

cjellick:
Actually, I can add a test for this. Hold off on merging it until i get a test please


code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -164,14 +166,19 @@ public Condition getInstanceHostConstraint(Instance instance) {
     }
 
     @Override
-    public List<? extends Host> getHosts(StoragePool pool) {
+    public List<? extends Host> getHosts(Collection<? extends StoragePool> pools) {
+        Collection<Long> poolIds = new HashSet<>();
+        for (StoragePool p : pools) {
+            poolIds.add(p.getId());
+        }
```
alena1108: Minor: seems like we can return right away when pools/poolIds is empty

alena1108:
LGTM

PR #2460, opened 2017-02-03
Support legacy port allocation

cjellick:
Support the scenario where the user has upgraded rancher but not
upgraded the scheduler service. To do this, I've reintroduced the
previous port constraint logic and added a check to see if the
scheduler is of a version that is less than the version in which we
added port allocation logic.

### DIFF
--- a/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/AgentInstanceDao.java
+++ b/code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/AgentInstanceDao.java
@@ -9,7 +9,7 @@ public interface AgentInstanceDao {
 
     Agent getAgentByUri(String uri);
 
-    Instance getInstanceByAgent(Agent agent);
+    Instance getInstanceByAgent(Long agentId);
 
     boolean areAllCredentialsActive(Agent agent);
 
--- a/code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/SimpleAllocator.java
+++ b/code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/SimpleAllocator.java
@@ -4,12 +4,14 @@ import io.cattle.platform.agent.AgentLocator;
 import io.cattle.platform.agent.RemoteAgent;
 import io.cattle.platform.agent.instance.dao.AgentInstanceDao;
 import io.cattle.platform.allocator.constraint.Constraint;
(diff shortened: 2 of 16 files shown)


code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/SimpleAllocator.java
```diff
@@ -84,6 +91,11 @@
     @Inject
     GenericMapDao mapDao;
 
+    private String cachedRequiredVersion;
+    private Integer requiredMajor;
+    private Integer requiredMinor;
+    private Integer requiredPatch;
```
alena1108: we probably shouldn't worry about patch version. Unless we are planning to re-use this code for other version checks (in this case, it would probably make sense to add it to some utils class)

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/AgentInstanceDao.java
```diff
@@ -9,7 +9,7 @@
 
     Agent getAgentByUri(String uri);
 
-    Instance getInstanceByAgent(Agent agent);
+    Instance getInstanceByAgent(long agentId);
```
alena1108: just being paranoid - is there a chance this method is called based on instance.getAgentId() // where getAgentId can potentially return null?

code/iaas/simple-allocator/src/main/java/io/cattle/platform/simple/allocator/SimpleAllocator.java
```diff
@@ -435,4 +450,76 @@ private ResourceRequest populateResourceRequestFromInstance(Instance instance, S
...
+            actualMajor = Integer.valueOf(majorTemp).intValue();
+            actualMinor = Integer.valueOf(actualParts[1]).intValue();
+        } catch (NumberFormatException e) {
+            // Image is not following semantic versioning. Assume custom, don't use legacy
+            return false;
+        }
+
+        if (actualMajor < requiredMajor) {
+            return true;
+        } else if (actualMajor == requiredMajor && actualMinor < requiredMinor) {
```
alena1108: we should perhaps use .intValue() when do Integer comparison. Or Integer1.equals(Integer2)

cjellick:
https://github.com/rancher/rancher/issues/7681

deniseschannon:
https://github.com/rancher/rancher/issues/7680

PR #2466, opened 2017-02-08
don't return scheduler agent if host is reconnecting or disconnected

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/7729

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/impl/AgentInstanceDaoImpl.java
```diff
@@ -95,11 +97,16 @@ public boolean areAllCredentialsActive(Agent agent) {
...
+                .join(INSTANCE_HOST_MAP)
+                    .on(INSTANCE.ID.eq(INSTANCE_HOST_MAP.INSTANCE_ID))
+                .join(HOST)
+                    .on(INSTANCE_HOST_MAP.HOST_ID.eq(HOST.ID))
                 .where(INSTANCE.ACCOUNT_ID.eq(accountId)
                     .and(INSTANCE.AGENT_ID.isNotNull())
                         .and(INSTANCE.STATE.eq(InstanceConstants.STATE_RUNNING))
                         .and(INSTANCE.HEALTH_STATE.in(HealthcheckConstants.HEALTH_STATE_HEALTHY,
                                 HealthcheckConstants.HEALTH_STATE_UPDATING_HEALTHY)))
+                        .and(HOST.AGENT_STATE.isNull().or(HOST.AGENT_STATE.notIn(AgentConstants.STATE_DISCONNECTED, AgentConstants.STATE_DISCONNECTING, AgentConstants.STATE_RECONNECTING)))
```
cjellick: host.agent_state is not good enough as it is not entirely accurate. Also, we just want to target state=active. Here's what we need overall:
```
select
...
join instance_host_map
join host
join agent on host.agent_id = id
where
agent.state == 'active'
```

Can you also add a test that exercises this? the test can be in test_resource_based_scheduler and should do the folowing:
1. Use a new_context
2. Add a container to that context that will register itself as an external scheduler using this logic: https://github.com/rancher/cattle/blob/d35dcbebb23d99b4b619ad0ff2955b0e2f7102d2/tests/integration/cattletest/core/test_resource_based_scheduling.py#L218-L226
3. Delete the host that the container was deployed on
4. Add a new host to the context
5. Ensure you can deploy a container

cjellick:
@StrongMonkey Also test the scenario manually with real hosts

tests/integration/cattletest/core/test_resource_based_scheduling.py
```diff
@@ -95,6 +95,29 @@ def test_resource_based_scheduler(new_context, super_client):
...
+    client.wait_success(host3.deactivate())
+    client.wait_success(super_client.delete(host1))
+    client.wait_success(super_client.delete(host2))
+    client.wait_success(super_client.delete(host3))
+
+    register_simulated_host(new_context)
+    c = client.create_container({'imageUuid': image, 'networkMode': 'host'})
+    c = client.wait_success(c)
+    assert c.state == 'running'
+    c.stop(remove=True)
```
cjellick: remove this line. unneeded

code/iaas/agent-instance/src/main/java/io/cattle/platform/agent/instance/dao/impl/AgentInstanceDaoImpl.java
```diff
@@ -95,11 +97,18 @@ public boolean areAllCredentialsActive(Agent agent) {
...
+                .join(HOST)
+                    .on(INSTANCE_HOST_MAP.HOST_ID.eq(HOST.ID))
+                .join(AGENT)
+                    .on(AGENT.ID.eq(HOST.AGENT_ID))
                 .where(INSTANCE.ACCOUNT_ID.eq(accountId)
                     .and(INSTANCE.AGENT_ID.isNotNull())
                         .and(INSTANCE.STATE.eq(InstanceConstants.STATE_RUNNING))
                         .and(INSTANCE.HEALTH_STATE.in(HealthcheckConstants.HEALTH_STATE_HEALTHY,
                                 HealthcheckConstants.HEALTH_STATE_UPDATING_HEALTHY)))
+                        .and(AGENT.STATE.eq(AgentConstants.STATE_ACTIVE))
```
cjellick: Use CommonStateConstants.ACTIVE

code/iaas/model/src/main/java/io/cattle/platform/core/constants/AgentConstants.java
```diff
@@ -4,6 +4,7 @@
 
     public static final String TYPE = "agent";
     public static final String USER = "user";
+    public static final String STATE_ACTIVE = "active";
```
cjellick: Use CommonStateConstants

tests/integration/cattletest/core/test_resource_based_scheduling.py
```diff
@@ -95,6 +95,29 @@ def test_resource_based_scheduler(new_context, super_client):
                              'type': 'computePool'}])
 
 
+def test_delete_host(new_context, super_client):
+    host1 = new_context.host
+    host2 = register_simulated_host(new_context)
```
cjellick: No need to add host2 or host3. A single host is fine for the test

tests/integration/cattletest/core/test_resource_based_scheduling.py
```diff
@@ -95,6 +95,29 @@ def test_resource_based_scheduler(new_context, super_client):
...
+    host3 = register_simulated_host(new_context)
+
+    mock_sched(new_context, super_client)
+    image = new_context.image_uuid
+    client = new_context.client
+
+    client.wait_success(host1.deactivate())
+    client.wait_success(host2.deactivate())
+    client.wait_success(host3.deactivate())
+    client.wait_success(super_client.delete(host1))
```
cjellick: You should not need super_client, you just need to get the updated version of the host. Can do it like this:
host1 = client.wait_success(host1.deactivate())
host1 = client.wait_success(client.delete(host1))
because wait_success returns the reloaded version of host

tests/integration/cattletest/core/test_resource_based_scheduling.py
```diff
@@ -95,6 +95,29 @@ def test_resource_based_scheduler(new_context, super_client):
                              'type': 'computePool'}])
 
 
+def test_delete_host(new_context, super_client):
```
cjellick: Did you verify that this test fails with the old code?
I just want to make sure that we know this test fails without the fix and passes with the fix. (You should do this with all bug fix tests actually)

StrongMonkey: I didn't. Seems it will pass also. The test is not actually exercising the logic we wrote. When we deleted the sim host, the container gets deleted also. Need to think about another approach to fake the scenario we encountered.

StrongMonkey:
@cjellick updated.

cjellick:
LGTM

PR #2477, opened 2017-02-13
Container name unique per environment

alena1108:
https://github.com/rancher/rancher/issues/7384

@ibuildthecloud @vincent99 still uncertain whether putting this restriction on API side is correct, given that internally we do allow duplicated names (during in service upgrade use case, failed health check instance recreate, etc)

### DIFF
--- a/resources/content/schema/base/container.json
+++ b/resources/content/schema/base/container.json
@@ -4,6 +4,7 @@
             "type": "map[string]",
             "create": true
         },
+        
         "command": {
             "type": "array[string]",
             "nullable": true,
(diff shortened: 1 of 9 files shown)


ibuildthecloud:
I think it's fine to attempt to prevent this knowing that we really can't stop it fully.

cjellick:
What's the point of disallowing it? Are we still going to allow null names?

vincent99:
Containers can go on stacks which now have to have unique (and not null) names for DNS

cjellick:
What do you mean "can go on stacks" @vincent99?

alena1108:
@vincent99 I didn't put not null restriction on the container name yet, should I do it?

PR #2479, opened 2017-02-14
compare labels keys in case insensitive manner

alena1108:
https://github.com/rancher/rancher/issues/6980

@cjellick


cjellick:
LGTM

PR #2482, opened 2017-02-15
bug fix in new metadata service link generation

alena1108:
https://github.com/rancher/rancher/issues/7819

treat empty name as same as null


cjellick:
Lgtm

PR #2493, opened 2017-02-16
wsp v0.20.0

cjellick:


cjellick:
Going to merge this. Drone is whack right now

PR #2498, opened 2017-02-17
tweak system label

StrongMonkey:


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -40,6 +40,9 @@
 
 @Named
 public class InstancePreCreate extends AbstractObjectProcessLogic implements ProcessPreListener, Priority {
+    
+    private static final String SYSTEM_LABEL = "io.rancher.container.system";
```
cjellick: move to SystemLabels

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -65,6 +68,7 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
         setDns(instance, labels, data);
         setLogConfig(instance, data);
         setSecrets(instance, data);
+        setSystemLabel(instance, data);
```
cjellick: pass in existing labels

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -156,6 +158,13 @@ protected void setDns(Instance instance, Map<String, Object> labels, Map<Object,
             }
         }
     }
+    
+    protected void setSystemLabel(Instance instance, Map<String, Object> labels, Map<Object, Object> data) {
+        if(instance.getSystem()) {
```
cjellick: Since getSystem() returns a Boolean oject and not the primitive boolean, this can cause a null pointer exception.

Protect against it by making the check `if (Boolean.TRUE.equals(instance.getSystem())) { ... }`

cjellick:
@StrongMonkey we need a test that exercise the instance logic. To test, create a system stack like this:
https://github.com/rancher/cattle/blob/454993ee3353257e44a95ac45c13d688a13a1487/tests/integration/cattletest/core/common_fixtures.py#L223

Then add a service to that stack that has a label. Activate the service, get the resulting container and assert that it has the label you specified when creating the service and the one we're adding in InstancePreCreate.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/impl/AllocatorServiceImpl.java
```diff
@@ -221,10 +221,6 @@ public void volumeDeallocate(Volume volume) {
                     List<String> newHosts = (List<String>) CollectionUtils.getNestedValue(eventResult.getData(), SCHEDULER_PRIORITIZE_RESPONSE);
                     hosts.retainAll(newHosts);
                 }
-
```
cjellick: change name of method to callExternalSchedulerForHostsSatisfyingLabels

StrongMonkey:
https://github.com/rancher/rancher/issues/7871

cjellick:
Shoot, this PR is good to go, but adding the io.rancher.container.system label makes cattle incompatible with the network manager that shipped in 1.4 because the mere presence of that label causes the network-manager to not setup resolv.conf

cjellick:
LGTM

PR #2503, opened 2017-02-17
compare labels values in case insensitive manner

alena1108:
https://github.com/rancher/rancher/issues/7874

@cjellick could you merge only if allocator doesn't care about the case of the label value. Otherwise we'd have to fix it differently, by re-creating the label mapping when the case of the old/new value doesn't match


cjellick:
Is case-insensitive in allocator:
https://github.com/rancher/cattle/blob/bad3b581dcac43985b7f1ae0dba93087964a00d0/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java#L482-L483

cjellick:
LGTM

PR #2505, opened 2017-02-21
Add host event create for scheduler update

cjellick:
We need to be able to ensure that when a host is added, removed, or
updated, global service reconciliation is kicked off AFTER the external
scheduler has had a chance to update its internal state.

This event handler allows for that. The flow for CRUDing the host will
be thus:
1. Host is CRUDed in cattle
2. Existing service reconcile trigger logic runs, but the host change
has not yet propogated to metadata and thus it is not in the scheduler,
so the scheduler behaves as though it doesnt know about the change and
global services are not changed.
3. Metadata updates and scheduler updates its internal state
4. Scheduler sends a external host event informing cattle that it
updated its state and global service reconciliation needs to be
triggered.
5. ExternalHostSchedulerUpdateEventCreate handles the event and kicks
off reconcile for global services.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalHostEventCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalHostEventCreate.java
@@ -136,7 +136,7 @@ public class ExternalHostEventCreate extends AbstractObjectProcessHandler {
 
     @Override
     public String[] getProcessNames() {
-        return new String[] {ExternalEventConstants.KIND_EXTERNAL_EVENT + ".create"};
+        return new String[] {ExternalEventConstants.PROCESS_EXTERNAL_EVENT_CREATE};
     }
 
 }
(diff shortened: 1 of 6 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ExternalHostSchedulerUpdateEventCreate.java
```diff
@@ -0,0 +1,54 @@
...
+
+        return null;
+    }
+
+    public void setLookup(ServiceLookup lookup) {
+        this.lookup = lookup;
+    }
+
+    @Override
+    public String[] getProcessNames() {
```
alena1108: do you think it would make sense to add it to the list of service reconcile triggers instead of creating a separate file? https://github.com/rancher/cattle/blob/master/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/ServicesReconcilePostTrigger.java

alena1108:
LGTM

cjellick:
@ibuildthecloud noticed nullpointerexceptions. @StrongMonkey tested end-to-end and we realized that we needed to give agent permissions to create this type of event.

I addressed both, but the PR is still fundamentally the same as it was when @alena1108 lgtm'd. so, i'm going to merge

StrongMonkey:
https://github.com/rancher/rancher/issues/7870

PR #2515, opened 2017-02-23
Fix secrets test

cjellick:
The test is checking the length of the payload, but the expected length
was based off of a previous version of the secrets binary.


cloudnautique:
LGTM. Secrets API changed rewrapText format.

PR #2521, opened 2017-02-27
fix native container for docker compose

StrongMonkey:
@cjellick It appears that containers created by docker-compose always have a weird network name called `projectDir-default`, which cattle can't find it in database and will schedule instance.remove for them https://github.com/rancher/cattle/blob/master/code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java#L61-L66 if network is null. ~~My change is just to go through all the logic in `resolveNetwork` and then if still can't find network, return the default `dockerBridge` because network `projectDir-default`is still valid and is treated as docker bridge indeed. But this can't cover all the cases as we can't tell the network type just from the name. However, it will be sufficient to let those containers go through the rancher container lifecycle~~ Tested end to end, it works.

### DIFF
--- a/code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
+++ b/code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
@@ -59,7 +59,7 @@ public class DockerInstancePreCreate extends AbstractObjectProcessLogic implemen
         data.put(DockerInstanceConstants.FIELD_NETWORK_MODE, mode);
 
         Network network = networkService.resolveNetwork(instance.getAccountId(), mode);
-        if (network == null && StringUtils.isNotBlank(mode)) {
+        if (network == null && StringUtils.isNotBlank(mode) && !instance.getNativeContainer()) {
             objectProcessManager.scheduleProcessInstance(InstanceConstants.PROCESS_REMOVE, instance, null);
             throw new ExecutionException(String.format("Failed to find network for networkMode %s", mode),
                     null, state.getResource());


StrongMonkey:
https://github.com/rancher/rancher/issues/6911

StrongMonkey:
https://github.com/rancher/rancher/issues/7941

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
```diff
@@ -59,11 +56,16 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
         data.put(DockerInstanceConstants.FIELD_NETWORK_MODE, mode);
 
         Network network = networkService.resolveNetwork(instance.getAccountId(), mode);
+        // I don't have too much confidence about deleting this block of code, but there is a scenario which mode isn't belong to any of 
+        // [bridge,host,container,managed]. This happens when we create containers using docker-compose, network mode return by docker 
+        // inspect is actually something like [projectDir-default], which equals to docker bridge
+        /*
         if (network == null && StringUtils.isNotBlank(mode)) {
```
ibuildthecloud: Just add a check to only do this for `instance.isNativeContainer`.  So if it's backported we just leave it as is, it it's deployed from rancher we delete.  We need this check to make sure a container doesn't deploy for the ipsec/vxlan network before the network has been created.

StrongMonkey: yeah, will do. I previously thought this class is only called for native containers.

cjellick:
LGTM

thebalaa:
How can update my rancher agent to get this fix?

vincent99:
@thebalaa this will be in the agent for rancher 1.5.0

PR #2523, opened 2017-02-27
don't override userPort field

StrongMonkey:
It will address the issue where we restart the container and then the userPort field is overridden. https://github.com/rancher/rancher/issues/7782


code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instancehostmap/DockerPostInstanceHostMapActivate.java
```diff
@@ -65,7 +65,7 @@
         Priority {
 
     private static final Logger log = LoggerFactory.getLogger(DockerPostInstanceHostMapActivate.class);
-
+
```
ibuildthecloud: nit: Can you removed the whitespace change?

StrongMonkey: done!

PR #2526, opened 2017-02-28
NPE fix in container metadata generation

alena1108:
https://github.com/rancher/rancher/issues/7952


cjellick:
LGTM, assuming green
https://github.com/rancher/rancher/issues/7952

PR #2550, opened 2017-03-15
Fixed standalone container default serviceIndex value

alena1108:
https://github.com/rancher/rancher/issues/8185

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
@@ -95,7 +95,7 @@ public class InstancePreCreate extends AbstractObjectProcessLogic implements Pro
         if (instance.getName() != null) {
             return;
         }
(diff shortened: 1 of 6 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -95,7 +95,7 @@ protected void setName(Instance instance, Map<Object, Object> data) {
         if (instance.getName() != null) {
             return;
         }
-        String name = "r-" + instance.getUuid().substring(0, 4);
+        String name = "r-" + instance.getUuid().substring(0, 10);
```
cjellick: is this change unrelated to the serviceIndex issue? just want to make sure i understand

alena1108: @cjellick unrelated, forgot to mention - we generate container name if passed as null via API. Increased the size that we strip off the UUID to reduce potential chances for duplicates.

cjellick:
LGTM

PR #2561, opened 2017-03-20
increased label.value field size

alena1108:
https://github.com/rancher/rancher/issues/8168

@ibuildthecloud @LLParse apparently k8s puts a lot of data in stateful set pod container's label. Increased the label.value size in cattle db to mediumtext to fix it.

### DIFF
--- a/resources/content/db/changelog.xml
+++ b/resources/content/db/changelog.xml
@@ -124,4 +124,5 @@
     <include file="db/core-117.xml"/>
     <include file="db/core-118.xml"/>
     <include file="db/core-119.xml"/>
+    <include file="db/core-120.xml"/>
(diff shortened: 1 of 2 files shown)


LLParse:
64Kb should certainly be enough... @vincent99 is there any problem with doing this?

ibuildthecloud:
mediumtext is 64mb

LLParse:
Whoops, was looking at mysql text (64kb). mysql mediumtext appears to be 16mb

ibuildthecloud:
Just going w/ truncating in https://github.com/rancher/cattle/pull/2668.  The main reason for these tables is indexing the values.  So we really aren't too concerned with search for labels longer than 4k.

PR #2567, opened 2017-03-21
ebs-volume fix

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/6714

StrongMonkey:
This PR will lock the volume name on allocation phase so that the instance could wait until the first one is allocated. Once the first one is allocated the other will land on the same host per volume constraints. This could potentially prevents instances which use the same environment volume going to different hosts.

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/impl/AllocatorServiceImpl.java
```diff
@@ -348,6 +348,15 @@ protected LockDefinition getInstanceLockDef(Instance origInstance, List<Instance
         for (Volume v : volsToLock) {
                 locks.add(new AllocateConstraintLock(AllocateConstraintLock.Type.VOLUME, v.getId().toString()));
         }
+        
+        // lock the volume name if the driver is rancher-ebs
+        List<Volume> volumes = InstanceHelpers.extractVolumesFromMounts(origInstance, objectManager);
+        for (Volume volume: volumes) {
+            String driver = DataAccessor.fieldString(volume, VolumeConstants.FIELD_VOLUME_DRIVER);
+            if (driver.equals("rancher-ebs")) {
```
ibuildthecloud: Somewhere a tiny little java programming fairy just died.

ibuildthecloud:
@StrongMonkey I think the fix needs to be something like, for unallocated volumes lock.  @cjellick thoughts?

StrongMonkey:
@ibuildthecloud Yeah, that makes sense. Once it is allocated we don't need to prevent race condition.

StrongMonkey:
Ok, I am back. So the lock should lock on the **unallocated** **singleHostRW** volume ID because if it is unallocated, the singleHostRW is empty and if multiple allocation requests come in at the same time they may break the constraints.

cjellick:
Yeah, that sounds right. Also do it for the other access mode...its soemthing like singleContainerRW or something like that.

________________________________

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/impl/AllocatorServiceImpl.java
```diff
@@ -348,6 +350,17 @@ protected LockDefinition getInstanceLockDef(Instance origInstance, List<Instance
         for (Volume v : volsToLock) {
                 locks.add(new AllocateConstraintLock(AllocateConstraintLock.Type.VOLUME, v.getId().toString()));
         }
+        
+        List<Volume> volumes = InstanceHelpers.extractVolumesFromMounts(origInstance, objectManager);
+        for (Volume volume: volumes) {
+            StorageDriver driver = objectManager.loadResource(StorageDriver.class, volume.getStorageDriverId());
+            String accessMode = DataAccessor.fieldString(driver, StorageDriverConstants.FIELD_VOLUME_ACCESS_MODE);
+            if (accessMode.equals(VolumeConstants.ACCESS_MODE_SINGLE_HOST_RW) || accessMode.equals(VolumeConstants.ACCESS_MODE_SINGLE_INSTANCE_RW)) {
+                if (!volume.getAllocationState().equals("active")) {
```
ibuildthecloud: Can you switch to CommonStates.ACTIVE, also as a little trick in java when doing string compare against constants reverse the expression.  So

```java
CommonState.ACTIVE.equals(volume.getAllocationState())
```
Doing that you never have to worry about a null pointer exception.

StrongMonkey: got it!

LLParse:
@cjellick @ibuildthecloud Would it make sense to lock regardless of access mode? Consider the following race for multiHostRW:

1. User starts up rancher-nfs
2. User creates a vol `foo` in the UI
3. User starts a service scale=3 and container anti-affinity (spread across hosts) that attempt to bind-mount volume `foo`.

The end result would be 3 separate volume create requests occurring at the same time, assuming the Rancher DVP integration hasn't changed dramatically since I last played with the storage plugins.

LLParse:
I should mention that EBS create is _not_ idempotent at this point - not sure about the others. But the implication here is we need to make create idempotent if create() might run concurrently for the same volume.

cjellick:
Regarding @LLParse's comment: The scheduler, and in particular these locks, are not responsible for locking actual create logic. They only lock the decision making process on where a volume should land. We want it to be as concurrent as possible. The nfs scenario you described must be handled by the storage driver, not cattle. convoy accomplished this using `flock` when CRUDing volumes

cjellick:
But @LLParse's comment regarding create not being idempotent concers me. I need to think through the implications of that

cjellick:
@StrongMonkey  since you cannot test this in cattle, it should have a corresponding validation test. Please add the test and reference the validation test PR in here

StrongMonkey:
@cjellick this PR https://github.com/rancher/validation-tests/pull/295 will include the test cases for EBS volume. I am still adding test to it and will cover this race condition test case later on.

cjellick:
Actually, you can add an integration test for this. We do tests for faked storage pools of this kind in `tests/integration/cattletest/core/test_shared_volumes.py`.
Just search for singleHostRW

StrongMonkey:
@cjellick I am not sure writing integration test is a good idea. This is a race-condition bug and i don't see the integration test hits it(since they are not doing real action as they are creating simulated containers and volumes).

cjellick:
Simulated containers would have no impact on whether this bug gets hit or not. The allocator doesn't care if they are real or fake. I really prefer to see integration tests wherever possible. They are run much more frequently and catch regressions much more quickly. You are just writing tests to prove you're code works today, you're writing tests to make sure no one breaks it in the future.

I'd still like to see an integration test. I can help you with writing it after I get through the other PRs.

StrongMonkey:
@cjellick  ok, thinking.... I ll find a way to reliably reproduce it in the integration test

cjellick:
You dont need to reliably reproduce the race condition...you just need to recreate the scenario that would cause it, which is to create a service with scale > 1 using a an environment-scoped volume that belongs to a signle-host-rw storage pool in an environment with more than one host

cjellick:
Still failed flake8 @StrongMonkey. plz fix

StrongMonkey:
@cjellick now it should be fixed

cjellick:
LGTM

PR #2577, opened 2017-03-24
WIP several fixes around stdandalone container upgrade

alena1108:
1) Previous revision to instance
2) TBD - @joshwget this will be the fix for the issue you saw when revision id wasn't returned to you on the instance.upgrade. Marking WIP till we figure the issue out.


alena1108:
Addressed as a part of https://github.com/rancher/cattle/pull/2560, closing

PR #2582, opened 2017-03-27
Delete token on logout

mrajashree:
https://github.com/rancher/rancher/issues/7581

Validation test PR: https://github.com/rancher/validation-tests/pull/297

### DIFF
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
@@ -16,6 +16,7 @@ import io.cattle.platform.object.ObjectManager;
 import io.cattle.platform.object.util.DataAccessor;
 import io.cattle.platform.token.TokenException;
 import io.cattle.platform.token.TokenService;
+import io.cattle.platform.util.type.CollectionUtils;
 import io.github.ibuildthecloud.gdapi.context.ApiContext;
 import io.github.ibuildthecloud.gdapi.exception.ClientVisibleException;
 import io.github.ibuildthecloud.gdapi.request.ApiRequest;
@@ -396,4 +397,46 @@ public abstract class AbstractTokenUtil implements TokenUtil {
         return UNRESTRICTED_ACCESSMODE.equalsIgnoreCase(accessMode);
     }
 
+    public Identity jsonToIdentity(Map<String, Object> jsonData) {
+        String externalId = ObjectUtils.toString(jsonData.get("externalId"));
+        String externalIdType = ObjectUtils.toString(jsonData.get("externalIdType"));
+        String name = ObjectUtils.toString(jsonData.get("name"));
+        String profilePicture = ObjectUtils.toString(jsonData.get("profilePicture"));
+        String profileUrl = ObjectUtils.toString(jsonData.get("profileUrl"));
+        String login = ObjectUtils.toString(jsonData.get("login"));
+        return new Identity(externalIdType, externalId, name, profileUrl, profilePicture, login);
+    }
+
+    public Token getUserIdentityFromJWT() {
+        Token token = new Token();
+        Map<String, Object> jsonData = getJsonData();
+        if (jsonData == null) {
+            return null;
+        }
+
+        Object idObject = jsonData.get(USER_IDENTITY);
+        if (idObject != null) {
+            Map<String, Object> idMap = CollectionUtils.toMap(idObject);
+            Identity userIdentity = jsonToIdentity(idMap);
+            String userType = ObjectUtils.toString(jsonData.get(USER_TYPE), null);
+            token.setUserIdentity(userIdentity);
+            token.setUserType(userType);
+        }
+        return token;
+    }
+
+    public Token retrieveCurrentToken() {
+        Token token = new Token();
+
+        if (findAndSetJWT()) {
+            Token userToken = getUserIdentityFromJWT();
+            if(userToken != null) {
+                token.setUserIdentity(userToken.getUserIdentity());
+                token.setUserType(userToken.getUserType());
+                token.setJwt(getJWT());
+            }
+        }
+        log.debug("retrieveCurrentToken returning {}", token);
+        return token;
+    }
 }
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
@@ -2,11 +2,12 @@ package io.cattle.platform.iaas.api.auth.identity;
 
 import io.cattle.platform.api.auth.Identity;
 import io.cattle.platform.api.auth.Policy;
+import io.cattle.platform.iaas.api.auth.AbstractTokenUtil;
 import io.cattle.platform.iaas.api.auth.SecurityConstants;
 import io.cattle.platform.iaas.api.auth.dao.AuthTokenDao;
 import io.cattle.platform.iaas.api.auth.integration.external.ExternalServiceAuthProvider;
 import io.cattle.platform.iaas.api.auth.integration.interfaces.TokenCreator;
-import io.cattle.platform.iaas.api.auth.AbstractTokenUtil;
+import io.cattle.platform.object.ObjectManager;
 import io.github.ibuildthecloud.gdapi.context.ApiContext;
 import io.github.ibuildthecloud.gdapi.exception.ClientVisibleException;
 import io.github.ibuildthecloud.gdapi.factory.SchemaFactory;
@@ -16,14 +17,19 @@ import io.github.ibuildthecloud.gdapi.request.resource.impl.AbstractNoOpResource
 import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import javax.inject.Inject;
+import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.lang3.StringUtils;
 
 
 public class TokenResourceManager extends AbstractNoOpResourceManager {
+
+    @Inject
+    ObjectManager objectManager;
 
     @Inject
     AuthTokenDao authTokenDao;
@@ -88,14 +94,24 @@ public class TokenResourceManager extends AbstractNoOpResourceManager {
 
     @Override
     protected Object listInternal(SchemaFactory schemaFactory, String type, Map<Object, Object> criteria, ListOptions options) {
+        Token token = listToken();
+        return Collections.singletonList(token);
+    }
+
+    protected Token listToken() {
         Token token = new Token();
 
-        //get RedirectUrl if applicable
         if (SecurityConstants.AUTH_PROVIDER.get() == null || SecurityConstants.NO_PROVIDER.equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get())) {
             return token;
         }
 
         if (SecurityConstants.INTERNAL_AUTH_PROVIDERS.contains(SecurityConstants.AUTH_PROVIDER.get())) {
+            for (TokenCreator tokenCreator : tokenCreators) {
+                if (tokenCreator.isConfigured() && tokenCreator.providerType().equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get())) {
+                    token = tokenCreator.getCurrentToken();
+                    break;
+                }
+            }
             return token;
         } else {
             //get redirect Url from external service
@@ -115,4 +131,34 @@ public class TokenResourceManager extends AbstractNoOpResourceManager {
         this.tokenCreators = tokenCreators;
     }
 
+    @Override
+    protected Object deleteInternal(String type, String id, Object obj, ApiRequest request) {
+        if (!StringUtils.equals(AbstractTokenUtil.TOKEN, request.getType())) {
+            return null;
+        }
+        return deleteToken(obj, request);
+    }
+
+    protected Object deleteToken(Object obj, ApiRequest request) {
+        Token token = new Token();
+        String jwt = "";
+
+        token = listToken();
+        jwt = token.getJwt();
+
+        if(StringUtils.isBlank(jwt)) {
+            throw new ClientVisibleException(ResponseCodes.INTERNAL_SERVER_ERROR,
+                    "JWTNotProvided", "Request does not contain JWT cookie", null);
+        }
+
+        request.setResponseCode(ResponseCodes.NO_CONTENT);
+        HttpServletResponse response = request.getServletContext().getResponse();
+        String cookieString="token=;Path=/;Expires=Thu, 01 Jan 1970 00:00:00 GMT;";
+        response.addHeader("Set-Cookie", cookieString);
+        request.getServletContext().setResponse(response);
+        if(authTokenDao.deleteToken(jwt)) {
+            return obj;
+        }
+        return null;
+    }
 }
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureTokenCreator.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureTokenCreator.java
@@ -51,4 +51,9 @@ public class AzureTokenCreator extends AzureConfigurable implements TokenCreator
     public String getName() {
         return AzureConstants.TOKEN_CREATOR;
     }
+
+    @Override
+    public Token getCurrentToken() {
+        return azureTokenUtils.retrieveCurrentToken();
+    }
 }
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/external/ExternalServiceTokenUtil.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/external/ExternalServiceTokenUtil.java
@@ -1,19 +1,14 @@
 package io.cattle.platform.iaas.api.auth.integration.external;
 
-import io.cattle.platform.api.auth.Identity;
 import io.cattle.platform.core.model.Account;
 import io.cattle.platform.iaas.api.auth.AbstractTokenUtil;
-import io.cattle.platform.iaas.api.auth.identity.Token;
 import io.cattle.platform.object.util.DataAccessor;
-import io.cattle.platform.util.type.CollectionUtils;
 import io.github.ibuildthecloud.gdapi.context.ApiContext;
 import io.github.ibuildthecloud.gdapi.request.ApiRequest;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 
-import org.apache.commons.lang.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 
 public class ExternalServiceTokenUtil extends AbstractTokenUtil {
@@ -90,32 +85,5 @@ public class ExternalServiceTokenUtil extends AbstractTokenUtil {
     @Override
     public String getName() {
         return "";
-    }
-
-    public Identity jsonToIdentity(Map<String, Object> jsonData) {
-        String externalId = ObjectUtils.toString(jsonData.get("externalId"));
-        String externalIdType = ObjectUtils.toString(jsonData.get("externalIdType"));
-        String name = ObjectUtils.toString(jsonData.get("name"));
-        String profilePicture = ObjectUtils.toString(jsonData.get("profilePicture"));
-        String profileUrl = ObjectUtils.toString(jsonData.get("profileUrl"));
-        String login = ObjectUtils.toString(jsonData.get("login"));
-        return new Identity(externalIdType, externalId, name, profileUrl, profilePicture, login);
-    }
-
-    public Token getUserIdentityFromJWT() {
-        Token token = new Token();
-        Map<String, Object> jsonData = getJsonData();
-        if (jsonData == null) {
-            return null;
-        }
-        Object idObject = (Object)jsonData.get(USER_IDENTITY);
-        if (idObject != null) {
-            Map<String, Object> idMap = CollectionUtils.toMap(idObject);
-            Identity userIdentity = jsonToIdentity(idMap);
-            String userType = ObjectUtils.toString(jsonData.get(USER_TYPE), null);
-            token.setUserIdentity(userIdentity);
-            token.setUserType(userType);
-        }
-        return token;
     }
 }
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/ad/ADTokenCreator.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/ad/ADTokenCreator.java
@@ -70,4 +70,9 @@ public class ADTokenCreator extends ADConfigurable implements TokenCreator {
     public String getName() {
         return ADConstants.TOKEN_CREATOR;
     }
+
+    @Override
+    public Token getCurrentToken() {
+        return adUtils.retrieveCurrentToken();
+    }
 }
--- a/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/LocalAuthTokenCreator.java
+++ b/code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/LocalAuthTokenCreator.java
@@ -68,4 +68,9 @@ public class LocalAuthTokenCreator extends LocalAuthConfigurable implements Toke
     public String getName() {
         return LocalAuthConstants.TOKEN_CREATOR;
     }
+
+    @Override
+    public Token getCurrentToken() {
+        return localAuthTokenUtils.retrieveCurrentToken();
+    }
 }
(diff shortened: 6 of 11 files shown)


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/ad/ADTokenCreator.java
```diff
@@ -22,6 +22,8 @@
 public class ADTokenCreator extends ADConfigurable implements TokenCreator {
 
     @Inject
+    ADTokenUtils adTokenUtil;
+    @Inject
```
prachidamle: This is not needed, there is already an     @Inject ADTokenUtils adUtils;

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureTokenCreator.java
```diff
@@ -51,4 +51,24 @@ public void reset() {
...
+                token.setId(userToken.getId());
+            }
+        }
+        return token;
+    }
+
+    @Override
+    public String getJWTUtil() {
+        return azureTokenUtils.getJWT();
+    }
```
prachidamle: You don't have to implement entire getCurrentToken() in each TokenCreator. Each TokenCreator is having the exact same code. 
Instead add this logic to a method retrieveCurrentToken() to AbstractTokenUtil and then just invoke that method (azureTokenUtils.retrieveCurrentToken()) from each TokenCreator:: getCurrentToken()

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/external/ExternalServiceTokenUtil.java
```diff
@@ -108,7 +119,10 @@ public Token getUserIdentityFromJWT() {
         if (jsonData == null) {
             return null;
         }
-        Object idObject = (Object)jsonData.get(USER_IDENTITY);
+        AuthToken authToken= authTokenDao.getTokenByKey(getJWT());
```
prachidamle: Why do you need the methods jsonToIdentity() and getUserIdentityFromJWT() here? I believe the parent class AbstractTokenUtil already has these methods

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/Token.java
```diff
@@ -115,4 +116,13 @@ public String getRedirectUrl() {
     public void setRedirectUrl(String redirectUrl) {
         this.redirectUrl = redirectUrl;
     }
+
+    @Field(nullable = true)
```
prachidamle: I dont think this field can be nullable = true.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureTokenCreator.java
```diff
@@ -51,4 +51,24 @@ public void reset() {
     public String getName() {
         return AzureConstants.TOKEN_CREATOR;
     }
+
+    @Override
+    public Token getCurrentToken() {
```
prachidamle: Each TokenCreator is having exact same implementation of this method. Instead add this logic to AbstractTokenUtil::readCurrentToken() and invoke that from each tokenCreator:: getCurrentToken()

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureTokenCreator.java
```diff
@@ -51,4 +51,24 @@ public void reset() {
...
+                token.setUserIdentity(userToken.getUserIdentity());
+                token.setUserType(userToken.getUserType());
+                token.setId(userToken.getId());
+            }
+        }
+        return token;
+    }
+
+    @Override
+    public String getJWTUtil() {
```
prachidamle: Consider better naming for this method, say getTokenJWTValue()

mrajashree: Will change it

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/azure/AzureTokenCreator.java
mrajashree: Will do

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/external/ExternalServiceTokenUtil.java
mrajashree: I added them to the parent class from these classes. I can move them to the AbstractTokenUtil and use them

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/ad/ADTokenCreator.java
mrajashree: oh I missed that

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
```diff
@@ -396,4 +406,79 @@ public static boolean isUnrestrictedAccess(String accessMode){
...
+            Map<String, Object> idMap = CollectionUtils.toMap(idObject);
+            Identity userIdentity = jsonToIdentity(idMap);
+            String userType = ObjectUtils.toString(jsonData.get(USER_TYPE), null);
+            token.setUserIdentity(userIdentity);
+            token.setUserType(userType);
+        }
+        return token;
+    }
+
+    public String getRedirectUrl() {
```
prachidamle: getRedirectUrl() is calling the externalAuthService to get the redirect Url... this method is specific to external provider. Lets keep this with the ExternalProvider, since its does not apply to all providers.

mrajashree: Yes I had not included it initially and just realized that with the new changes I have left it int. I'll change it

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +138,50 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
...
+        String jwt = "";
+        Token token = null;
+        if (SecurityConstants.INTERNAL_AUTH_PROVIDERS.contains(SecurityConstants.AUTH_PROVIDER.get())) {
+            for (TokenCreator tokenCreator : tokenCreators) {
+                if (tokenCreator.isConfigured() && tokenCreator.providerType().equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get())) {
+                    if (!SecurityConstants.SECURITY.get()) {
+                        tokenCreator.reset();
+                    }
+                    jwt = tokenCreator.getTokenJWTValue();
+                    token = tokenCreator.getCurrentToken();
```
prachidamle: Lets set jwt key to the Token returned by tokenCreator.getCurrentToken();. This way you will not have to expose  and call the separate method tokenCreator.getTokenJWTValue()

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +138,50 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
...
+                    break;
+                }
+            }
+        } else {
+            if (externalAuthProvider.isConfigured()) {
+                jwt = tokenUtil.getJWT();
+                token = externalAuthProvider.readCurrentToken();
+            }
+        }
+
```
prachidamle: Seems like we can do some optimization between deleteToken and listInternal, since they both call the providers to read the token, this logic can be put in a utility method and called from both deleteToken and listInternal.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/local/LocalAuthTokenCreator.java
```diff
@@ -68,4 +68,14 @@ public void reset() {
...
+
+    @Override
+    public Token getCurrentToken() {
+        return localAuthTokenUtils.retrieveCurrentToken();
+    }
+
+    @Override
+    public String getTokenJWTValue() {
+        return localAuthTokenUtils.getJWT();
+    }
```
prachidamle: We may not need this if you start setting JWT to the token returned by getCurrentToken()

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
mrajashree: Yes, it was separate till now since I was also calling getJwt for all TokenCreators during delete in the for loop. And listInternal didn't need that. Now that we're setting JWT on token I'll also remove the getJwt for all tokenCreators and have a common method for getting current token for list and delete

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +138,30 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
         this.tokenCreators = tokenCreators;
     }
 
+    @Override
+    protected Object deleteInternal(String type, String id, Object obj, ApiRequest request) {
+        if (!StringUtils.equals(AbstractTokenUtil.TOKEN, request.getType())) {
+            return null;
+        }
+        return deleteToken(obj, request);
```
prachidamle: Did we remove the check to see if the "id" passed is correct and matches the id of the token in the cookie

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
```diff
@@ -396,4 +397,49 @@ public static boolean isUnrestrictedAccess(String accessMode){
...
+        return new Identity(externalIdType, externalId, name, profileUrl, profilePicture, login);
+    }
+
+    public Token getUserIdentityFromJWT() {
+        Token token = new Token();
+        Map<String, Object> jsonData = getJsonData();
+        if (jsonData == null) {
+            return null;
+        }
+        AuthToken authToken= authTokenDao.getTokenByKey(getJWT());
```
prachidamle: what if getJWT() returns null? Check how authTokenDao.getTokenByKey(null) behaves. If it throws NPE, then we should have a null check added before calling it

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
```diff
@@ -396,4 +397,49 @@ public static boolean isUnrestrictedAccess(String accessMode){
...
+        String tokenId = (String)ApiContext.getContext().getIdFormatter().formatId(objectManager.getType(Account.class), authToken.getId());
```
prachidamle: The formatId is being called by passing "type" = Account.class, this needs to be corrected to use the correct type

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -83,19 +90,35 @@ private Token createToken(ApiRequest request) {
         token.setIdentities(transFormedIdentities);
         token.setUserIdentity(identityManager.untransform(token.getUserIdentity(), true));
         token.setJwt(authTokenDao.createToken(token.getJwt(), token.getAuthProvider(), ((Policy) ApiContext.getContext().getPolicy()).getAccountId()).getKey());
+        AuthToken authToken= authTokenDao.getTokenByKey(token.getJwt());
+        String tokenId = (String)ApiContext.getContext().getIdFormatter().formatId(objectManager.getType(Account.class), authToken.getId());
```
prachidamle: Above comment applies here as well.

prachidamle:
LGTM

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -83,19 +89,35 @@ private Token createToken(ApiRequest request) {
...
-        //get RedirectUrl if applicable
         if (SecurityConstants.AUTH_PROVIDER.get() == null || SecurityConstants.NO_PROVIDER.equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get())) {
             return token;
         }
 
         if (SecurityConstants.INTERNAL_AUTH_PROVIDERS.contains(SecurityConstants.AUTH_PROVIDER.get())) {
+            for (TokenCreator tokenCreator : tokenCreators) {
+                if (tokenCreator.isConfigured() && tokenCreator.providerType().equalsIgnoreCase(SecurityConstants.AUTH_PROVIDER.get())) {
+                    if (!SecurityConstants.SECURITY.get()) {
+                        tokenCreator.reset();
```
cjellick: Why does listing tokens cause this reset?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/AbstractTokenUtil.java
```diff
@@ -396,4 +397,53 @@ public static boolean isUnrestrictedAccess(String accessMode){
...
+        if (findAndSetJWT()) {
+            Token userToken = getUserIdentityFromJWT();
+            if(userToken != null) {
+                token.setUserIdentity(userToken.getUserIdentity());
+                token.setUserType(userToken.getUserType());
+                token.setId(userToken.getId());
+                token.setJwt(getJWT());
+            }
+        }
+        log.info("retrieveCurrentToken returning {}", token);
```
cjellick: This seems in appropriate to log at info level. What do you think?

cjellick: Seems more like a debug statement

mrajashree: @cjellick Oh, yes that makes sense. I had just copied this method as is from ExternalServiceAuthProvider. I'll change it to debug level then

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
cjellick: Was just talking to @vincent99 about this check (which is in the below deleteToken method) and we want to change this. We actually don't want the UI to have to know and remember that ID. The cookie with the token should be enough info.

If you need to have an ID to work around cattle API quirks regarding DELETE on collections, then we can make the ID some constant string like "current" (ie /v2-beta/tokens/current)...this could also set us up in the future for implementing a /v2-beta/tokens/all  to invalidate all tokens for the current user.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +137,35 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
...
+        if (!StringUtils.equals(request.getId(), token.getId())) {
+            throw new ClientVisibleException(ResponseCodes.INTERNAL_SERVER_ERROR,
+                    "IdMismatch", "Id in request does not match token's id", null);
+        }
+
+        request.setResponseCode(ResponseCodes.NO_CONTENT);
+        if(authTokenDao.deleteToken(jwt)) {
+            return obj;
+        }
+        return null;
```
cjellick: @vincent99 is very much interested in having having us clear the token cookie server side so that he doesn't have to know about/do that.
He says you can do this by seting the expires to a date in the past and setting the value as a blank string. Can you do that?

resources/content/schema/token/token-auth.json
```diff
@@ -2,8 +2,8 @@
     "authorize" : {
         ".*" : "",
         "schema.*": "r",
-        "token": "cr",
-        "token.accountId": "r",
+        "token": "crd",
+        "token.id": "r",
```
cjellick: Hopefully, you can revert having to expose this ID.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
mrajashree: @cjellick yes, without the ID in URL, the DELETE request would get `405` that's why added the ID. So should I change it to some constant then?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
mrajashree: @cjellick I didn't get that. Currently the token will get removed from the db when DELETE is called. So do we not want to delete the token and only set it's expiry to Now()/something in past?

resources/content/schema/token/token-auth.json
mrajashree: @cjellick I added this id only because DELETE would call getById first, and in it's absence I would get a `MethodNotAllowed`. So if we switch to some constant string then yes I can remove this.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -83,19 +89,35 @@ private Token createToken(ApiRequest request) {
         token.setIdentities(transFormedIdentities);
         token.setUserIdentity(identityManager.untransform(token.getUserIdentity(), true));
         token.setJwt(authTokenDao.createToken(token.getJwt(), token.getAuthProvider(), ((Policy) ApiContext.getContext().getPolicy()).getAccountId()).getKey());
+        AuthToken authToken= authTokenDao.getTokenByKey(token.getJwt());
+        String tokenId = (String)ApiContext.getContext().getIdFormatter().formatId(objectManager.getType(Token.class), authToken.getId());
```
cjellick: Hoping we can get rid of this, but FYI it looks like you are double obfuscating the token:
```
curl 'http://192.168.42.5:8080/v2-beta/token' -H 'Origin: http://192.168.42.5:8080' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: en-US,en;q=0.8' -H 'x-api-action-links: actionLinks' -H 'content-type: application/json' -H 'x-api-no-challenge: true' -H 'accept: application/json'  --data-binary '{"code":"craig:craig","authProvider":"localauthconfig"}' | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   587  100   532  100    55  31080   3213 --:--:-- --:--:-- --:--:-- 31294
{
  "id": "1t1t10",  // <------------------------------------------------ Not right
  "type": "token",
  "links": {
    "self": "http://192.168.42.5:8080/v2-beta/tokens/1t1t10"
  },
  "baseType": "token",
  "actionLinks": {},
  "accountId": "1a1",
  "authProvider": "localAuthConfig",
  "code": null,
  "enabled": true,
  "jwt": "XkaFQJvDLpjntwrk54pbk4JYEvudHHqkmMZWMSy6",
  ... 
}
```

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
cjellick: I speakin of the the cookie in the request/response. You are doing the right thing for the token in the DB, but you should also clear the token cookie in the response

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
cjellick: yes

mrajashree: @cjellick it's working with `v2-beta/token/current`. Removed id, I'll update the PR

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
mrajashree: @cjellick @vincent99 I'm able to clear token value and set it's expiry to some date in past in the headers using `request.getServletContext().getResponse().setHeader()`. But for cookies there is addCookie, couldn't find how to set it/change it's value

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +133,63 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
...
+        List<Cookie> cookies = getAllCookies(httpRequest, "token");
+        if (cookies != null) {
+            if (cookies.size() == 2) {
+                cookies.get(0).setPath("/");
+                cookies.get(1).setPath("/v2-beta/token");
+            }
+
+            for (Cookie cookie : cookies) {
+                cookie.setValue("");
+                cookie.setMaxAge(0);
```
cjellick: I spoke with @vincent99 and unfortunately, max-age is not supported by internet explorer. So, you need to also set expires. The java cookie api doesnt have a setter for that, so you need to switch to adding a cookie via the set-cookie header. The value can just be a constant string that has a blank value, expires date in the past, and whatever other fields are required.

Also, I think you just need to do it once, you dont need to get all the cookies and loop through them.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +133,63 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
...
+                response.addCookie(cookie);
+            }
+        }
+
+        request.getServletContext().setResponse(response);
+        response.setHeader("token", "");
+        response.setHeader("Expires", "Thursday, 1 January 1970");
```
cjellick: I dont think you need this

mrajashree:
```
curl -i -X DELETE \
>   'http://localhost:8080/v2-beta/token/current?CSRF=0D8F1095B8' \
>   -H 'accept: application/json' \
>   -H 'cache-control: no-cache' \
>   -H 'cookie: token=Y1Ndom61YecdYhGYpjFTDFznvKwr6W48Kczjczpc' \
>   -b token=Y1Ndom61YecdYhGYpjFTDFznvKwr6W48Kczjczpc
```
HTTP/1.1 204 No Content
Content-Type: application/json; charset=utf-8
Date: Thu, 13 Apr 2017 22:36:01 GMT
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Server: Jetty(9.2.11.v20150529)
Set-Cookie: PL=rancher;Path=/
Set-Cookie: token=;Path=/;Expires=Thu, 01 Jan 1970 00:00:00 GMT;
Set-Cookie: token=;Path=/v2-beta/token;Expires=Thu, 01 Jan 1970 00:00:00 GMT;
X-Api-Account-Id: 1a4
X-Api-Client-Ip: ::1
X-Api-Schemas: http://localhost:8080/v2-beta/schemas
X-Api-User-Id: 1a4

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -115,4 +131,36 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
...
+
+        if(StringUtils.isBlank(jwt)) {
+            throw new ClientVisibleException(ResponseCodes.INTERNAL_SERVER_ERROR,
+                    "JWTNotProvided", "Request does not contain JWT cookie", null);
+        }
+
+        request.setResponseCode(ResponseCodes.NO_CONTENT);
+        HttpServletResponse response = request.getServletContext().getResponse();
+        String cookieString1="token=;Path=/;Expires=Thu, 01 Jan 1970 00:00:00 GMT;";
+        String cookieString2="token=;Path=/v2-beta/token;Expires=Thu, 01 Jan 1970 00:00:00 GMT;";
```
cjellick: Shouldnt need this. Should only need it on the root /

cjellick:
Also added a comment to the validation test PR

cjellick:
LGTM

PR #2589, opened 2017-03-30
should deactivate only when volume is already created

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/8357

cjellick:
LGTM

PR #2609, opened 2017-04-04
Limit concurrent sessions

mrajashree:
https://github.com/rancher/rancher/issues/8024

LDAP validation test PR: https://github.com/rancher/validation-tests/pull/298

TF `test_global_service_health ` seems unrelated


code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -89,7 +113,36 @@ private Token createToken(ApiRequest request) {
...
+        }
+
+        String accountId = ObjectUtils.toString(jsonData.get(AbstractTokenUtil.ACCOUNT_ID), null);
+        if (accountId == null) {
+            return null;
+        }
+        TokenUtil tokenUtil = tokenAuthLookup.tokenUtils();
+        Account account = authDao.getAccountByExternalId(accountId, tokenUtil.userType());
+        if (account != null) {
+            authenticatedAsAccountId = account.getId();
```
prachidamle: Why can't you use the accountId set in the token that is returned by the tokenCreator?

AbstractTokenUtil :: tokenCreator() creates or gets the account trying to login and sets the accountId to the token returned. Is this not the "authenticatedAsAccountId?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -89,7 +113,36 @@ private Token createToken(ApiRequest request) {
...
+        }
+        long tokenAccountId = (accountDao.findByUuid(AbstractTokenUtil.TOKEN)).getId();
+
```
prachidamle: I dont think you should get the accountId of kind "token" from accountDao this way.

Use ((Policy) ApiContext.getContext().getPolicy()).getAccountId() - that is what the code used while creating the token. This should resolve to "token" accountId.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthTokenDaoImpl.java
```diff
@@ -77,4 +78,16 @@ public AuthToken getTokenByAccountId(long accountId) {
     public boolean deleteToken(String key) {
         return create().executeDelete(getTokenByKey(key)) == 1;
     }
+
+    @Override
+    public AuthToken getTokenByAuthenticatedAsAccountId(long authenticatedAsAccountId, long tokenAccountId) {
```
prachidamle: Shouldn't there be more than one Tokens returned here?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -138,6 +191,14 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
         this.tokenCreators = tokenCreators;
     }
 
+    public void deletePreviousToken(long authenticatedAsAccountId, long tokenAccountId) {
+        AuthToken authToken = authTokenDao.getTokenByAuthenticatedAsAccountId(authenticatedAsAccountId, tokenAccountId);
```
prachidamle: What if there were multiple concurrent sessions -- multiple valid tokens?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -41,7 +46,20 @@
...
+    TokenAuthLookup tokenAuthLookup;
+
+    @Inject
+    AuthDao authDao;
+
+    @Inject
+    AccountDao accountDao;
+
     private List<TokenCreator> tokenCreators;
+    private static final DynamicBooleanProperty LIMIT_CONCURRENT_SESSIONS = ArchaiusUtil.getBoolean("api.auth.limit.concurrent.sessions");
```
cjellick: need to think about this name some more. restrict might be better than limit. cc @vincent99

prachidamle:
LGTM

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
prachidamle: yeah, limit kind of sounds like you can set the "number" of concurrent sessions allowed to say "2" or "3" etc. Restrict or Disable might be better.

cjellick:
Regarding @prachidamle's upgrade statement. Offline discussion determined that for users who wanted to turn this feature on, they could truncate the auth_token table. We would need to doc that, possibly in the release notes and in emails to specfic users asking for it.

resources/content/db/core-122.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <changeSet author="rancher (generated)" id="dump1">
+        <addColumn tableName="auth_token">
+            <column name="authenticated_as_account_id" type="BIGINT"/>
```
cjellick: Does this point at the account table? If so, shouldn't it be a foreign key?

mrajashree: @cjellick Yes it does, it has the same value as the `id` field of account table. So I'll change this to foreign key then?

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/identity/TokenResourceManager.java
```diff
@@ -131,6 +162,16 @@ public void setTokenCreators(List<TokenCreator> tokenCreators) {
         this.tokenCreators = tokenCreators;
     }
 
+    public void deletePreviousToken(long authenticatedAsAccountId, long tokenAccountId) {
+        List<AuthToken> authTokenList = authTokenDao.getTokenByAuthenticatedAsAccountId(authenticatedAsAccountId, tokenAccountId);
```
cjellick: The getting and deleting of tokens should be done in a single DB transaction. Could you move all this logic down into the DAO? In fact, in the function, instead of doing 1 select query and N delete queries, you should be able to do this in a single delete query with a where clause

mrajashree: @cjellick yes I can move this all to the DAO and do the delete as you said with a single query.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/dao/impl/AuthTokenDaoImpl.java
```diff
@@ -77,4 +80,16 @@ public AuthToken getTokenByAccountId(long accountId) {
...
+    @Override
+    public List<AuthToken> getTokenByAuthenticatedAsAccountId(long authenticatedAsAccountId, long tokenAccountId) {
+        return create()
+                .selectFrom(AUTH_TOKEN)
+                .where(AUTH_TOKEN.AUTHENTICATED_AS_ACCOUNT_ID.eq(authenticatedAsAccountId))
+                        .and(AUTH_TOKEN.ACCOUNT_ID.eq(tokenAccountId))
+                        .and(AUTH_TOKEN.VERSION.eq(SecurityConstants.TOKEN_VERSION))
+                        .and(AUTH_TOKEN.PROVIDER.eq(SecurityConstants.AUTH_PROVIDER.get()))
+                        .and(AUTH_TOKEN.EXPIRES.greaterThan(new Date()))
+                .orderBy(AUTH_TOKEN.EXPIRES.desc()).fetch().into(AuthToken.class);
```
cjellick: What is the point of the 
```
.and(AUTH_TOKEN.VERSION.eq(SecurityConstants.TOKEN_VERSION))
.and(AUTH_TOKEN.PROVIDER.eq(SecurityConstants.AUTH_PROVIDER.get()))
``` 
conditions? Seems like the only thing we should care about is the two ids.
Checking the expires seems fine to me, though also not entirely necessary.

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/OpenLDAP/OpenLDAPIdentityProvider.java
```diff
@@ -70,7 +70,7 @@
                 }
                 Set<Identity> identities = getIdentities(dn);
                 Token token = getTokenUtils().createToken(identities, null);
-                authToken = authTokenDao.createToken(token.getJwt(), getConstantsConfig().getConfig(), account.getId());
+                authToken = authTokenDao.createToken(token.getJwt(), getConstantsConfig().getConfig(), account.getId(), account.getId());
             }
```
prachidamle: At this point above, is the account_id column in the auth_token db set to the "authenticatedAsAccountId" value instead of the account of kind "token"?

mrajashree: @prachidamle  yes, it's the `authenticatedAsAccountId`

prachidamle: We should just make sure that this code is NOT hit while using a token cookie (browser session) and is only used while calling APIs via keys

mrajashree: @prachidamle Confirmed that this code is hit only when creating a token by using api keys, not through browser session

cjellick:
LGTM

prachidamle:
LGTM :)

PR #2617, opened 2017-04-05
Don't call external sched during deploy planning

cjellick:
The global service deployment planner was calling the external
scheduler to check if any hosts should be excluded from its list due to
scheduling rules there. This was specifically added to allow the
io.rancher.scheduler.require_any host label to work with global
services. But we cannot make this call here because the calling code
occupies a thread in the core executore service, which must be fast or
it will back up the system.


deniseschannon:
https://github.com/rancher/rancher/issues/8326

PR #2619, opened 2017-04-07
Reconcile healthchecks on host.activate

cjellick:
https://github.com/rancher/rancher/issues/8416

This fixes a bug that occurs if you take you environment down to zero
hosts by removin the hosts from your cloud provider and they go into a
disconnected state in Rancher. In that scenario, healthchecked
instances on the disconnected host would never get healthchecks
scheduled on the newly added hosts and thus would never get moved off
the disconnected host.

The process that we were previously keying off of
(networkserviceproviderinstancemap.create) is no longer executed. It
was executed to lanuch the network-agent container on the host, but
we've completely gone away from that model. We were previously using
that process because we needed the network-agent to be running in order
for that host to process healthchecks, but that is no longer necessary
because healthcheks are deployed in their own containers. The logic
later in this process that schedules the new healthchecks ensures that
the host will have a healthcheck container deployed to it, if it
doesn't already.

### DIFF
--- a/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
+++ b/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
@@ -480,13 +480,11 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
                 .leftOuterJoin(AGENT)
                     .on(AGENT.ID.eq(HOST.AGENT_ID))
                 .where(
-                        AGENT.ID.isNull()
-                                .or(AGENT.STATE.in(CommonStatesConstants.ACTIVE,
-                                        AgentConstants.STATE_FINISHING_RECONNECT, AgentConstants.STATE_RECONNECTED))
+                AGENT.ID.isNull()
+                .or(AGENT.STATE.in(CommonStatesConstants.ACTIVE, AgentConstants.STATE_FINISHING_RECONNECT, AgentConstants.STATE_RECONNECTED))
                 .and(HOST.REMOVED.isNull())
                 .and(HOST.ACCOUNT_ID.eq(accountId))
-                                .and(HOST.STATE.in(CommonStatesConstants.ACTIVATING, CommonStatesConstants.ACTIVE,
-                                        CommonStatesConstants.UPDATING_ACTIVE)))
+                .and(HOST.STATE.in(CommonStatesConstants.ACTIVATING, CommonStatesConstants.ACTIVE, CommonStatesConstants.UPDATING_ACTIVE)))
                 .fetchInto(Host.class);
     }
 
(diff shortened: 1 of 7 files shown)


code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/impl/HostMapHealthcheckInstancesLookup.java
```diff
@@ -25,9 +25,6 @@
         HealthcheckInstanceHostMap hostMap = (HealthcheckInstanceHostMap) obj;
 
         List<Instance> instances = new ArrayList<>();
-        if (!(obj instanceof HealthcheckInstanceHostMap)) {
```
cjellick: Expand the above code and you'll see this check was redundant.

cjellick:
Potential hosts for the HCIHM are retrieved here:
https://github.com/rancher/cattle/blob/2e53df9ea42f7aa20b73c66097a9b4286844bf07/code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/impl/HealthcheckServiceImpl.java#L261
...
https://github.com/rancher/cattle/blob/c28f841054eac49ae89562d4be4726a9024ce9cf/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java#L451-L466

Hosts that are active-ish with active (or reconnecting)

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -455,13 +455,11 @@ public void next(Record3<String, String, String> record) {
                 .leftOuterJoin(AGENT)
                     .on(AGENT.ID.eq(HOST.AGENT_ID))
                 .where(
```
cjellick: Just changed the formatting below

code/implementation/simulator/agent-connection/src/main/java/io/cattle/platform/agent/connection/simulator/AgentConnectionSimulator.java
```diff
@@ -32,6 +32,10 @@ public AgentConnectionSimulator(ObjectManager objectManager, Agent agent, List<A
 
     @Override
     public Event execute(Event event) {
```
cjellick: This causes the simulated agent to behave as if it were disconnected. I trigger it in SimulatorStartStopProcessor below

cjellick:
@ibuildthecloud you initially said you were not 100% convinced with this fix, but then added a test and and verified the functionality manually with hosts started from the UI (since the flow is different bc the host exists before the agent does) and you said it looked good.

Feel free to re-review, just giving you some context

cjellick:
I had a bug in the test, which I fixed

PR #2621, opened 2017-04-07
set initializing/reinitializing timeouts for LB

deniseschannon:


deniseschannon:
https://github.com/rancher/rancher/issues/8356

PR #2644, opened 2017-04-25
merge container affinity labels for primary and sickkicks

StrongMonkey:
When adding locks for container affinity rules, we need to merge the labels for both primary containers and sidekicks. What happens now is that the sidekicks are often scheduled first and they don't have a container-affinity label and ending up landing on the same hosts. The primary container do have the labels but they are scheduled with the sidekicks then will violate the affinity_rules.


StrongMonkey:
https://github.com/rancher/rancher/issues/8070

cjellick:
LGTM

PR #2645, opened 2017-04-25
WIP: Adding pre listener for container stop

prachidamle:
This listener should find all LB services that this container's service is a target of, and send an event to all those LB service instances. Upon reply from the LB services, container stop can proceed

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
@@ -1,0 +1,173 @@
+package io.cattle.platform.servicediscovery.process;
+
+import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.*;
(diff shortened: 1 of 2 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
```diff
@@ -0,0 +1,149 @@
...
+        if (service == null) {
+            return null;
+        }
+        List<Service> activeLbServices = lookupAllActiveLBServices(service);
+        List<? extends Instance> lbInstances = getLBServiceInstances(activeLbServices);
+        //send drain event to each lbInstance and wait for reply upto timeout.
+
+        for (Instance lbInstance : lbInstances) {
+            for (InstanceHostMap ihm : mapDao.findNonPurged(InstanceHostMap.class, Instance.class, lbInstance.getId())) {
+                Object agentResource = getObjectByRelationship("host", ihm);
```
cjellick: Dont do this line. Actually, you dont need this inner ihm loop at all. I think you just want to pass lbInstance to lookupAgent

PR #2675, opened 2017-05-03
Fix volumeTemplate stackID problems

cjellick:
- When validating volumeTemplate name uniqueness, scope query to account
- When looking for volumeTemplates during DU creation, if we can't find
  by name and stackId, do a second query with stackId null and
  external=true


cjellick:
https://github.com/rancher/rancher/issues/8642
https://github.com/rancher/rancher/issues/8497 (though @sangeethah may create a new bug)

cjellick:
I'm still not sure if having a null stackID makes sense. Here is yet another bug caused by the fact that up until this point, we assumed volumeTemplates were not null: https://github.com/rancher/rancher/issues/8629

cjellick:
closing as the proper fix is to always set stackId

PR #2683, opened 2017-05-04
Handle duplicate instance host maps gracefully

cjellick:
When stopping or starting an instance and calling the external
scheduler, we have to get the host UUID to send to the scheduler. In
the very odd scenario that more than one instance host map exists,
first determine if they are all pointing at the same host and if so,
return that host. If they do not all point at the same host, just log
an error and return null so that stop/start isn't stuck forever.


cjellick:
I tested the rest of the life cycle with duplicate ihm entries (both with same host id and different host ids) and once you get past this error, the rest of the container's lifecycle is handled gracefully

cjellick:
There is no natural way to test this as it requires manipulating the database. But, you can force the situation by creating another instance_host_map entry in the database thusly:
```
insert into instance_host_map (kind, uuid, state, created, data, instance_id, host_id) VALUES ('instanceHostMap', '<some string that wont conflict with other rows>', 'active', '2017-05-04 16:35:38', '{}', <instance id>, <host id>);
```

PR #2689, opened 2017-05-12
don't override pre-existing affinity rules

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/8762

cjellick:
Needs a test. I think you could do this with a cattle integration test. You don't need a real ebs volume, just have to pass in the ec2_az driver opt.

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreStart.java
```diff
@@ -66,7 +66,12 @@ public void doWithLockNoResult() {
             Map<String, Object> driver_opts = DataAccessor.fieldMap(v, VolumeConstants.FIELD_VOLUME_DRIVER_OPTS);
             if (driver_opts.containsKey(VolumeConstants.EC2_AZ)) {
                 String zone = driver_opts.get(VolumeConstants.EC2_AZ).toString();
-                labels.put(HostAffinityConstraint.LABEL_HEADER_AFFINITY_HOST_LABEL, String.format("%s=%s", VolumeConstants.HOST_ZONE_LABEL_KEY, zone));
+                String label = String.format("%s=%s", VolumeConstants.HOST_ZONE_LABEL_KEY, zone);
+                if (labels.containsKey(HostAffinityConstraint.LABEL_HEADER_AFFINITY_HOST_LABEL)) {
```
cjellick: To protect against the (unlikely but still possible) scenario of `labels.get(HostAffinityConstraint.LABEL_HEADER_AFFINITY_HOST_LABEL)` returning NULL and thus getting an NPE on the next line, could you change this if statement to !StringUtils.isEmpty(labels.get(HostAffinityConstraint.LABEL_HEADER_AFFINITY_HOST_LABEL)))`?

cjellick: I don't see any need for two separate checks. Just one:
```
if (!StringUtils.isEmpty(labels.containsKey(HostAffinityConstraint.LABEL_HEADER_AFFINITY_HOST_LABEL))) {
    label = labels.get(HostAffinityConstraint.LABEL_HEADER_AFFINITY_HOST_LABEL).toString() + "," + label;
}
```
Am I missing something?
Also, `!isEmpty` is better than `!= null` because it will prevent the weird scenario where that label is set with a empty string and you get a value like `",,io.rancher.host.zone=east"`. I realize that this seems overkill but its good defensive programming with no extra cost in complexity.

StrongMonkey: we still need to check if the object is null, labels is a map<String,Object>. I can do the check in one line.

cjellick: Ah, that's the part that I was missing: that the value is an object.

StrongMonkey: @cjellick done!

cjellick:
LGTM assuming tests pass

PR #2690, opened 2017-05-15
Add User field to identity

mrajashree:
https://github.com/rancher/rancher/issues/8744
`User=true` will be set by all providers separately for all user accounts
`Me=true` will be set by cattle's listInternal, for all identities of the authenticatedAsAccount

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/auth/Identity.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/auth/Identity.java
@@ -19,6 +19,7 @@ public class Identity {
     private String login = null;
     private String role = null;
     private String projectId = null;
+    private boolean user = false;
 
     @Field(required = false, nullable = true)
     public String getName() {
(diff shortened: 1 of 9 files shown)


mrajashree:
This for loop https://github.com/rancher/cattle/pull/2690/commits/6401fad6dbaec8280d682cf9071b729e80584043#diff-d09c650621963e26b109b1edbdfa0658R71
sets `Me=true` on all identities of the current policy. The policy returns identities only for authenticatedAsAccount. So I think right now any more checks won't be needed and we can set `me=true` on all identities of the policy. 
@cjellick @prachidamle Let me know if this seems correct

code/framework/api/src/main/java/io/cattle/platform/api/auth/Identity.java
```diff
@@ -19,6 +19,8 @@
     private String login = null;
     private String role = null;
     private String projectId = null;
+    private Boolean me = false;
```
ibuildthecloud: Can you use `boolean` and not `Boolean`?  Is there a specific reason you used the boxed type?

resources/content/schema/user/user-auth.json
```diff
@@ -300,6 +300,8 @@
         "identity.projectId": "cr",
         "identity.role": "cr",
         "identity.all": "cr",
+        "identity.me": "cr",
+        "identity.user": "cr",
```
prachidamle: I think these can be only "r"

prachidamle:
We should also change RancherIdentityProvider to add the "user=true" flag on the identities returned

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/ldap/LDAPIdentityProvider.java
```diff
@@ -194,6 +194,7 @@ protected Identity attributesToIdentity(LdapName dn){
             String accountName;
             String externalId = dn.toString();
             String login;
+            Boolean user = false;
```
cjellick: should be lowercase boolean

mrajashree: @cjellick ah yes, hadn't checked the PR again after updating it and this didn't show any errors in eclipse so didn't realize

code/framework/api/src/main/java/io/cattle/platform/api/auth/Identity.java
mrajashree: @ibuildthecloud changed it to `boolean`.

cjellick:
Build failures are unrelated. Saw them in other merge builds

cjellick:
LGTM

PR #2691, opened 2017-05-16
switch default catalog branch to master

cjellick:


deniseschannon:
@cjellick We need to think of a strategy on what catalog branch we should use in our master container.

PR #2693, opened 2017-05-16
Dont process stop events if container not started

cjellick:
If container is still in starting, then we need to let that finish.
This specifically addresses a bug wherein container create takes a very
long time due to copying data from the image into a volume, but it
makes sense more generally as well.


cjellick:
https://github.com/rancher/rancher/issues/8142

cjellick:
Got some feedback from @ibuildthecloud. Need to tweak it.

ibuildthecloud:
LGTM

PR #2702, opened 2017-05-18
SAML specific ExternalAuthProvider changes 

mrajashree:
https://github.com/rancher/rancher/issues/8744
rancher-auth-service sets the setting `api.auth.external.provider.no.identity.lookup` to true, since unlike github/ldap, saml doesn't have identity lookup. Because of this setting being true, rancher-auth-service's `v1-auth/identities` is never called for saml while listing identities. It then reads allowed identities from db and gets only the `externalId` and `externalIdType`. This PR sets `user` flag on identities whose externalIdType matches the `USER_TYPE` setting.
`USER_TYPE` is obtained from the provider


ibuildthecloud:
YOU ARE YELLING IN YOUR COMMIT MESSAGE

mrajashree:
@ibuildthecloud oops lol

cjellick:
LGTM

PR #2705, opened 2017-05-19
WIP Attempt to find projectId in URL path

cjellick:
This will allow us to put the project ID in proxied microservices (like
auth and catalog service). For example, the URLs will be able to look
like: http://localhost:8080/v1-catalog/projects/1a5/something

@ibuildthecloud @vincent99 curious what you think about this.


vincent99:
+10 to the idea

cjellick:
(Well, it _is_ your idea)

vincent99:
The code suitably java-ey; it would be nice if each microservice could list projects and their relevant links so maybe `/v1-foo/projects/1a5` should be handled too?

PR #2710, opened 2017-05-24
Dont find an agent if the resource is removed

cjellick:
When getting the "agentResource" that is later used to find an agent,
don't return the resource if it has been removed.

This will cause AgentBasedProcessHandler logic to short circuit and
complete successfully.

The concrete case for this if when a storage pool has been removed but
there are still volumeStoragePoolMaps associated with it that need
removed.


cjellick:
On second thought, @ibuildthecloud put in this PR two months ago that addresses the same bug but in a different manner
https://github.com/rancher/cattle/pull/2587

So, I'm going to close this PR as superfluous.

PR #2713, opened 2017-05-24
add resource requests to allocation error message

StrongMonkey:


StrongMonkey:
for https://github.com/rancher/rancher/issues/8636

cjellick:
@StrongMonkey what does the message look like when it fails for a combination of reasons. Specifically, what is the error message in this scenario?
1. have two hosts, one with label foo=bar
2. deploy a container that uses port 80 on that host
3. create a service of scale=1 that needs port 80 and has a label constraint that it must land on a host with a foo=bar label?

StrongMonkey:
The msg is described in the issue linked. @cjellick

cjellick:
Let's make it more user friendly:
Change:
```
Allocation failed: needs host with label foo: bar, ResourceRequests: [ports: [8080:8080/tcp], ResourceType: instanceReservation, Amount: 1]
```
to:
```
Allocation failed: needs host with label foo: bar and resources constraints: [ports: 8080:8080/tcp, 1 instanceReservation]
```

cjellick:
Changed it again, want `and resource constraints[ 80/tcp, 90/tcp ports, 1 instanceReservation]`

cjellick:
LGTM

PR #2715, opened 2017-05-25
Fix host template + host validation

cjellick:
When creating a host, if only hostTemplateId is set and no *Config
field is set, don't fail validation. Also, since we set the driver
field based off of the selected config, update that logic to account
for hostTemplate logic.


ibuildthecloud:
LGTM

PR #2717, opened 2017-05-25
Filter password settings

mrajashree:
https://github.com/rancher/rancher/issues/8625
For saml, no secret value is stored, so there's no corresponding setting for the provider

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/settings/SettingsFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/settings/SettingsFilter.java
@@ -4,9 +4,11 @@ import io.cattle.platform.api.auth.Policy;
 import io.cattle.platform.api.utils.ApiUtils;
 import io.cattle.platform.core.addon.ActiveSetting;
 import io.cattle.platform.core.model.Setting;
+import io.github.ibuildthecloud.gdapi.context.ApiContext;
 import io.github.ibuildthecloud.gdapi.request.ApiRequest;
 import io.github.ibuildthecloud.gdapi.util.RequestUtils;
 
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -19,7 +21,8 @@ public class SettingsFilter implements Predicate {
     Set<String> publicSettings;
     boolean all;
     boolean canListAll;
-
+    private static final Set<String> SETTINGS_HIDDEN = new HashSet<>(Arrays.asList(new String[]{"api.auth.ldap.service.account.password",
+            "api.auth.ldap.openldap.service.account.password", "api.auth.azure.admin.password", "api.auth.github.client.secret"}));
     public SettingsFilter(List<String> publicSettings, ApiRequest apiRequest) {
         String value = null;
         Map<String, Object> params = apiRequest == null ? null : apiRequest.getRequestParams();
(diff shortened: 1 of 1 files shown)


code/implementation/settings-api/src/main/java/io/cattle/platform/api/settings/manager/SettingsFilter.java
```diff
@@ -14,11 +15,14 @@
...
 import org.apache.commons.collections.Predicate;
 
+import com.netflix.config.DynamicStringListProperty;
+
 public class SettingsFilter implements Predicate {
 
     Set<String> publicSettings;
     boolean all;
     boolean canListAll;
+    public static final DynamicStringListProperty SETTINGS_HIDDEN = ArchaiusUtil.getList("settings.hidden");
```
ibuildthecloud: We have no ability to really create a hard coded setting so a setting to hide settings just creates a security problem.  I would opt for a change in design in that you can create a setting that has a hidden=true flag on it.  Then we can properly encrypt the values.  If that is too much up front effort and you want to do a two phased approach, then start here with a hardcoded list of properties to hide.

A third option is to have a regexp setting of what to hide, like `.*password.*|.*secret.*`.  If it was one simple static setting then we could put it in the `Dockerfile` of `rancher/server` and it would hide the secret.

@cjellick

code/implementation/settings-api/src/main/java/io/cattle/platform/api/settings/manager/SettingsFilter.java
```diff
@@ -36,10 +40,18 @@ public SettingsFilter(List<String> publicSettings, ApiRequest apiRequest) {
     @Override
     public boolean evaluate(Object object) {
         String name = null;
+        List<String> settingsHidden = SETTINGS_HIDDEN.get();
```
ibuildthecloud: Please use a `O(1)` data structure for lookup.  Settings gets called a lot more than one would think.

code/implementation/settings-api/src/main/java/io/cattle/platform/api/settings/manager/SettingsFilter.java
```diff
@@ -36,10 +40,18 @@ public SettingsFilter(List<String> publicSettings, ApiRequest apiRequest) {
...
         if (object instanceof ActiveSetting) {
             name = ((ActiveSetting) object).getName();
+            if (settingsHidden.contains(name)) {
+                ((ActiveSetting) object).setActiveValue(null);
```
ibuildthecloud: Please don't repeat this logic some much.

cjellick:
I want to do the two phase approach:
Phase 1: A hard coded list of settings to not return the value of in the API
Phase 2: Introduce a setting type where hidden=true.

Also implied from the above, I'd like to go with the hard coded list of settings.

code/implementation/settings-api/src/main/java/io/cattle/platform/api/settings/manager/SettingsFilter.java
mrajashree: @ibuildthecloud Using `DynamicStringMapProperty` now for `settings.hidden` instead of `DynamicStringListProperty`

mrajashree:
TF `test_container_ha_default` seems unrelated

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/settings/SettingsFilter.java
```diff
@@ -4,21 +4,26 @@
...
 import org.apache.commons.collections.Predicate;
 
+import com.netflix.config.DynamicStringMapProperty;
+
 public class SettingsFilter implements Predicate {
 
     Set<String> publicSettings;
     boolean all;
     boolean canListAll;
+    public static final DynamicStringMapProperty SETTINGS_HIDDEN = new DynamicStringMapProperty("settings.hidden", Collections.emptyMap());
```
cjellick: We don't want a setting as then the setting could just be overidden to unhide. Just a hard coded list in the code

mrajashree: @cjellick Right, I totally misunderstood previous comments, updating now

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/settings/SettingsFilter.java
```diff
@@ -19,6 +21,9 @@
     Set<String> publicSettings;
     boolean all;
     boolean canListAll;
+    List<String> secretSettings = Arrays.asList("api.auth.ldap.service.account.password", "api.auth.ldap.openldap.service.account.password",
+            "api.auth.azure.admin.password", "api.auth.github.client.secret");
+    HashSet<String> settingsHidden = new HashSet<>(secretSettings);
```
cjellick: Make this a constant (`private static final Set`)
Dont ever declare a variable as the underlying implmentation like HashSet. Always use the interface Set. This is true of all.
After declaring as a constant, you can use a static initializer to add the valus:
```
static {
  secretSettings.add("api...");
}
```
You should not have to create the List at all...just the set

PR #2725, opened 2017-05-27
Check that hostTemplate contains <driver>Config for a single driver

imikushin:
Also check that driver and name are provided and name is unique.


code/iaas/api-logic/src/test/java/io/cattle/platform/iaas/api/host/HostTemplateValidationFilterTest.java
```diff
@@ -0,0 +1,165 @@
...
+import org.mockito.Mockito;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static io.cattle.platform.iaas.api.host.HostTemplateValidationFilter.*;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+public class HostTemplateValidationFilterTest {
```
cjellick: Generally, we write tests like this at the integration level, in python in the tess/inegration directory

cjellick:
LGTM

PR #2729, opened 2017-05-30
Add tweaks to recognize special bind mounts

StrongMonkey:
We are rewriting the volume logic for rancher managed storage driver. The rancher managed volume are passed as a bind mount to the container. So we need to parse the mountpoint to see if its a rancher mountpoint.
If so, we make this mountpoint tied to the specific volume so that API/UI would know the active mounts for the volume and show the correct state.


code/implementation/docker/common/src/main/java/io/cattle/platform/docker/transform/DockerTransformerImpl.java
```diff
@@ -108,6 +109,10 @@
             String hostPath = (String)mountObj.get(SRC);
             String name = (String)mountObj.get(NAME);
             String externalId = null;
+            if (StringUtils.startsWith(hostPath, "/var/lib/rancher/volumes/rancher-")) {
```
ibuildthecloud: You should feel bad :)

PR #2735, opened 2017-05-31
Disconnect websockets

cjellick:
When destroying old tokens because concurrent tokens are not allowed,
also emit and event that will cause any existing websocket connections
to /subscribe to terminate.

This commit only narrowly addresses the above use case, but in another
PR, we can expand this functionality to handle user logouts and agent
removals/disconnects.


ibuildthecloud:
LGTM

cjellick:
had java formatting error bc a line was too long. had to repush

cjellick:
Failure unrelated

PR #2736, opened 2017-05-31
Fix to add stack_name to container metadata

prachidamle:
This adds the stack_name to the metadata for a container. This is needed especially for standalone containers.

For service containers, rancher-metadata adds this field by looking at the enclosing service, but for a standalone container it cannot find the stack_name since there is no service enclosing that container.

Hence cattle should set the stack_name. 

--------------------------

Also tried setting stack_uuid here, but rancher-metadata panics at this line: https://github.com/rancher/rancher-metadata/blob/cd59990d7569efdb76ad3d0eca82ed70510458be/generator.go#L78
(It tries to look for container.service_uuid which will be nil for a standalone container, so the  type conversion to string fails here. )
So this code in rancher-metadata should be fixed to add checks around service_uuid, if we also need stack_uuid from cattle.

@ibuildthecloud pls review.


prachidamle:
unrelated TF's

prachidamle:
@ibuildthecloud @cjellick does the fix look ok?

PR #2751, opened 2017-06-09
Volume fixes

cjellick:
Fixes multiple causes of stack volumes not getting delete properly on the hosts

Necessary changes:
- Set driver and device number on stack vols (needed for proper volume deletion in agent)
- Dont fallback to deactivate+remove for stack vols (needed so that volumes are not prematurely deleted)

Downside:
- If a stack is deleted and a volume it owned was mounted to a non-stack container, stack deletion will spin forever. But, our old behavior of just deleting the volume anyway (and then not actually deleting it on the host because it was in use and thus ultimately orphaning it on the host wasn't any better)

### DIFF
--- a/code/framework/resource-monitor/src/main/java/io/cattle/platform/object/monitor/impl/ResourceMonitorImpl.java
+++ b/code/framework/resource-monitor/src/main/java/io/cattle/platform/object/monitor/impl/ResourceMonitorImpl.java
@@ -133,7 +133,12 @@ public class ResourceMonitorImpl implements ResourceMonitor, AnnotatedEventListe
 
     @Override
     public <T> T waitForState(T obj, final String desiredState) {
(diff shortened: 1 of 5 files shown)


cjellick:
For https://github.com/rancher/rancher/issues/9038

cjellick:
@ibuildthecloud this does two thing that we've talked about. But i'm worried about that volume.remove change on stacke.remove and wanted to get your thoughts:

This will 1) cause the removal of any stack that has volumes to take X(15?) more seconds to get removed. 2) cause a stack to spin in stack.remove forever if a user did something weird to use a stack volume in another container.

Do you think either of those (or anything else you can think of) are deal breakers?

Note

cjellick:
Got an offline LGTM from @ibuildthecloud

PR #2761, opened 2017-06-13
reject bad image names

StrongMonkey:


StrongMonkey:
This pr rejects the image name ends with a non-alphanumeric character. For example
```
mydomain/myimage:  invalid
mydomain/myimage; invalid
```
This will solve the problem where java regexp parse stuff too long. While I found this is still a problem:
`mydomain/myimagelonglonglonglonglonglonglonglonglong::tag`
Java regexp still takes a long time to parse something like this. I am not going to address this as the parsing will be more complicated when thing goes further.
However, the alternative solution is https://github.com/rancher/cattle/pull/2762, where we drop all the regexp as all we want to do is just to get the registry server name. And [that doesn't need regexp](https://github.com/moby/moby/blob/24698a07cba89a08ac6fcc2c6e4097491b922883/vendor/github.com/docker/distribution/reference/normalize.go#L62).

ibuildthecloud:
I reject your reject bad image names.

PR #2762, opened 2017-06-13
drop the world

StrongMonkey:


StrongMonkey:
This PR will drop all the regexp we ever used, cause all we want to do is just the server name. The side effect of this is that we will lose the ability to validate image names, but we also avoid having complicated regexp problems in parsing these names.

ibuildthecloud:
```java
 if (uuid.startsWith(KIND_PREFIX)) {
            uuid = uuid.substring(KIND_PREFIX.length());
        } else if (uuid.startsWith(SIM_PREFIX)) {
            uuid = uuid.substring(SIM_PREFIX.length());
        }
```

StrongMonkey:
https://github.com/rancher/rancher/issues/7641

PR #2769, opened 2017-06-15
Dont let restricted user access infra containers

cjellick:


cjellick:
Addresses: https://github.com/rancher/rancher/issues/9064

PR #2770, opened 2017-06-15
rancher/agent:v1.2.3-rc1

cjellick:


cjellick:
Got an LGTM from @ibuildthecloud

PR #2771, opened 2017-06-15
Awesome Secrets

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/9059

StrongMonkey:
```
version: "3.1"
services:
  redis:
    image: redis:latest
    deploy:
      replicas: 1
    secrets:
      - my_secret
      - my_other_secret
secrets:
  my_secret:
    file: ./my_secret.txt
```
This case won' t work because we don't save the `file` fields on service API so when exporting, cattle doesn't know the original value for that file. Need to both rancher-compose and cattle changes to make this case works.

PR #2778, opened 2017-06-20
Make genericObject.resourceData updatable

cjellick:
This will provide greater flexibility for microservices that use
genericObject to store data.

See any obvious downsides to this @ibuildthecloud? Seems limited since the only udpatable field is the relatively newly introduced resourceData


deniseschannon:
https://github.com/rancher/rancher/issues/9113

ibuildthecloud:
LGTM

PR #2779, opened 2017-06-21
Allow environment members to add users

amy:
To enable set the following setting:
`set.member.roles=owner,member`

Addresses part of issue: https://github.com/rancher/rancher/issues/9017


vincent99:
@amy @cjellick Why?  This is eliminating the primary difference between member and owner, the other being the ability to delete it.

And I see no logic to prevent a member from adding a new owner and gaining that ability.

cjellick:
@vincent99 because customers. But yes, you're right. We need to add validation such that only owner can add other owners.
@amy I'm going to modify this PR to do that.

PR #2786, opened 2017-06-22
Setting for roles allowed to add users

cjellick:
Add a setting to control which roles are allowed to add members to a
project (as opposed to the current hard-coded "owner").


ibuildthecloud:
LGTM

PR #2791, opened 2017-06-23
pass cattle pid to microservice

StrongMonkey:


StrongMonkey:
for ppl using cattle_on_eclipse_for_mac😀

PR #2795, opened 2017-06-26
Changes needed for public-dns (service/container name limit and uniqueness, install_uuid)

prachidamle:
1) Upon Startup, insert an install_uuid into setting table
2) Limit length of service and container name to 50
3) Ensure that the containerName and serviceName are unique in a stack
(cannot have a container foo and service foo in same stack)
4) Container name has to be a valid DNS name (no hyphen at start/end,
alpha-numeric)
5) Add install.uuid to settings.public and also provide agent containers access to public
    settings

Added integration tests for above changes.

This PR is part of https://github.com/rancher/rancher/issues/8790


prachidamle:
Current Cattle integration tests are failing on master - fail in common_fixtures:: 
Tries to list superadmin account: and gets data[]
http://localhost:8080/v2-beta/projects/1a5/accounts?name=superadmin

https://github.com/rancher/cattle/blob/master/tests/integration/cattletest/core/common_fixtures.py#L373

@ibuildthecloud @cjellick any idea?

prachidamle:
Integration tests are failing due to new DNS name validation rules applied to container_names.

Primarily the tests are failing since "underscore" is not allowed by this PR in container_names - Do we allow "_"? @ibuildthecloud @vincent99 @cjellick

vincent99:
Cattle/the agent will convert backported container names from unerscore to hyphen.

Public DNS should just skip any entries that have a label that is invalid for whatever reason.  (There can be old installs that have spaces, random symbols, etc in them.)

prachidamle:
@cjellick please review/merge the Cattle changes. All tests have now passed.

cjellick:
We are not going to do this for 1.6

PR #2796, opened 2017-06-27
Use event for instance data cache invalidation

cjellick:
This change adresses a bug where in an HA setup, when a load balancer
was updated to have a new port rule, the newly published port may not
show up in metadata and thus not get programmed properly by networking.
The bug was caused by not invalidating the instance data cache on all
nodes and thus potentially serving up stale data to the metadata
service.

Addresses https://github.com/rancher/rancher/issues/9172


ibuildthecloud:
The failure in `test_ebs_volume_hard_affinity_no_override` seems odd but probably unrelated.  Besides that LGTM

PR #2797, opened 2017-06-27
Check for empty machine.extractedConfig

cjellick:
Without this check, its possible to get an NPE if the user attempts to
download the machine config while the machine is still provisioning.

Addresses https://github.com/rancher/rancher/issues/9169


prachidamle:
LGTM

PR #2798, opened 2017-06-27
fix id obfuscator

StrongMonkey:


cjellick:
Close in favor of #2799. Backport to master will happen separately

PR #2799, opened 2017-06-27
fix id obfuscator 1.6

StrongMonkey:


cjellick:
Generally, I don't like the approach passing the formatted id into the constraints as a string. You should pass the formatter into the constraints and only use the formatter in the toString() method

StrongMonkey:
not sure if tests seem unrelated

StrongMonkey:
https://github.com/rancher/rancher/issues/8653

cjellick:
LGTM

PR #2800, opened 2017-06-28
Insert an install.uuid in setting table during Cattle startup.

prachidamle:
Code to insert an install.uuid in setting table during Cattle startup.
Also added a test to verify that the setting exists.

The uuid can be read via API /v2-beta/settings/install.uuid

@cjellick please review/merge.


prachidamle:
Unrelated TF "Not enough environment ports to create service"

cjellick:
LGTM

PR #2805, opened 2017-06-29
Infra fixes

cjellick:
Only allow roles whitelisted in `modify.infrastructure.roles` to
- CUD hosts and physical hosts (including actions and the machine config download link)
- CRUD registrationTokens

This is needed because gms overrides the schemas for hosts and physicalhosts, so using auth config overrides was not possible.


code/implementation/register/src/main/java/io/cattle/platform/register/process/RegisterTokenAccountCreate.java
```diff
@@ -1,66 +0,0 @@
-package io.cattle.platform.register.process;
```
cjellick: We determine this entire class was unneeded legacy code. Was randomly creating registration tokens that should never be used.

cjellick:
https://github.com/rancher/rancher/issues/9170

code/framework/api/src/main/java/io/cattle/platform/api/resource/jooq/AbstractJooqResourceManager.java
```diff
@@ -85,11 +91,46 @@ protected Object listInternal(SchemaFactory schemaFactory, String type, Map<Obje
...
+            }
+
+            String capability = ObjectUtils.toString(attributes.get(ObjectMetaDataManager.CAPABILITY), null);
+            if (ObjectMetaDataManager.MODIFY_INFRA_CAPABILITY.equals(capability)) {
+                addCapability = true;
+                break;
+            }
+        }
+
+        if (addCapability && infraAccess.canModifyInfrastructure((Policy) ApiContext.getContext().getPolicy())) {
```
alena1108: Nit: can the check for infraAccess.canModifyInfrastructure((Policy) ApiContext.getContext().getPolicy()) be moved up, so when it is false, we can return immediately? This way we can save a bit on going over capabilities.

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/filter/HostValidationFilter.java
```diff
@@ -0,0 +1,56 @@
...
+    ResourceManagerLocator locator;
+    @Inject
+    ObjectManager objMgr;
+    @Inject
+    InfrastructureAccessManager infraAccess;
+
+    @Override
+    public Class<?>[] getTypeClasses() {
+        return new Class<?>[] { Host.class };
+    }
```
alena1108: Should we also validate other resource actions (like upgrade here https://github.com/rancher/cattle/pull/2805/files#diff-322c2fc915f7c6ca0ede9e26989ae879R94) in the validation filter? By overriding resourceAction method similar to how we do it for CRUD

cjellick: The actions actually automatically get validation based off of their presence/absence as controlled by capability attribute. Its baked into the framework.
So for example, if a restricted user attempts to hit `/hosts/1h1?action=deactivate` they get a 404 because they cannot see that action.

alena1108:
LGTM

PR #2806, opened 2017-06-30
add combined options to export one file

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/9069

alena1108:
@StrongMonkey can we have a couple integration tests added for this feature? We already have some validating rancher/docker compose; would be nice to have one or two for the combined approach.

StrongMonkey:
@alena1108 sure, i ll try to see if i can add several tests

PR #2818, opened 2017-07-10
make cleanBadInstance interval configurable

StrongMonkey:


StrongMonkey:
@alena1108 @ibuildthecloud Can either of you review it to see if this is correct place to make the change? Related to the issue  https://github.com/rancher/rancher/issues/8952.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/ServiceDeploymentPlanner.java
```diff
@@ -189,6 +193,10 @@ public Service getService() {
     public void cleanupBadUnits() {
         List<DeploymentUnit> watchList = new ArrayList<>();
         Iterator<DeploymentUnit> it = this.badUnits.iterator();
+        Long interval = CLEAN_BAD_INSTANCES_INTERVAL.get();
+        try {
```
alena1108: @StrongMonkey Blocking reconcile thread doesn't sound like an ideal idea to me. I'd do something more in the line with:

* introduce "mark_as_bad_time" timestamp on a unit level
* throw ServiceReconcileException if the current_time-mark_as_bad_time < cleanup_delay
* reconcile would be attempted again until the bad unit is cleaned up properly

But also from the ticket, it also sounds like the user not only wants to delay the cleanup, which in turn would delay the recreation of the replacement container. They'd probably want something like "recreate, but delay the cleanup of the bad guy", which would require swapping the order of current delete/recreate. And that can get tricky as some resources like retained ip address, can't be assigned to more than one container at a time.
 
@cjellick Let us know your thoughts on the above

 @ibuildthecloud i'm sure this code got changed a lot on a master branch, so the approach might be different there

StrongMonkey:
@alena1108 i will close my pr as we are doing something differently in 2.0

PR #2819, opened 2017-07-11
Requested IPs

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/5297

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DefaultDeploymentUnitInstance.java
```diff
@@ -330,7 +331,19 @@ protected ServiceIndex createServiceIndex() {
...
             String requestedIp = null;
             if (requestedIpObj != null) {
-                requestedIp = requestedIpObj.toString();
+                List<String> availableIPs = new ArrayList<>();
+                if (requestedIpObj instanceof List<?>) {
+                    for (Object item : (List<?>) requestedIpObj) {
+                        availableIPs.add(ObjectUtils.toString(item, null));
+                    }
+                }
+                List<String> usedIPs = context.serviceDao.getUsedIPsFromServiceIndexByServiceID(service);
```
alena1108: Any particular reason we are validating against used IPs used only within the same service? Why not within services of the same account? But even that might not be enough as the requested ip address can be used by standalone container. Also from what I recall, in ip allocation code down below, pool manager would return nothing if the ip address is already allocated in the subnet:

https://github.com/rancher/cattle/blob/b3d7be606152793f615107bdd8ae2017e2e32b62/code/iaas/logic-common/src/main/java/io/cattle/platform/network/impl/NetworkServiceImpl.java#L83

if we prove that the pool manager validates against the duplicates for service requested ips, we can omit adding one extra layer of validation in services code

tests/integration/cattletest/core/test_svc_discovery.py
```diff
@@ -3554,3 +3554,56 @@ def test_sidekicks_container_anti_affinity(client, context):
     for instance in instances:
         result[instance.hostId] = True
     assert len(result) == 3
```
alena1108: Would be nice to have a test validating the case when requested ip address is already allocated to existing service/container.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DefaultDeploymentUnitInstance.java
StrongMonkey: @alena1108 My original thoughts: At createServiceIndex(), we only select available IpAddress from `requestedIpAddress` fields, which is not used by any active service index. Once ip is selected, `context.sdService.allocateIpToServiceIndex` should validate whether ip is used by other services. Before my changes, this part of code also doesn't do any validation against select IP, it just basically get the IP from `requestedIPAddress` and pass it down. But since we are making this field array of string we need to make sure that inside the same service, the ip is not allocated by another service index.

StrongMonkey: I guess i may still need to make this change: if ip is not available in poolManager, continue on the next available ip.

StrongMonkey:
The tests passed locally but failed in CI. Not sure if is related to idempotence.

code/iaas/logic/src/main/java/io/cattle/platform/process/ipaddress/IpAddressActivate.java
```diff
@@ -71,7 +73,7 @@ protected IPAssignment allocateIp(IpAddress ipAddress, Network network) {
         }
 
         if (ip == null) {
-            ip = networkService.assignIpAddress(network, ipAddress, requestedIp);
+            ip = networkService.assignIpAddress(network, ipAddress, Arrays.asList(StringUtils.stripAll(StringUtils.split(requestedIp, ","))));
```
alena1108: We should probably trim the trailing and leading whitespaces here too

alena1108: ah, nvm, didn't notice stripAll

code/iaas/resource-pool/src/main/java/io/cattle/platform/resource/pool/impl/ResourcePoolManagerImpl.java
```diff
@@ -148,14 +153,23 @@ protected String getItem(Map<Object, Object> keys, Object pool, String qualifier
             return null;
         }
 
+        // trying to allocate resource from tryItems
+        for (int i = 0; i < tryItems.size(); i++) {
```
alena1108: Nit: it would probably be neater to use the construct: 
```
for (item in tryItems) {
}
```

code/iaas/resource-pool/src/main/java/io/cattle/platform/resource/pool/impl/ResourcePoolManagerImpl.java
```diff
@@ -148,14 +153,23 @@ protected String getItem(Map<Object, Object> keys, Object pool, String qualifier
...
+            if (tryItems.get(i) != null && generator.isInPool(tryItems.get(i))) {
```
alena1108: We can also validate against the string being not null, but empty here. You can cover both cases by using !StringUtils.isEmpty() clause

alena1108:
LGTM

PR #2822, opened 2017-07-12
Add stack uuid to container metadata

prachidamle:
Adding stack_uuid to container metadata.
With this PR, following PR must also be merged which fixes a nil panic error caused due to addition of stack_uuid for standalone container metadata.
https://github.com/rancher/rancher-metadata/pull/63

This PR is part of rancher/rancher#8790


cjellick:
We are not going to do this for 1.6

PR #2826, opened 2017-07-14
WIP: Retain ip

StrongMonkey:


alena1108:
@StrongMonkey whats the story behind this PR? we want retainIp on per launch config basis?

niusmallnan:
@alena1108 
About this issue https://github.com/rancher/rancher/issues/9317 .
Our new networking option(host-subnet), need to enhance the logic of the scheduler.
It needs retain_ip and AllocatedIPAddress.

alena1108:
@niusmallnan so it is not really about defining it on per container (launchConfig) basis on the API level, but rather propagating the info to the scheduler, correct? 

In 2.0, we start supporting the retainIp field on per container level basis. If we decide to do the same in 1.6.x branch, we need to support 2 formats - on per service basis, and on per container basis - for backwards compatibility. It means that all referencing places in the services code (upgrade is one fo them I believe) have to be revised and modified if needed. Or in 1.6.x we can just derive the boolean flag from the corresponding service field for now, and ip address from the instance's service index.

@StrongMonkey ^^

PR #2832, opened 2017-07-17
Bug fix in LB rules cleanup when selector rule is present

alena1108:
https://github.com/rancher/rancher/issues/9118


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -469,7 +469,7 @@ protected void updateObjectEndPoints(final Object object, final String resourceT
                     .toString());
         }
 
-        if (oldPortToIp.contains(newPortToIp) && newPortToIp.contains(oldPortToIp)) {
+        if (oldPortToIp.containsAll(newPortToIp) && newPortToIp.containsAll(oldPortToIp)) {
```
alena1108: this line is unrelated to the bug fix, just addressed the warning

cjellick:
LGTM

PR #2847, opened 2017-07-26
Removing the check to see if token_type saved in a auth_token matches current auth provider's token_type

prachidamle:
With porting of auth_providers to external auth service this check breaks the auth_token check for upgraded setups, also it is not really necessary to do this check.

https://github.com/rancher/rancher/issues/9472


cjellick:
LGTM

cjellick:
actually code is good but commit message is too long please reformat to follow the rules outlined in "The seven rules of a great Git commit message" from https://chris.beams.io/posts/git-commit/

PR #2852, opened 2017-07-27
Remove externalId from call to refreshToken

mrajashree:
The externalId was being passed to refreshToken in auth-service since ldap didn't have an access token like github. But now auth-service sets the access token to the externalId


mrajashree:
https://github.com/rancher/rancher/issues/9478

prachidamle:
LGTM

cjellick:
LGTM

PR #2853, opened 2017-07-27
Return error messages from auth service

prachidamle:
Auth Service now returns the error messages as returned by auth service, since better error messages are desired.

https://github.com/rancher/rancher/issues/9462
https://github.com/rancher/rancher/issues/9477


cjellick:
LGTM

PR #2858, opened 2017-07-28
export mode with number

StrongMonkey:
Converting mode as number. Rancher-compose executor can't parse mode as string. https://github.com/rancher/rancher/issues/9501


StrongMonkey:
Also the number also matches the doc from both [docker](https://docs.docker.com/compose/compose-file/#long-syntax-2) and [rancher](http://rancher.com/docs/rancher/v1.6/en/cattle/secrets/#example-changing-parameters-docker-composeyml).

alena1108:
LGTM

PR #2861, opened 2017-07-31
Extend instance.dns_search_internal db column from 256 to 1024

alena1108:
@cjellick could you review


deniseschannon:
rancher/rancher#9337

cjellick:
LGTM

PR #2866, opened 2017-08-03
Limit Auth Service reloading to few settings

prachidamle:
Too many reloads seem to cause unstability with DynamicProperty values
and also can cause auth-service to be initialised wrongly

https://github.com/rancher/rancher/issues/9419


prachidamle:
TF "Not enough environment ports to create service" is unrelated.

prachidamle:
@cjellick pls. review

cjellick:
LGTM

PR #2867, opened 2017-08-03
Present a generic auth error message to user

prachidamle:
We should not provide detail error message from auth-service to user. 

Instead provide a generic message and log the details.

https://github.com/rancher/rancher/issues/9462


prachidamle:
@cjellick pls. review

cjellick:
LGTM

PR #2868, opened 2017-08-04
Send 403 error when access is forbidden

prachidamle:
When access mode chosen by Admin is "restricted" or "required", not all valid users can login. Only those specified by the siteaccess mode can login. Rest all users should be presented with Forbidden/403/"Your account does not have access" error message.

We were instead throwing Unauthorized error for such users even if their creds are valid.

https://github.com/rancher/rancher/issues/9477


prachidamle:
The TF is a bit confusing because it errors out with an API error 403. However the code commit returns a 403 with specific auth message which is not seen in the TF. 

Also from the code, I dont see my code being called from any other place than auth.

So TF seems unrelated. To be safe @cjellick can you rerun the tests

cjellick:
LGTM

PR #2873, opened 2017-08-10
fix NPE in getIdentity

mrajashree:
https://github.com/rancher/rancher/issues/9456

Caused by regression added with pr for handleResponse https://github.com/rancher/cattle/pull/2854
The original code was retuning null if jsonData was null. The new changes also returned null but only from handleResponse and not from the getIdentity function


mrajashree:
TF unrelated

cjellick:
LGTM

PR #2881, opened 2017-08-16
Do not set rancher network label on k8s host level containers

alena1108:
https://github.com/rancher/rancher/issues/9678

@ibuildthecloud for some reason we set these labels in 2 places - cattle and kubernetes-agent - I've applied the fix in both places ([kubernetes-agent PR](https://github.com/rancher/kubernetes-agent/pull/47))

@leodotcloud Also there seems to be the bigger issue that the presence of `io.rancher.container.network=true` label overrides the networkMode to managed for native containers started with `--net=host` (start container with `docker run -d -it --net=host --label io.rancher.container.network=true ubuntu:latest` and you'll see 10.42.x.x address set on the container); not sure if that needs to be fixed. This PR just makes sure to not set the label on k8s containers with hostNetwork=true.


cjellick:
LGTM

leodotcloud:
@alena1108 Created https://github.com/rancher/rancher/issues/9700 to track the issue that you mentioned.

PR #2883, opened 2017-08-17
New thread pools for services and system operations

alena1108:
https://github.com/rancher/rancher/issues/9680

@ibuildthecloud @cjellick open to thread pool re-name suggestions and/or diff values for pool defaults


ibuildthecloud:
As a bonus you should create a thread pool just for system containers too.

alena1108:
@ibuildthecloud added a separate pool for system blocking calls; could you check if the pool defaults make sense

ibuildthecloud:
@alena1108 I would start with just copying the sizes from the current pool.  So we had one pool 15-35, add two more 15-35.  So we have effectively 45-135.  Then just perf test it and see the memory impact.  The rule of thumb with pool sizes it that you will always get it wrong :)

ibuildthecloud:
LGTM

PR #2887, opened 2017-08-22
Report resource constraints when scheduling fails

cjellick:
We previously implemented a change to include resource constains in the
error message presented to users when the external scheduler could not
find any matching hosts. However, we missed a case where the external
scheduler found an inactive host that was later filtered out by cattle
due to the fact that it was inactive.

This fix addresses that change by moving the logic for including the
resource constraint up a level and always including the constaints if
they are present.

adresses https://github.com/rancher/rancher/issues/9554


StrongMonkey:
LGTM, thanks for the fix!

PR #2890, opened 2017-08-23
Ignore errors during catalog refresh on startup

cjellick:
A change in the catalog-service refresh api behavior necessiates this
change so that cattle can move forward with starting when
catalog-service fails to refresh catalogs.

The api behavior has changed to return a 500 response code f a refresh fails.
Previously, it was just swallowing the errors and always returning a 200.
This, combined with the fact that it would take 10 minutes to return
a response in airgap setups, meant that cattle would effectively block
for 10 minutes in airgap setups before creating any project
templates/infrastructure stacks.

### DIFF
--- a/code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/catalog/impl/CatalogServiceImpl.java
+++ b/code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/catalog/impl/CatalogServiceImpl.java
@@ -30,7 +30,8 @@ import org.apache.http.HttpResponse;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.ResponseHandler;
 import org.apache.http.client.fluent.Request;
-import org.apache.http.client.fluent.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.Yaml;
 
 import com.netflix.config.DynamicBooleanProperty;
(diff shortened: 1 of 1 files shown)


code/implementation/system-stack/src/main/java/io/cattle/platform/systemstack/catalog/impl/CatalogServiceImpl.java
```diff
@@ -247,13 +249,15 @@ protected void refresh() throws IOException {
...
+                url = url.substring(0, url.length() - 1);
             }
+            Request.Post(String.format("%s?refresh&action=refresh", url)).execute().handleResponse((resp) -> {
+                int status = resp.getStatusLine().getStatusCode();
+                if (status >= 400) {
+                    log.warn("Skipping catalog refresh because it failed. Response code: {}, reason: {}", status, resp.getStatusLine().getReasonPhrase());
+                }
+                return null;
+            });
             firstCall = false;
```
alena1108: Not sure how firstCall parameter is used, just wanted to make sure that it is being set to false even though the catalog refresh had failed, is acceptable.

cjellick: Yes, its acceptable/intentional and this doesnt change how it behaves because previous versions of catalog-service basically never returned anything other than 200 to refresh calls. The idea is is that we'll make one best effort attempt at updating and then move on. Up until now, catalog-service always returned 200 on refresh calls, even if the refresh failed, so this doesn't really change the functionality other than explicitly logging a warning.

alena1108:
LGTM

PR #2894, opened 2017-08-25
Logic to send out the 'target.drain' event to haproxy

prachidamle:
This is a WIP PR for the haproxy drain feature. @cjellick  can you pls review.


prachidamle:
The validation test failures (assert svc["drain_timeout_ms"] == 15000) will go away once the rancher_compose_executor PR is merged and version of compose is bumped up.

https://github.com/rancher/rancher-compose-executor/pull/126

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
```diff
@@ -0,0 +1,220 @@
...
+        Instance instance = (Instance) state.getResource();
+        if (instance == null) {
+            return null;
+        }
+        // container.stop -> find container’s service -> find all the LB services of the account of the container
+        // service -> filter out those services having lbConfig.portRules having serviceId = container’s service
+        Service service = loadService(instance);
+        if (service == null) {
+            return null;
+        }
```
cjellick: We should disable draining by default. So once you get the service, find the drainTimeout and if it is 0 or null, do not drain.
This means also taking out the deafult value for the field in the json file you modified

cjellick:
LGTM

PR #2898, opened 2017-08-30
Removed host-api package reference: not needed as its a part of agent now

alena1108:


alena1108:

PR #2899, opened 2017-08-31
Don't add rancher internal search domains when rancher container dns priority is None

kinarashah:
Adding a new system label, "io.rancher.container.dns.priority" and skip adding the internal search domains when its value is None. This along with [plugin-manager fix](https://github.com/rancher/plugin-manager/pull/86) solves [rancher/rancher#9303](https://github.com/rancher/rancher/issues/9303). 

To test, 
1) start any regular user service by creating the service with label "io.rancher.container.dns.priority" : "None". Check its /etc/resolv.conf to confirm it doesn't have any rancher specific domains. 

2) In kubernetes, 
a) use the [k8s catalog template](https://github.com/rancher/rancher-catalog/pull/858) which starts kubelet with this label set. 
b) login to kubelet and see no rancher specific domains are set. 
c) create a simple replication controller, and see that the POD's /etc/resolv.conf doesn't have any rancher specific search domains set. 
d) verify that the skydns resolution still works fine.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
@@ -149,14 +149,27 @@ public class InstancePreCreate extends AbstractObjectProcessLogic implements Pro
 
         for (Long networkId : networkIds) {
             Network network = objectManager.loadResource(Network.class, networkId);
+            Boolean setRancherSearchDomain = true;
             for (String dns : DataAccessor.fieldStringList(network, NetworkConstants.FIELD_DNS)) {
                 List<String> dnsList = DataAccessor.appendToFieldStringList(instance, DockerInstanceConstants.FIELD_DNS, dns);
                 data.put(DockerInstanceConstants.FIELD_DNS, dnsList);
                 data.put(InstanceConstants.FIELD_DNS_INTERNAL, Joiner.on(",").join(dnsList));
             }
-            for (String dnsSearch : DataAccessor.fieldStringList(network, NetworkConstants.FIELD_DNS_SEARCH)) {
-                List<String> dnsSearchList = DataAccessor.appendToFieldStringList(instance, DockerInstanceConstants.FIELD_DNS_SEARCH, dnsSearch);
-                data.put(DockerInstanceConstants.FIELD_DNS_SEARCH, dnsSearchList);
+            if (labels.containsKey(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY)) {
+                String setSearchDomain = labels.get(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY).toString();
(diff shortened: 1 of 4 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -154,6 +154,9 @@ protected void setDns(Instance instance, Map<String, Object> labels, Map<Object,
                 data.put(DockerInstanceConstants.FIELD_DNS, dnsList);
                 data.put(InstanceConstants.FIELD_DNS_INTERNAL, Joiner.on(",").join(dnsList));
             }
+            if (labels.containsKey(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY) && labels.get(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY).toString().trim().equals("None")) {
```
alena1108: @kinarashah 
1) we should add handling for the case when label value is null
2) do comparison in case insensitive manner

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/deployment/impl/unit/DefaultDeploymentUnitInstance.java
```diff
@@ -353,6 +353,16 @@ protected ServiceIndex createServiceIndex() {
 
     @Override
     public List<String> getSearchDomains() {
+        Object serviceLabels = ServiceDiscoveryUtil.getLaunchConfigObject(service, launchConfigName,
+                InstanceConstants.FIELD_LABELS);
+        if (serviceLabels != null) {
+            String setSearchDomain = ((Map<String, String>) serviceLabels)
+                    .get(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY);
+            if (setSearchDomain != null && !setSearchDomain.isEmpty() && StringUtils.equals(setSearchDomain.trim(), "None")) {
```
alena1108: Case insensitive comparison is needed

alena1108:
@kinarashah can you add an integration test for that?

code/iaas/logic/src/main/java/io/cattle/platform/process/instance/InstancePreCreate.java
```diff
@@ -154,6 +154,12 @@ protected void setDns(Instance instance, Map<String, Object> labels, Map<Object,
                 data.put(DockerInstanceConstants.FIELD_DNS, dnsList);
                 data.put(InstanceConstants.FIELD_DNS_INTERNAL, Joiner.on(",").join(dnsList));
             }
+            if (labels.containsKey(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY)) {
+                String setSearchDomain = labels.get(SystemLabels.LABEL_RANCHER_CONTAINER_DNS_PRIORITY).toString();
+                if (setSearchDomain != null && !setSearchDomain.isEmpty() && StringUtils.equalsIgnoreCase(setSearchDomain.trim(), "None")) {
```
alena1108: @kinarashah we have to set dnssearchinternal here

alena1108:
LGTM. TF is unrelated

PR #2906, opened 2017-09-09
add tweaks to recognize special bind mounts

StrongMonkey:
We are rewriting the volume logic for rancher managed storage driver. The rancher managed volume are passed as a bind mount to the container. So we need to parse the mountpoint to see if its a rancher mountpoint.
If so, we make this mountpoint tied to the specific volume so that API/UI would know the active mounts for the volume and show the correct state.


StrongMonkey:
https://github.com/rancher/agent/pull/229

StrongMonkey:
https://github.com/rancher/storage/pull/58

StrongMonkey:
@cjellick This three PR should put our new flex vol framework is a testable state. I tested with rancher-nfs, it is working well. We still need to do the scalability and reliability tests with all the rancher drivers to make sure everything goes well. I will continue testing this to a more reliable state before handling it to QA.

PR #2908, opened 2017-09-11
Fix missing ipAddress for ports when LB update & release port resource when remove

carolynhu:
https://github.com/rancher/rancher/issues/9752
There are three issues being fixed in this PR:
When we first create LB service, scheduler will allocate an ip address for the ports. 
(1).However, when we update LB, the original ip address assigned for the old ports is missing (2).also, there is no ip allocated for new ports.

https://github.com/rancher/rancher/issues/9879
(3).port resource fail to release when remove

### DIFF
--- a/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/AllocatorDao.java
+++ b/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/AllocatorDao.java
@@ -62,4 +62,6 @@ public interface AllocatorDao {
     String getAllocatedHostUuid(Volume volume);
 
     Iterator<AllocationCandidate> iteratorHosts(List<String> orderedHostUuids, List<Long> volumes, QueryOptions options);
+
+    void updateInstancePorts(List<Map<String, Object>> dataList);
  }
--- a/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
+++ b/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
@@ -42,8 +42,10 @@ import io.cattle.platform.core.model.tables.records.InstanceRecord;
 import io.cattle.platform.core.model.tables.records.PortRecord;
 import io.cattle.platform.core.model.tables.records.StoragePoolRecord;
 import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
+import io.cattle.platform.eventing.EventService;
 import io.cattle.platform.object.ObjectManager;
 import io.cattle.platform.object.util.DataAccessor;
+
 import io.github.ibuildthecloud.gdapi.condition.ConditionType;
 
 import java.util.ArrayList;
@@ -98,7 +100,9 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
     ObjectManager objectManager;
     @Inject
     GenericMapDao mapDao;
-
+    @Inject
+    EventService eventService;
+    
     @Override
     public boolean isInstanceImageKind(long instanceId, String kind) {
         return create().select(STORAGE_POOL.fields())
@@ -533,7 +537,7 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
         }
         return instances;
     }
-    
+
     /* (non-Java doc)
      * @see io.cattle.platform.allocator.dao.AllocatorDao#updateInstancePorts(java.util.Map)
      * Scheduler will return a list of map showing the allocated result in the following format:
@@ -559,7 +563,9 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
      * Port Table. Binding address field in port table and public port field in port table if public port is allocated by external scheduler (for agent)  
      */
     @SuppressWarnings("unchecked")
-    private void updateInstancePorts(List<Map<String, Object>> dataList) {
+    @Override
+    public void updateInstancePorts(List<Map<String, Object>> dataList) {
+    		
         for (Map<String, Object> data: dataList) {
             if (data.get(INSTANCE_ID) == null) {
                 continue;
@@ -576,7 +582,7 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
                 String protocol = (String) allocatedIp.get(PROTOCOL);
                 Integer publicPort = (Integer) allocatedIp.get(PUBLIC_PORT);
                 Integer privatePort = (Integer) allocatedIp.get(PRIVATE_PORT);
-                for (Port port: objectManager.children(instance, Port.class)) {
+                for (Port port : objectManager.children(instance, Port.class)) {
                     if (port.getPrivatePort().equals(privatePort) 
                             && StringUtils.equals(port.getProtocol(), protocol) 
                             && (port.getPublicPort() == null || port.getPublicPort().equals(publicPort))) { 
@@ -590,7 +596,6 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
         }
         return;
     }
-    
 
     @Override
     public Iterator<AllocationCandidate> iteratorHosts(List<String> orderedHostUuids, List<Long> volumes, QueryOptions options) {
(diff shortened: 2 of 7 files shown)


tests/integration-v1/cattletest/core/test_port.py
```diff
@@ -33,19 +33,19 @@ def test_container_port_create_start(super_client, new_context):
             assert port.publicPort is None
             assert port.protocol == 'tcp'
             assert port.instanceId == c.id
-            assert port.state == 'inactive'
+            assert port.state == 'requested'
```
alena1108: We should figure out why the test is failing; modifying the test so it just passes, is not quite correct.

carolynhu: @alena1108 just a test to see whether there is only one error with my pr ☺

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -591,6 +596,35 @@ private void updateInstancePorts(List<Map<String, Object>> dataList) {
         return;
     }
     
+    @SuppressWarnings("unchecked")
+	private void updateInstancePortsForLoadBalancer(List<Map<String, Object>> dataList, List<Port> newPorts) {
```
alena1108: I'd change the name for this method. It should not be load balancer specific. Why don't we just put all the logic to updateInstanceForNewPorts

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/AllocatorDao.java
```diff
@@ -62,4 +62,6 @@
     String getAllocatedHostUuid(Volume volume);
 
     Iterator<AllocationCandidate> iteratorHosts(List<String> orderedHostUuids, List<Long> volumes, QueryOptions options);
+
+    void updateInstanceForNewPorts(List<Map<String, Object>> dataList, List<Port> newPorts);
```
alena1108: Usually it is not a very good idea to pass a generic field to the method call. Can you see if we can make the "dataList" type more concrete + the name should reflect the meaning, dataList as a name is too generic as well

alena1108: Also are we just updating ports in this method, or instance as well? If just ports, the function name should reflect that

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/impl/AllocatorServiceImpl.java
```diff
@@ -205,6 +206,16 @@ public void volumeDeallocate(Volume volume) {
     }
 
     @Override
+    public void newPortAllocateForLoadBalancerUpdate(Instance instance, List<Port> newPorts) {
+        callExternalSchedulerForNewPorts(instance, newPorts);
```
alena1108: Lets name the methods as:

callExternalScheduleToAllocatePorts
callExternalSchedulerToRemovePorts

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/AllocatorService.java
```diff
@@ -19,5 +20,9 @@
 
     void ensureResourcesReservedForStart(Instance instance);
 
+    void newPortAllocateForLoadBalancerUpdate(Instance instance, List<Port> newPorts);
```
alena1108: Lets change the methods names to:

allocatePortsForInstanceUpdate
releasePortsForInstanceUpdate

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/NetworkDaoImpl.java
```diff
@@ -165,40 +165,51 @@ public Network getDefaultNetwork(Long accountId) {
...
             if (port.getRemoved() != null || port.getState().equalsIgnoreCase(CommonStatesConstants.REMOVING)) {
                 continue;
             }
-            existingPorts.put(toKey(port), port);
+            existingInstancePorts.put(toKey(port), port);
         }
 
         for (String port : newPortDefs) {
-            PortSpec spec = new PortSpec(port);
+            PortSpec servicePortSpec = new PortSpec(port);
```
alena1108: Lets not call it "servicePortSpec"; it should be "newPortSpec"

code/iaas/model/src/main/java/io/cattle/platform/core/util/PortSpec.java
```diff
@@ -93,7 +93,7 @@ public Integer getPublicPort() {
         return publicPort;
     }
 
-    public int getPrivatePort() {
+    public Integer getPrivatePort() {
```
alena1108: why this change? private port can never be null. Public port can though

code/iaas/model/src/test/java/io/cattle/platform/core/util/PortSpecTest.java
```diff
@@ -12,31 +12,31 @@ public void testValidFormats() {
         // format: ip:hostPort:containerPort | ip::containerPort | hostPort:containerPort | containerPort
         PortSpec spec = new PortSpec("80");
 
-        assertEquals(80, spec.getPrivatePort());
+        assertEquals(new Integer(80), spec.getPrivatePort());
```
alena1108: we should revert this change. Private port can never be null, and it should be primitive.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceUpdatePostListener.java
```diff
@@ -103,8 +109,18 @@ private void updateServicePorts(Service service, ProcessState state) {
...
-            instance = objectManager.setFields(instance, InstanceConstants.FIELD_PORTS, newPortDefs);
+            List<Port> toUpdate = objectManager.find(Port.class, PORT.INSTANCE_ID, instance.getId(), PORT.REMOVED, null);
+            List<String> toUpdatePortDefs = new ArrayList<>();
+
+            for (Port port : toUpdate) {
+                PortSpec portSpec = new PortSpec(port);
+                toUpdatePortDefs.add(portSpec.toSpec());
+            }
+
+            instance = objectManager.setFields(instance, InstanceConstants.FIELD_PORTS, toUpdatePortDefs);
```
alena1108: I'd put this call (objectManager.setFields(instance, InstanceConstants.FIELD_PORTS, toUpdatePortDefs) to corresponding allocatorDao method. So the listener calls allocator service, allocator service calls external scheduler to allocate the ip address, then Dao is called to persist the ports + update the instance.

code/iaas/model/src/main/java/io/cattle/platform/core/util/PortSpec.java
carolynhu: then, here https://github.com/carolynhu/cattle/blob/d688ecb308f04ac730cbdd456d751bb1f2663d6f/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/NetworkDaoImpl.java#L182
should be check "instancePortSpec.getPrivatePort() == servicePortSpec.getPrivatePort()" not equals

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceUpdatePostListener.java
```diff
@@ -117,6 +133,7 @@ private void updateServicePorts(Service service, ProcessState state) {
             for (Port port : toRemove) {
                 deactivateThenRemove(port, new HashMap<String, Object>());
             }
+            allocatorService.portResourceReleasedForLoadBalancerUpdate(instance, toRemove);
```
alena1108: Comment above applies for release as well.

alena1108:
@carolynhu could you also squash the commits into 1

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/AllocatorDao.java
carolynhu: The name can be changed such as "allocatedDataList", since this dataList is like: [{allocatedIPs=[{allocatedIP=192.168.60.60, privatePort=11, protocol=tcp, publicPort=11}, {allocatedIP=192.168.60.60, privatePort=22, protocol=tcp, publicPort=22}], instanceID=494}] which is hard to make it to a specific data type

alena1108:
LGTM. Drone TF is unrelated.

PR #2910, opened 2017-09-14
DNS: Add search domain instance-{containerUUID}.rancher.internal to every container

alena1108:
Corresponding PRs:

rancher/metadata#4
rancher/plugin-manager#88
rancher/rancher-dns#68

DNS changes:

1. Every container started via rancher gets only one rancher search domain:

instance-{containerUUID}.rancher.internal

2. On the rancher DNS side:
2.1 Answers will have client sections for every {containerUUID} + container ip (to support backwards compatibility) 
2.2 Search domains for clients are

* serviceName.stackName.environmentName.rancher.internal
* stackName.environmentName.rancher.internal
* environmentName.rancher.internal
* rancher.internal
* user domains

2.3 services are resolved as serviceName.stackName.environmentName.rancher.internal
2.4 containers are resolved as containerName.environmentName.rancher.internal
2.5 links are resolved as "alias" (no stack/global search domain appended as it used to be before)

### DIFF
--- a/modules/caas/backend/src/main/java/io/cattle/platform/inator/util/InstanceFactory.java
+++ b/modules/caas/backend/src/main/java/io/cattle/platform/inator/util/InstanceFactory.java
@@ -29,7 +29,6 @@ public class InstanceFactory {
 
         Map<String, Object> fields = new HashMap<>(launchConfig);
         addAdditionalFields(fields, stack, service, unit, launchConfigName);
-        addDns(fields, stack, service);
         addHostnameOverride(instanceName, fields);
 
         fields.put(ObjectMetaDataManager.NAME_FIELD, instanceName);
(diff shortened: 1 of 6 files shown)


modules/caas/backend/src/main/java/io/cattle/platform/inator/util/InstanceFactory.java
```diff
@@ -64,20 +63,9 @@ protected static void addDns(Map<String, Object> fields, StackWrapper stack, Rev
             return;
         }
         List<Object> dns = new ArrayList<>(CollectionUtils.toList(fields.get(InstanceConstants.FIELD_DNS_SEARCH)));
```
ibuildthecloud: Is this code really doing anything now?  I just gets and sets the field.  It will have the side effect of always having a non-null value

PR #2912, opened 2017-09-14
auth fix

StrongMonkey:
fix set project member and don't auto-create new projects for new users


modules/caas/auth/src/main/java/io/cattle/platform/iaas/api/auth/projects/ProjectMemberLinkHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import io.github.ibuildthecloud.gdapi.request.resource.LinkHandler;
+import io.github.ibuildthecloud.gdapi.request.resource.ResourceManager;
+
+import java.io.IOException;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class ProjectMemberLinkHandler implements LinkHandler {
+
+    ResourceManager projectMemberManager;
```
ibuildthecloud: Instead of injecting the ResourceManager, inject ResourceManagerLocator and on each invocation lookup the resource manager.  Because right now I don't think you are actually injecting the right resourcemanager.  Regardless, the prefered approach is to use the locator

PR #2927, opened 2017-09-15
Make auditLog.responseCode an int

cjellick:
store it and return it as an int, so schema should be made to match


cjellick:
PR for master was merged by @ibuildthecloud

cjellick:
addresses https://github.com/rancher/rancher/issues/9890#issuecomment-329841527

PR #2928, opened 2017-09-15
Update host endpoints in a separate loop; 	cluster is the owner of the port in PortUnit

alena1108:


modules/caas/backend/src/main/java/io/cattle/platform/loop/EndpointUpdateLoop.java
```diff
@@ -42,15 +50,19 @@ public Result run(List<Object> input) {
...
-        for (HostInfo hostInfo : metadataManager.getMetadataForCluster(metadata.getClusterId()).getHosts()) {
+        Metadata clusterMetadata = metadataManager.getMetadataForCluster(metadata.getClusterId());
+        for (HostInfo hostInfo : clusterMetadata.getHosts()) {
             if (StringUtils.isNotBlank(hostInfo.getAgentIp())) {
                 agentIps.put(hostInfo.getId(), hostInfo.getAgentIp());
             }
         }
 
+        boolean updated = false;
+        List<String> validStates = Arrays.asList(InstanceConstants.STATE_RUNNING, InstanceConstants.STATE_STARTING, InstanceConstants.STATE_RESTARTING);
```
ibuildthecloud: Just save these as static vars (and as a Set<String>, not List).  It's just wasted cycles to build the list everytime.

modules/caas/backend/src/main/java/io/cattle/platform/inator/unit/PortUnit.java
```diff
@@ -117,8 +117,9 @@ public void bind(InatorContext context, Map<String, Object> instanceData) {
                 if (spec.getPublicPort() == null && spec.getPrivatePort() == port) {
                     Object owner = getOwner(getDeploymentUnit(context));
                     Account account = getAccount(owner);
-                    PooledResource resource = svc.lockManager.lock(new PortUnitLock(account, this), () -> {
-                        return svc.poolManager.allocateOneResource(account, owner,
+                    Account clusterAccount = svc.clusterDao.getOwnerAcccountForCluster(account.getClusterId());
+                    PooledResource resource = svc.lockManager.lock(new PortUnitLock(clusterAccount, this), () -> {
```
ibuildthecloud: You don't need to load the clusterAccount, you can just use the clusterId as the lock key.  This way you avoid another DB call at this point

alena1108: @ibuildthecloud I'll change the lock definition, but looks like we still need a clusterAccount object for pool manager call below.

PR #2933, opened 2017-09-16
Allocate/release ports logic

alena1108:
* lock on cluster
* allocate on start/release on stop

@ibuildthecloud something tells me you might be unhappy with storing portSpecs to the host DB...Couldn't figure out the proper init point for metadata cache where I could catch all instances on the host to transform them to the host's ports constraints.


ibuildthecloud:
@alena1108 yeah, i'd like that info just be cached in memory.  I'll look at it more in depth later.

PR #2950, opened 2017-09-18
Upgrade revision to current for containers that do not need an upgrade

alena1108:


alena1108:
Closing as the revision doesn't have to match

PR #2983, opened 2017-09-21
be able to sync k8s containers

StrongMonkey:


StrongMonkey:
the k8s container are launched with network mode: container and is pointing to its sandbox. But we don't back-populate the pause container in rancher so this network container look up could happen forever. So remove the check if it is k8s pod. The side-effect would be if we launched a native pod in k8s with container networking, this is possible that we will most likely to drop the container field in back-populated containers. 😐

StrongMonkey:
https://github.com/rancher/agent/pull/238

ibuildthecloud:
Good catch, but we are going to have to address this some other way.... Time to put our thinking caps

ibuildthecloud:
superseded by https://github.com/rancher/cattle/pull/2994

PR #2989, opened 2017-09-22
Fix db cleanup 

kinarashah:
[rancher/rancher#7050](https://github.com/rancher/rancher/issues/7050)
[rancher/rancher#7606](https://github.com/rancher/rancher/issues/7606)

I tried reproducing both of the issues mentioned above but couldn't. I observed that 
1) Sometimes, instances without any references also didn't get deleted. 
2) Cleanup took way longer than expected, even with records to be purged ~100k. 

This is because ids were always fetched without any offset, hence the skipped records keep getting fetched again and again. Hence, the number of records getting deleted with every fetch was way below query_limit_rows limit. 
At some point, it so happens that the entire fetched batch has only skipped records and idsToDelete becomes 0 and loop breaks. 

This PR - 
1) fixes the logic of fetching ids with proper offset, so all instances supposed to be deleted get removed. 
2) optimize the way we find ids to be skipped, instead of getting foreign key references for every loop, store them beforehand. They wouldn't change because the cutoff time for the entire cleanup function remains the same. 

Tested with a table that had ~500k+ records to be purged. I adjusted QUERY_LIMIT_ROWS = 10k for easier debugging. 
A sample log showing it deleted expected number and within an expected time range. 
```
2017-09-22 21:07:19,891 INFO  [:] [] [] [] [ecutorService-4] [i.c.p.core.cleanup.TableCleanup     ] idsToDelete 8735
2017-09-22 21:07:48,380 INFO  [:] [] [] [] [ecutorService-4] [i.c.p.core.cleanup.TableCleanup     ] idsToDelete 8986 
2017-09-22 21:08:20,201 INFO  [:] [] [] [] [ecutorService-4] [i.c.p.core.cleanup.TableCleanup     ] idsToDelete 9122 
2017-09-22 21:08:53,219 INFO  [:] [] [] [] [ecutorService-4] [i.c.p.core.cleanup.TableCleanup     ] idsToDelete 8941  
2017-09-22 21:09:24,271 INFO  [:] [] [] [] [ecutorService-4] [i.c.p.core.cleanup.TableCleanup     ] idsToDelete 9173 
2017-09-22 21:09:58,735 INFO  [:] [] [] [] [ecutorService-4] [i.c.p.core.cleanup.TableCleanup     ] idsToDelete 8873 
```

Note - 
while testing, after one round of cleanup, you might still see instances with state = "purged" and removed < cutoff-time. This is because, those containers had sidekicks. Since sidekicks have a dependency, in the first round - they are not removed but will be successfully deleted in the next cleanup. 
You might see entries similar to - 
```
+--------+-------------------------------------------+--------+---------------------+
| id     | name                                      | state  | removed             |
+--------+-------------------------------------------+--------+---------------------+
| 413574 | ipsec-ipsec-1                             | purged | 2017-02-04 19:45:59 |
| 453905 | ipsec-ipsec-1                             | purged | 2017-02-16 20:06:44 |
| 453909 | network-services-metadata-1               | purged | 2017-02-07 18:35:51 |
| 467119 | elasticsearch-2-elasticsearch-datanodes-4 | purged | 2017-02-08 11:30:19 |
| 467715 | kubernetes-kubectld-1                     | purged | 2017-02-08 12:04:35 |
+--------+-------------------------------------------+--------+---------------------+
```

### DIFF
--- a/code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanup.java
+++ b/code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanup.java
@@ -82,6 +82,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
+import java.util.HashSet;
 
 import org.jooq.Field;
 import org.jooq.ForeignKey;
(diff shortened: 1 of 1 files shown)


code/iaas/model/src/main/java/io/cattle/platform/core/cleanup/TableCleanup.java
```diff
@@ -309,6 +325,13 @@ private void cleanup(String name, List<CleanableTable> tables, Date cutoffTime)
                     log.info(e.getMessage());
                     break;
                 }
+                if (idsToFix.size() > 0) {
+                    ids = create()
+                        .select(id)
+                        .from(table.table)
+                        .where(remove.lt(cutoffTime))
+                        .limit(QUERY_LIMIT_ROWS.getValue()).offset(idsToFix.size());
```
cjellick: I think there is technically no guarantee of order when doing a query and I've actually recently ran some queries against cattle dbs where the ids were in fact out of order. So, should we add an order by id clause to this query?

cjellick:
LGTM

PR #2990, opened 2017-09-22
hide host.remove links for not embedded cluster

StrongMonkey:


modules/caas/api/src/main/java/io/cattle/platform/api/host/HostsOutputFilter.java
```diff
@@ -77,6 +78,13 @@ private Resource addLinks(Object original, Resource converted) {
             if (!add && StringUtils.isNotEmpty((String) converted.getFields().get(FIELD_DRIVER))) {
                 add = canAccessConfig();
             }
+            Object imported = DataAccessor.fromDataFieldOf(original).withKey(HostConstants.FIELD_IMPORTED).get();
+            if (imported != null && (Boolean)imported) {
```
ibuildthecloud: protip: `if (imported instanceof Boolean)`  that will ensure you don't get a classcastexception but also "null" is not an instanceof anything, so it checks for null too.

modules/caas/backend/src/main/java/io/cattle/platform/agent/server/resource/impl/AgentResourcesMonitor.java
```diff
@@ -251,7 +249,10 @@ protected void setIpAddresses(Map<String, Host> hosts, Agent agent, AgentResourc
                 Map<String, Object> labels = CollectionUtils.toMap(data.get(HostConstants.FIELD_LABELS));
                 labels.putAll(CollectionUtils.toMap(data.get(HostConstants.FIELD_CREATE_LABELS)));
                 data.put(HostConstants.FIELD_LABELS, labels);
-
+                Cluster cluster = objectManager.loadResource(Cluster.class, agent.getClusterId());
+                if (cluster != null && StringUtils.equals(DataAccessor.fieldString(cluster, ClusterConstants.FIELD_ORCHESTRATION), ClusterConstants.ORCH_KUBERNETES) && !cluster.getEmbedded()) {
```
ibuildthecloud: protip :)  If you are comparing string to a constant, then do  `CONSTANT.equals(someString)`  That will ensure you never get a NullPointerException

modules/model/src/main/java/io/cattle/platform/core/constants/HostConstants.java
```diff
@@ -17,6 +17,7 @@
     public static final String FIELD_INSTANCE_IDS = "instanceIds";
     public static final String FIELD_IP_ADDRESS = "agentIpAddress";
     public static final String FIELD_LABELS = "labels";
+    public static final String FIELD_IMPORTED = "imported";
```
ibuildthecloud: This isn't a field so name the constant `DATA_IMPORTED` and put in a different section of the file

ibuildthecloud: To be honest the UI might want to show imported hosts differently.  So just make it a field :)

PR #2991, opened 2017-09-22
No network mode for secondary launch config

alena1108:
@ibuildthecloud this PR just disables setting networkMode and requestedIpAddress for sidekicks. There are a couple of issues that are yet to be addressed:

1) I still see ip address being allocated to the sidekick instance, and it is different from the primary container. Seeing that the ip address comes as a part of deploymentsyncrequest, I assume it is being allocated on the backend?

2) What would be the right place to set networkFromContainerId on the container?

### DIFF
--- a/modules/caas/api/src/main/java/io/cattle/platform/api/service/ServiceCreateValidationFilter.java
+++ b/modules/caas/api/src/main/java/io/cattle/platform/api/service/ServiceCreateValidationFilter.java
@@ -5,6 +5,7 @@ import static io.cattle.platform.core.model.tables.ServiceTable.*;
 import io.cattle.platform.core.addon.PortRule;
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.InstanceConstants;
+import io.cattle.platform.core.constants.NetworkConstants;
 import io.cattle.platform.core.constants.ServiceConstants;
 import io.cattle.platform.core.model.Service;
 import io.cattle.platform.core.model.Stack;
@@ -73,6 +74,8 @@ public class ServiceCreateValidationFilter extends AbstractValidationFilter {
 
         validateLaunchConfigs(service, request);
 
+        validateNetworMode(service, request);
+
         validateIpsHostName(request);
 
         request = validateAndSetImage(request, service, type);
@@ -98,6 +101,7 @@ public class ServiceCreateValidationFilter extends AbstractValidationFilter {
         Service service = objectManager.loadResource(Service.class, id);
 
         validateLaunchConfigs(service, request);
+        validateNetworMode(service, request);
         validateSelector(request);
         validateLbConfig(request, type);
         validatePorts(service, type, request);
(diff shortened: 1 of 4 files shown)


modules/caas/api/src/main/java/io/cattle/platform/api/service/ServiceCreateValidationFilter.java
```diff
@@ -97,7 +98,7 @@ public Object create(String type, ApiRequest request, ResourceManager next) {
     public Object update(String type, String id, ApiRequest request, ResourceManager next) {
         Service service = objectManager.loadResource(Service.class, id);
 
-        validateLaunchConfigs(service, request);
+        request = validateLaunchConfigs(service, request);
```
ibuildthecloud: No need to return the request.  You can just modifying it in validationLaunchConfig

modules/caas/api/src/main/java/io/cattle/platform/api/service/ServiceCreateValidationFilter.java
```diff
@@ -367,14 +368,30 @@ protected void validateIpsHostName(ApiRequest request) {
...
+    @SuppressWarnings("unchecked")
+    protected ApiRequest validateLaunchConfigs(Service service, ApiRequest request) {
         Map<String, Object> data = CollectionUtils.toMap(request.getRequestObject());
+
+        List<Object> modifiedSlcs = new ArrayList<>();
+        if (data.get(ServiceConstants.FIELD_SECONDARY_LAUNCH_CONFIGS) != null) {
+            List<Object> slcs = (List<Object>) data.get(ServiceConstants.FIELD_SECONDARY_LAUNCH_CONFIGS);
+            for (Object slcObj : slcs) {
+                Map<String, Object> slc = (Map<String, Object>) slcObj;
+                slc.put(InstanceConstants.FIELD_NETWORK_MODE, NetworkConstants.NETWORK_MODE_CONTAINER);
```
ibuildthecloud: The network mode needs to be set to host or none if the primary is set to host or none.  Also, the primary can not be "--net=container".

Additionally you need to validate standalone container.  That can be in a different PR.

modules/caas/api/src/main/java/io/cattle/platform/api/service/ServiceCreateValidationFilter.java
```diff
@@ -367,14 +368,30 @@ protected void validateIpsHostName(ApiRequest request) {
         }
     }
 
-    protected void validateLaunchConfigs(Service service, ApiRequest request) {
+    @SuppressWarnings("unchecked")
+    protected ApiRequest validateLaunchConfigs(Service service, ApiRequest request) {
         Map<String, Object> data = CollectionUtils.toMap(request.getRequestObject());
+
+        List<Object> modifiedSlcs = new ArrayList<>();
+        if (data.get(ServiceConstants.FIELD_SECONDARY_LAUNCH_CONFIGS) != null) {
```
ibuildthecloud: From a code readability perspective it would be nice of this logic was in a method like `validationNetworkMode`

PR #3028, opened 2017-09-28
Stack stop/startAll

alena1108:


modules/caas/api/src/main/java/io/cattle/platform/api/stack/StackDeactivateServicesActionHandler.java
```diff
@@ -41,24 +41,11 @@ public Object perform(Object obj, ApiRequest request) {
...
-        List<String> validStates = Arrays.asList(CommonStatesConstants.ACTIVE,
-                CommonStatesConstants.ACTIVATING,
-                CommonStatesConstants.UPDATING,
-                ServiceConstants.STATE_RESTARTING);
-        List<String> statesToSkip = Arrays.asList(CommonStatesConstants.REMOVED,
-                CommonStatesConstants.REMOVING,
-                CommonStatesConstants.INACTIVE,
-                CommonStatesConstants.DEACTIVATING);
         for (Service service : services) {
-            if (validStates.contains(service.getState())) {
```
alena1108: Looks like we allow deactivate from basically any state now, so removed validStates check

PR #3029, opened 2017-09-28
LocalStorageMb to host metadata

alena1108:


alena1108:
Closing as we are dropping this field

PR #3032, opened 2017-09-29
Handle null account_id in AgentRemove

cjellick:
When an agent is first created and boostrap into the system
(RegisterCreate), it is created without an account_id set.
The account_id is set in a downstream process (AgentCreate).
This creates a scenario where if something goes wrong in AgentCreate
and the agent is ultimately removed, it will be removed while having a
null account_id. This results in an NPE in AgentRemove.

This addresses that scenario by checking for a null account before
attempting to remove it.


cjellick:
addresses https://github.com/rancher/rancher/issues/9985

PR #3040, opened 2017-10-03
machine-package v0.12.2

cjellick:


deniseschannon:
https://github.com/rancher/rancher/issues/9528

PR #3043, opened 2017-10-03
API Keys stop working when Azure AD token expires, requires web ui login

prachidamle:
When using API keys, we need to make sure the Azure tokens are in API  request object, else we face erroneous conditions when refreshing the token.

https://github.com/rancher/rancher/issues/9986

@cjellick pls review.


prachidamle:
TF does not seem related to the Azure change

cjellick:
LGTM

PR #3050, opened 2017-10-09
Do not add duplicate LB services to avoid sending multiple events to same target

prachidamle:
Sending event to same target for same resourceID causes the event framework to not acknowledge the replies coming for those multiple events.

https://github.com/rancher/rancher/issues/10005

@cjellick pls. review

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
@@ -174,7 +174,7 @@ public class LoadBalancerServiceInstanceStopPreListener extends AgentBasedProces
     }
 
     private List<Service> lookupAllActiveLBServices(Service targetService) {
-        List<Service> lbServices = new ArrayList<>();
+        Map<Long, Service> lbServices = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


prachidamle:
TF's are not related to the commit.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceInstanceStopPreListener.java
```diff
@@ -174,7 +176,7 @@ protected void preProcessEvent(EventVO<?> event, ProcessState state, ProcessInst
     }
 
     private List<Service> lookupAllActiveLBServices(Service targetService) {
-        List<Service> lbServices = new ArrayList<>();
+        Set<Service> lbServices = new HashSet<>();
```
cjellick: I would much prefer that we used service.Id to build the set than the service itself. We don't really guarantee that the thing that implements service has a well implemented hashCode and equals method.

Looking at the current implementation, it compares each field on the object. That is actually much more than we need or want.

You could use a map like `Map<Long, Service>` to rely on keys being unique and then use `map.values()` to get the services at the end.

prachidamle: yep, my bad, should have used this first.

prachidamle:
@cjellick updated the PR, pls review.

cjellick:
LGTM

PR #3051, opened 2017-10-09
Do not send drain events if the target service state is not considered active

prachidamle:
https://github.com/rancher/rancher/issues/10011

@cjellick pls. review


cjellick:
LGTM

PR #3053, opened 2017-10-09
fix the issue that agent stuck in upgrade loop

loganhz:
[rancher/rancher#9164](https://github.com/rancher/rancher/issues/9164)

### DIFF
--- a/resources/content/config-content/bootstrap/bootstrap.sh
+++ b/resources/content/config-content/bootstrap/bootstrap.sh
@@ -114,7 +114,7 @@ upgrade()
             sleep 1
         done
 
-        docker run -d --privileged --name rancher-agent-upgrade -v /var/run/docker.sock:/var/run/docker.sock ${REQUIRED_IMAGE} upgrade
+        timeout 60 docker run --privileged --name rancher-agent-upgrade -v /var/run/docker.sock:/var/run/docker.sock ${REQUIRED_IMAGE} upgrade
         exit 0
     elif [ -n "${REQUIRED_IMAGE}" ]; then
         info Using image ${REQUIRED_IMAGE}


yasker:
Commit message:
1. Describe what's the issue and how you fixed it.
2. Add a reference to the bug.

yasker:
@loganhz Why this container was running as a daemon at first place?

loganhz:
@yasker I don't know the reason.  I think it shouldn't be a daemon.  This issue only happens when the host load is very high or docker responses very slow.

yasker:
@loganhz https://github.com/rancher/cattle/commit/caa4828c101a0fb2dc66f4aa63040923abb316e2

loganhz:
hi @ibuildthecloud, I saw you have added `-d` in [`caa4828`](https://github.com/rancher/cattle/commit/caa4828c101a0fb2dc66f4aa63040923abb316e2) two years ago. May I know the reason for it? 

@cjellick found that `Investigation revealed that there appears to be a race condition in our upgrade logic wherein if the rancher-agent-upgarde container takes about five seconds or longer to complete, rancher-agent removes and re-adds it.`

cjellick:
Change scares me. I don't have context on why it was changed. @ibuildthecloud explicitly made this change, so I bet there's a reason. I'd prefer he gave input

yasker:
I think @ibuildthecloud made that change so we can kill stuck container, which is probably more important than the race condition happens occasionally.

Still, need @ibuildthecloud 's input.

But the only other fix I can think of is increasing the timeout, which not I preferred usually.

yasker:
@loganhz Can you come up with another approach to fix the issue? Except increasing the timeout.

loganhz:
Changed to `timeout 60 docker run --privileged --name rancher-agent-upgrade -v /var/run/docker.sock:/var/run/docker.sock ${REQUIRED_IMAGE} upgrade`

deniseschannon:
Closing as both @cjellick and @ibuildthecloud were uncomfortable with this change. Attempting a different method.

PR #3056, opened 2017-10-09
fix null pointer exception

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/9950

cjellick:
Discussed with @StrongMonkey. I dont think the solution is sufficient. It strips the leading zero. Think we need to fully support round-tripping a value like `0400`

vincent99:
Shouldn't it just be parsed and treated as a string all the time?  I don't see how you're going to "round-trip" `0400` vs `256` (the result of that) back to the way they were originally expressed when they are the same integer (and different than `400`, and all valid).

ibuildthecloud:
yaml sucks.  I don't know how else to deal with this than always have mode as a string.  So in the yaml file 400 and "400" will work because we just do "to string" of 400.  But 0400 will not and I don't see how to support it.

StrongMonkey:
just discussed with @vincent99. We are going to support `mode` if mode is defined as 
1. Integer. The only accepted format is xxxx or xxxx. xxx is not going to be supported as it is hard to support 0400 and 400 at the same time.
2. String representation of `mode`.

StrongMonkey:
https://github.com/rancher/rancher-compose-executor/pull/131

StrongMonkey:
https://github.com/rancher/rancher-compose-executor/pull/131

tests/integration/cattletest/core/test_secrets.py
```diff
@@ -204,10 +213,16 @@ def test_service_secrets_fields(context, secret_context):
         {
             "name": "my_secret2",
             "secretId": secret2.id,
-            "mode": "444",
+            "mode": "0444",
```
alena1108: @StrongMonkey Shall we pass 444 and validate it is 0444 in the exported yml? To actually exercise the code in export that appends 0

StrongMonkey: @alena1108 sure, changed

alena1108:
@StrongMonkey spoke with @vincent99, here is his input on the ticket:

I was thinking more like parse assuming it's a base 8 number, regardless of whether it's quoted or has a 0 in front (edited). like `strconv.ParseInt(ObjectUtils.toString(secretOpts.get(NAME)), 8, 64)`. and blow up if it's invalid because you put in a base-10 number that's not valid octal (i.e. `mode: 292`). I guess that's almost what the PR does by adding a "0" to the front, but not quite. And then always serialize to a 4 char string.

We should also add more tests.

StrongMonkey:
it is easy to validate string input to determine whether it is octal format, but it is hard to differentiate if user puts in a number. if you put 0444 or 292 then yaml will parse 292 for you and you can't really tell which one is the original input. That's why i don't want to do any validation as it is just complicating the problem.

vincent99:
I'm saying don't guess, just take the value as a string and assume it's octal.  This does "what I meant" for anything a user types in, because nobody is actually intending to type in decimal permissions.

So you don't "validate", but ParseInt will return an error if it's not a valid octal number (which basically means it has an 8 or 9 in it).

Input | Interpreted as | Exported as
----|-----------------|-------------
`0` | `--- --- ---` | `"0000"`
`444` | `r-- r-- r--` | `"0444"`
`0444` | `r-- r-- r--` | `"0444"`
`"444"` | `r-- r-- r--` | `"0444"`
`"0444"` | `r-- r-- r--` | `"0444"`
`"1777"` | `rws rwx rwx` | `"1777"`
`"7777"` | `rws rwx rwt` | `"7777"`
`7777` | `rws rwx rwt` | `"7777"`
`292` | 422 Validation Error | n/a
`0292` | 422 Validation Error | n/a
`"292"` | 422 Validation Error | n/a
`"0292"` | 422 Validation Error | n/a

(I don't know why you'd use sticky/setuid/setgid on a secret, but they are technically valid)

PR #3057, opened 2017-10-10
Add drain provider label to LoadBalancers based on image version

prachidamle:
This is added to make sure we send events during drain only to agents where the LB is listening for events. So LB's with older images are not sent the events.

https://github.com/rancher/rancher/issues/10013

@cjellick pls review.


cjellick:
LGTM. Waiting for build

PR #3059, opened 2017-10-10
Upgrade of LB should add the agent and environmentAdmin role always

prachidamle:
On LB upgrades, we were adding the label ""io.rancher.container.agent.role" only if it is not already present.

Now since the value of this label is changing since 1.6.11, we should change the label always on upgrades. No need of the if conditional really.

https://github.com/rancher/rancher/issues/10061

@cjellick pls. review


prachidamle:
drone failures

cjellick:
LGTM

PR #3061, opened 2017-10-10
Cleanup fix - fetch new records on every iteration

kinarashah:
[rancher/rancher#7050](https://github.com/rancher/rancher/issues/7050)
If idsToFix =0, new records still need to be fetched according to old query definition of ids.


cjellick:
LGTM

PR #3066, opened 2017-10-11
LB image comparison should apply patch comparison major, minor

prachidamle:
Also we should use the current rancher LB image name from settings and not a hardcoded value.

@alena1108 pls. review.

@sangeethah I think we should re-test https://github.com/rancher/rancher/issues/10069 after this is merged since the change is in the same area as the fix for this bug.

### DIFF
--- a/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/util/ServiceDiscoveryUtil.java
+++ b/code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/util/ServiceDiscoveryUtil.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 
 import com.netflix.config.DynamicStringProperty;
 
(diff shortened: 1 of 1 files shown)


code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/util/ServiceDiscoveryUtil.java
```diff
@@ -436,19 +438,30 @@ public static void injectBalancerLabelsAndHealthcheck(Map<Object, Object> launch
...
-            DockerImage img = DockerImage.parse(imageUuid);
-            String[] imageParts = img.getFullName().split(":");
-            if (imageParts.length <= 1) {
+            Pair<String, String> instanceImage = getImageAndVersion(imageUuid.toLowerCase());
+            if (instanceImage.getLeft().isEmpty() || instanceImage.getRight().isEmpty()) {
                 return false;
             }
-            if (!imageParts[0].equals("rancher/lb-service-haproxy")) {
+            Pair<String, String> defaultImage = getImageAndVersion(LB_IMAGE_UUID.get().toLowerCase());
+            if (!defaultImage.getLeft().equalsIgnoreCase(instanceImage.getLeft())) {
```
alena1108: looks like equalsignorecase is redundant as we are lowercasing both images anyway

code/iaas/service-discovery/api/src/main/java/io/cattle/platform/servicediscovery/api/util/ServiceDiscoveryUtil.java
```diff
@@ -436,19 +438,30 @@ public static void injectBalancerLabelsAndHealthcheck(Map<Object, Object> launch
     private static boolean doesLBHaveDrainSupport(Map<Object, Object> launchConfig) {
         if(launchConfig.get(InstanceConstants.FIELD_IMAGE_UUID) != null) {
```
alena1108: Nit: it would be cleaner to return to have  if(launchConfig.get(InstanceConstants.FIELD_IMAGE_UUID) == null)  {
return false
}

and then have the rest of the code outside of the if/else block

alena1108:
LGTM

PR #3067, opened 2017-10-12
Cleanup service from alias links on removal

alena1108:
https://github.com/rancher/rancher/issues/10042

@ibuildthecloud still have doubts if we should fix the bug and cleanup the service from links on target service removal. Given that reference to it is in serviceName/stackName, not the serviceId + on when create dns config, the target service is ignored if can't be found.


alena1108:

PR #3070, opened 2017-10-13
Add originalLogin field to token

mrajashree:
https://github.com/rancher/rancher/issues/9922


prachidamle:
@mrajashree can you check the TF's

code/iaas/auth-logic/src/main/java/io/cattle/platform/iaas/api/auth/integration/interfaces/TokenUtil.java
```diff
@@ -28,7 +28,7 @@
 
     Account getOrCreateAccount(Identity user, Set<Identity> identities, Account account);
 
-    Token createToken(Set<Identity> identities, Account account);
+    Token createToken(Set<Identity> identities, Account account, String originalLogin);
```
prachidamle: Can you create an overloaded method here and keep the original unchanged - this way not all auth providers need to change (aure,local,openldap) - only those who have this domain\username issue can use the new method.

prachidamle:
LGTM

PR #3072, opened 2017-10-13
raise error when scheduler failed to reserve

StrongMonkey:
We need to catch the error scheduler rasied during reserving and delete the port we already created. This prevents lb consuming a conflicting port and cause chaos.

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceUpdatePostListener.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceUpdatePostListener.java
@@ -2,6 +2,7 @@ package io.cattle.platform.servicediscovery.process;
 
 import static io.cattle.platform.core.model.tables.PortTable.*;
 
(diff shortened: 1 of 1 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/10104

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/LoadBalancerServiceUpdatePostListener.java
```diff
@@ -106,18 +107,26 @@ private void updateServicePorts(Service service, ProcessState state) {
                 port = objectManager.create(port);
             }
 
-            allocatorService.allocatePortsForInstanceUpdate(instance, toCreate);
-            
-            for (Port port : toRetain.values()) {
-                createThenActivate(port, new HashMap<String, Object>());
-            }
+            try {
```
alena1108: Nit: it would be better for visibility and understanding to surround just `allocatePortsForInstanceUpdate` call with try/catch. As neither createThenActivate/deactivateThenRemove/releasePortsForInstanceUpdate can throw this exception

alena1108:
LGTM

PR #3073, opened 2017-10-13
hack: binding 0.0.0.0 to rancher/net:holder in scheduling phase

StrongMonkey:


StrongMonkey:
closing as we adopted a different approach.

PR #3074, opened 2017-10-16
Default batch size to 1 for system services

alena1108:


deniseschannon:
https://github.com/rancher/rancher/issues/10103

ibuildthecloud:
Does this set to always 1 or default to 1?

alena1108:
@ibuildthecloud "set to always 1" when system. It is already defaulted to 1 in the API, therefor was my question whether we should default it to 1 instead of 2 for system services on rancher-compose side.

alena1108:
Closing as the fix will be done on the rancher-compose side https://github.com/rancher/rancher-compose-executor/pull/135

PR #3083, opened 2017-10-19
Do not send drain event for container with null IP address

prachidamle:
https://github.com/rancher/rancher/issues/10176

@alena1108 pls review. (Also changing LB to not panic)


prachidamle:
TF unrelated

alena1108:
LGTM

PR #3087, opened 2017-10-23
don't compare the old and new spec

StrongMonkey:
if we update the port for lb for the first time, if there is a port conflict it will get stuck at updating. Then if we update the same port spec
for the second time, it will compare the current port specs with the old spec(the first update), which is no different and it willl return success immediately.
So, don't compare the stuff and let all of them go through the external scheduler phase to decide whether those ports are being used or not.


StrongMonkey:
https://github.com/rancher/rancher/issues/10175

alena1108:
LGTM

PR #3093, opened 2017-10-27
Fail to process healthcheck when network is down

alena1108:
https://github.com/rancher/rancher/issues/10217

@cjellick tested by:

* manually changed the state to inactive in the DB for ipsec/networkdriver services
* created new instance with healthcheck enabled. Make sure it is stuck in Init state
* changed the state back to active for the service. Validated that the healthstate update event is resent = instance becomes healthy


cjellick:
LGTM

PR #3094, opened 2017-10-30
Fail allocation if cant do scheduler IP scheduling

cjellick:
Fail allocation if the external scheduler is unavailable and a
user has configured the scheduler IP feature and an allocation
request is created that has a port request that would need the feature
to operate correctly.

Addresses https://github.com/rancher/rancher/issues/10187

### DIFF
--- a/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
+++ b/code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
@@ -11,6 +11,7 @@ import static io.cattle.platform.core.model.tables.InstanceTable.*;
 import static io.cattle.platform.core.model.tables.LabelTable.*;
 import static io.cattle.platform.core.model.tables.MountTable.*;
 import static io.cattle.platform.core.model.tables.PortTable.*;
+import static io.cattle.platform.core.model.tables.ServiceTable.*;
 import static io.cattle.platform.core.model.tables.ServiceExposeMapTable.*;
 import static io.cattle.platform.core.model.tables.StorageDriverTable.*;
 import static io.cattle.platform.core.model.tables.StoragePoolHostMapTable.*;
@@ -95,6 +96,9 @@ public class AllocatorDaoImpl extends AbstractJooqDao implements AllocatorDao {
 
     static final List<String> IHM_STATES = Arrays.asList(new String[] { CommonStatesConstants.INACTIVE, CommonStatesConstants.DEACTIVATING,
             CommonStatesConstants.REMOVED, CommonStatesConstants.REMOVING, CommonStatesConstants.PURGING, CommonStatesConstants.PURGED });
+
+    static final List<String> INACTIVE_STATES = Arrays.asList(new String[] { CommonStatesConstants.INACTIVE, CommonStatesConstants.DEACTIVATING,
+            CommonStatesConstants.UPDATING_INACTIVE, CommonStatesConstants.REMOVING });
 
     @Inject
     ObjectManager objectManager;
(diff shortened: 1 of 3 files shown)


cjellick:
Tested by:
1. Create an enviroment with scheudler IPs configures
2. create a service (svcFoo) with a port mapping with no host IP defined (to trigger scheduler IP logic) with a healthcheck
3. forcing scheduler to go unhealthy by changing the image to something that wont pass the healtcheck
4. once scheduler is unhealthy, cause the service from step 2 to fail its healthcheck
5. observe that cattle attempts to resolve svcFoo's failing healthcheck by deploying a new container
6. new container fails to schedule with error message `Cannot perform IP scheduling because external scheduler is unavailable`
7. restore the scheduler to fully functional and observe that the svcFoo is successfully redeployed

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -702,6 +706,12 @@ protected Condition inHostList(List<String> hostUUIDs) {
         return HOST.UUID.in(hostUUIDs);
     }
 
+    protected Condition elligibleHostCondition() {
+        return AGENT.ID.isNull().or(AGENT.STATE.eq(CommonStatesConstants.ACTIVE))
```
alena1108, line 710: should we add Agent.Removed.isnull to the condition?

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/service/impl/AllocatorServiceImpl.java
```diff
@@ -704,6 +704,21 @@ private String getHostUuid(Instance instance) {
...
     private void callExternalSchedulerToReserve(AllocationAttempt attempt, AllocationCandidate candidate) {
         List<Long> agentIds = getAgentResource(attempt.getAccountId(), attempt.getInstances());
+        if (agentIds.isEmpty() && !attempt.getInstances().get(0).getSystem() && allocatorDao.schedulerServiceEnabled(attempt.getAccountId())
+                && allocatorDao.isScheulderIpsEnabled(attempt.getAccountId())) {
+            List<ResourceRequest> resourceRequests = gatherResourceRequests(attempt.getInstances(), attempt.getVolumes(), null);
+            for (ResourceRequest r : resourceRequests) {
+                if (r instanceof PortBindingResourceRequest) {
+                    PortBindingResourceRequest pr = (PortBindingResourceRequest)r;
+                    for (PortSpec ps : pr.getPortRequests()) {
+                        if (StringUtils.isEmpty(ps.getIpAddress()) || !"0.0.0.0".equals(ps.getIpAddress())) {
```
alena1108: We can probably use !StringUtils.equals("0.0.0.0", ps.getIpAddress()). It should cover both cases

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
```diff
@@ -738,4 +745,36 @@ protected Condition append(Condition base, Condition next) {
...
+    @Override
+    public boolean schedulerServiceEnabled(Long accountId) {
+        return create()
+                .select(SERVICE.ID)
+                .from(SERVICE)
+                .where(SERVICE.ACCOUNT_ID.equal(accountId)
+                .and(SERVICE.SYSTEM.isTrue())
+                .and(SERVICE.REMOVED.isNull())
+                .and(SERVICE.STATE.notIn(INACTIVE_STATES))
+                .and(SERVICE.DATA.like("%io.rancher.container.agent_service.scheduling%"))).fetch().size() > 0;
```
alena1108, line 778: I guess we can assume that the label value will never be false :)

code/iaas/allocator/src/main/java/io/cattle/platform/allocator/dao/impl/AllocatorDaoImpl.java
cjellick: This method is an extraction from another method so that I could reuse it. Since that logic is long established and apparently not causing any issues, I'd rather not change it.

alena1108:
LGTM

PR #3095, opened 2017-10-30
Drop io.rancher.container.network for managed network mode

kinarashah:
This PR - 
1) drops io.rancher.container.network label for non-native containers and native containers with --net=host or --net=container. 
2) Assigns managed network mode to native containers only if the label is set and --net is default or none.

### DIFF
--- a/code/iaas/logic-common/src/main/java/io/cattle/platform/network/impl/NetworkServiceImpl.java
+++ b/code/iaas/logic-common/src/main/java/io/cattle/platform/network/impl/NetworkServiceImpl.java
@@ -103,20 +103,21 @@ public class NetworkServiceImpl implements NetworkService {
     public String getNetworkMode(Map<String, Object> instanceData) {
         Map<String, Object> labels = CollectionUtils.toMap(instanceData.get(InstanceConstants.FIELD_LABELS));
         String mode = ObjectUtils.toString(labels.get(SystemLabels.LABEL_CNI_NETWORK));
-
-        if (mode == null && "true".equals(labels.get(SystemLabels.LABEL_RANCHER_NETWORK))) {
-            mode = NetworkConstants.NETWORK_MODE_MANAGED;
-        }
-
-        if (mode == null) {
-            if (instanceData.containsKey(DockerInstanceConstants.FIELD_NETWORK_MODE)) {
-                mode = ObjectUtils.toString(instanceData.get(DockerInstanceConstants.FIELD_NETWORK_MODE));
-            } else {
+        
+        if(mode == null) {
+            String dataMode = ObjectUtils.toString(instanceData.get(DockerInstanceConstants.FIELD_NETWORK_MODE));
+            if (inAllowedModesForManaged(dataMode) && "true".equals(labels.get(SystemLabels.LABEL_RANCHER_NETWORK))) {
                 mode = NetworkConstants.NETWORK_MODE_MANAGED;
+            } else {
+                mode = dataMode;
             }
         }
 
         return mode;
     }
-
+    
+    private boolean inAllowedModesForManaged(String dataMode) {
+        return dataMode==null || dataMode.equals(NetworkConstants.NETWORK_MODE_NONE) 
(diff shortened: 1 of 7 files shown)


code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
```diff
@@ -77,6 +82,15 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
         return new HandlerResult(data).withShouldContinue(true);
     }
 
+    private Boolean removeLabel(Boolean isNativeContainer, String data_mode) {
```
alena1108: should be primitive boolean. Also I'd change the method name to "needToRemoveLabel"; current name kinda implies that the method will do the removal itself.

alena1108: In Java, camelcase is the standard for the parameter name. So it should be dataMode

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
```diff
@@ -77,6 +82,15 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
...
+        if(!isNativeContainer) {
+            return true;
+        } else if (data_mode.equals("host") && data_mode.equals("container")) {
```
alena1108: I think we have host/container under constants somewhere, NetworkConstants? Hardcoding is undesirable.

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
alena1108: if dataMode is empty, we should return false right away

code/iaas/logic-common/src/main/java/io/cattle/platform/network/impl/NetworkServiceImpl.java
```diff
@@ -103,20 +103,21 @@ public void releaseIpAddress(Network network, Object owner) {
...
-            if (instanceData.containsKey(DockerInstanceConstants.FIELD_NETWORK_MODE)) {
-                mode = ObjectUtils.toString(instanceData.get(DockerInstanceConstants.FIELD_NETWORK_MODE));
-            } else {
+        
+        if(mode == null) {
+            String dataMode = ObjectUtils.toString(instanceData.get(DockerInstanceConstants.FIELD_NETWORK_MODE));
+            if (inAllowedModes(dataMode) && "true".equals(labels.get(SystemLabels.LABEL_RANCHER_NETWORK))) {
                 mode = NetworkConstants.NETWORK_MODE_MANAGED;
+            } else {
+                mode = dataMode;
```
alena1108: are we fine assigning mode=dataMode if dataMode when !inAllowedModes? it falls under this else condition

code/implementation/docker/compute/src/main/java/io/cattle/platform/docker/process/instance/DockerInstancePreCreate.java
```diff
@@ -77,6 +83,17 @@ public HandlerResult handle(ProcessState state, ProcessInstance process) {
         return new HandlerResult(data).withShouldContinue(true);
     }
 
+    private boolean needToRemoveLabel(Boolean isNativeContainer, String dataMode) {
+        if(!isNativeContainer) {
+            return true;
+        }
+        if(!dataMode.isEmpty() && dataMode!=null
```
alena1108: !StringUtils.isEmpty(dataMode) should cover both conditions. Current check will hit NPE when dataMode is null.

alena1108:
@kinarashah do we have a test for 2)?

alena1108:
LGTM

leodotcloud:
Original issue: https://github.com/rancher/rancher/issues/9700

PR #3101, opened 2017-11-03
networkUp only after service active for 5s

kinarashah:
[rancher/rancher#10217](https://github.com/rancher/rancher/issues/10217)

### DIFF
--- a/code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
+++ b/code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
@@ -3,12 +3,9 @@ package io.cattle.iaas.healthcheck.process;
 import static io.cattle.platform.core.model.tables.ServiceTable.*;
 
 import io.cattle.iaas.healthcheck.service.HealthcheckService;
-import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.HealthcheckConstants;
-import io.cattle.platform.core.constants.ServiceConstants;
 import io.cattle.platform.core.dao.ServiceDao;
 import io.cattle.platform.core.model.HealthcheckInstanceHostMap;
-import io.cattle.platform.core.model.Service;
 import io.cattle.platform.core.model.ServiceEvent;
 import io.cattle.platform.engine.handler.HandlerResult;
 import io.cattle.platform.engine.handler.ProcessHandler;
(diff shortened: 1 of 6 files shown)


code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
```diff
@@ -22,15 +24,20 @@
...
 import javax.inject.Inject;
 import javax.inject.Named;
 
 
+
 @Named
 public class ServiceEventCreate extends AbstractObjectProcessHandler implements ProcessHandler, Priority {
 
     private static final List<String> UP_STATES = Arrays.asList(CommonStatesConstants.ACTIVE, CommonStatesConstants.UPDATING_ACTIVE);
+    private static final Long SECOND_MILLIS = 1000L;
```
alena1108: I'd just make configuration in miliseconds. No real need to define 1000L

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
```diff
@@ -104,6 +111,19 @@ private boolean isNetworkUp(long accountId) {
                 return false;
             }
         }
+        if(networkDriverService.getState().equals(CommonStatesConstants.ACTIVE)) {
+            Object obj = DataAccessor.fields(networkDriverService).withKey(ServiceConstants.LAST_ACTIVE).get();
```
alena1108: there is a DataAccessor.FieldDate method that can be used

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
```diff
@@ -22,15 +24,20 @@
...
 import javax.inject.Named;
 
 
+
 @Named
 public class ServiceEventCreate extends AbstractObjectProcessHandler implements ProcessHandler, Priority {
 
     private static final List<String> UP_STATES = Arrays.asList(CommonStatesConstants.ACTIVE, CommonStatesConstants.UPDATING_ACTIVE);
+    private static final Long SECOND_MILLIS = 1000L;
+    private static final long STAY_ACTIVE_SECONDS = ArchaiusUtil.getLong("stay_active.seconds").getValue() * SECOND_MILLIS;
```
alena1108: Should we name it with correlation with service.lastActive parameter?

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
kinarashah: It internally converts the long value to Date and returns null if key not present, I'll need to check for null and convert back to long anyway. 
Or am I understanding it wrong ?

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
kinarashah: lastActiveTime = time service was last active 
Changing it to, STAY_ACTIVE_MILLISECONDS = duration network service needs to stay active for before network's considered Up. 
Does that sound ok?

alena1108:
@kinarashah can you look at the failure to see if it is related to the changes we've made? Looks like it might.

alena1108:
LGTM

PR #3103, opened 2017-11-03
Handle leading/trailing whitespace in labels

cjellick:
We were inconsistently comparing label keys and values because
sometimes we use sql queries to find equality and sometimes we use
java's equals method. The sql queries ignore leading/trailing spaces
while java equals doesn't.

Fix this by adding an equals check after a label is retrieved via query
and by trimming label keys and values before creating them in the db.


superseb:
https://github.com/rancher/rancher/issues/10271

alena1108:
LGTM

PR #3104, opened 2017-11-06
Add ProfitBricks API url to whitelist

alibazlamit:
Add the ProfitBricks API URL to the default whitelist


alibazlamit:
@vincent99 Could you please take a look at this PR so we can move ahead with releasing our UI skeleton.

Thanks

PR #3106, opened 2017-11-07
Trim label key and value during comparison

cjellick:
This addresses a bug where if a label.value with a trailing space already
exists (because it was added by a previous host), adding a new host
with the same label.key and label.value with a trailing space, the
label with the trailing space would be used.


alena1108:
LGTM

deniseschannon:
https://github.com/rancher/rancher/issues/10279

PR #3110, opened 2017-11-13
Filter out machineDriverConfig info for read-only and restricted users

carolynhu:
https://github.com/rancherlabs/the-ranch/issues/87


cjellick:
LGTM

PR #3111, opened 2017-11-14
Add Env admin generic object Resource

gitlawr:
Regarding: https://github.com/rancher/rancher/issues/10303
Added an envAdminGenericObject resource type which is only accessible by env admin role. It is stored in generic_object table with kind=envAdminGenericObject. EnvAdminGenericObject use generic object as the parent. 

To avoid operating envAdminGenericObject via generic object API, a generic object resource manager filter is added(This is because resource data is available from parent type, e.g. given "container,parent=instance", you can get container data in instance API). To avoid database changes and reuse current DB table, `kind=envAdminGenericObject` is used to differentiate envAdminGO & GO.

@cjellick @yasker

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/genericobject/EnvAdminGenericObjectFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/genericobject/EnvAdminGenericObjectFilter.java
@@ -1,0 +1,123 @@
+package io.cattle.platform.iaas.api.filter.genericobject;
+
+import io.cattle.platform.core.constants.GenericObjectConstants;
+import io.cattle.platform.core.model.tables.records.GenericObjectRecord;
+import io.cattle.platform.iaas.api.filter.common.AbstractDefaultResourceManagerFilter;
+import io.cattle.platform.object.meta.ObjectMetaDataManager;
+import io.cattle.platform.util.type.CollectionUtils;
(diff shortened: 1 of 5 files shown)


code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/genericobject/EnvAdminGenericObjectFilter.java
```diff
@@ -0,0 +1,123 @@
...
+    public Object update(String type, String id, ApiRequest request, ResourceManager next) {
+        if(isEnvAdminGenericObject(type, id, request, next)){
+            return null;
+        }
+        return next.update(type, id, request);
+    }
+
+    @Override
+    public Object delete(String type, String id, ApiRequest request, ResourceManager next) {
+        if(isEnvAdminGenericObject(type, id, request, next)){
```
cjellick, line 109: are you making it impossible to delete envAdminGenObj?

gitlawr: This is a filter intercepting genericobject API. Deletion can be done via envAdminGenObj API where type=envAdmGO

gitlawr:
Considering that this solution cannot achieve real isolation from env & envAdm(environment members can still get access from container shell), I'm gonna close this PR.

PR #3115, opened 2017-12-04
network healthcheck changes

leodotcloud:
https://github.com/rancher/rancher/issues/10326

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
@@ -5,9 +5,11 @@ import io.cattle.platform.api.auth.Policy;
 import io.cattle.platform.api.utils.ApiUtils;
 import io.cattle.platform.core.constants.AgentConstants;
 import io.cattle.platform.core.constants.CommonStatesConstants;
+import io.cattle.platform.core.constants.HealthcheckConstants;
 import io.cattle.platform.core.constants.ServiceConstants;
 import io.cattle.platform.core.dao.AgentDao;
 import io.cattle.platform.core.dao.ServiceDao;
+import io.cattle.platform.core.dao.InstanceDao;
 import io.cattle.platform.core.model.Agent;
 import io.cattle.platform.core.model.HealthcheckInstance;
 import io.cattle.platform.core.model.HealthcheckInstanceHostMap;
@@ -25,7 +27,6 @@ import io.github.ibuildthecloud.gdapi.util.ResponseCodes;
 import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
 
 import java.util.List;
-
 import javax.inject.Inject;
 
 public class ServiceEventFilter extends AbstractDefaultResourceManagerFilter {
@@ -41,6 +42,9 @@ public class ServiceEventFilter extends AbstractDefaultResourceManagerFilter {
 
     @Inject
     ServiceDao serviceDao;
+
+    @Inject
+    InstanceDao instanceDao;
 
     @Override
     public Class<?>[] getTypeClasses() {
@@ -93,9 +97,10 @@ public class ServiceEventFilter extends AbstractDefaultResourceManagerFilter {
         if (!healthcheckInstanceHostMap.getAccountId().equals(resourceAccId)) {
             throw new ClientVisibleException(ResponseCodes.FORBIDDEN, VERIFY_AGENT);
         }
-
-        if (!isNetworkUp(resourceAccId)) {
-            throw new ClientVisibleException(ResponseCodes.CONFLICT);
+        if(!isNetworkStack(resourceAccId, healthcheckInstance.getInstanceId())) {
+            if (!isNetworkUp(resourceAccId)) {
+                throw new ClientVisibleException(ResponseCodes.CONFLICT);
+            }
         }
 
         event.setInstanceId(healthcheckInstance.getInstanceId());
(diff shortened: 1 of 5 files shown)


code/packaging/app-config/src/main/resources/META-INF/cattle/system-services/defaults.properties
```diff
@@ -69,6 +69,8 @@ metadata.increment.processes=account.update, \
     hostipaddressmap.activate, \
     host.remove, \
     host.update, \
+    host.activate, \
+    host.deactivate, \
     instancehostmap.activate, \
```
cjellick, line 82: @alena1108 any bad repercussions from adding host activate and decactivate to metadata.increment.process?

code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/dao/impl/MetaDataInfoDaoImpl.java
```diff
@@ -219,7 +219,7 @@ protected HostMetaData map(List<Object> input) {
             }
         };
 
-        HostTable host = mapper.add(HOST);
+        HostTable host = mapper.add(HOST, HOST.STATE);
```
cjellick: My understanding is that add(HOST) would add all the fields from host, including state. Is that not the case?

leodotcloud: I guess you are right, will remove this line and test.

code/packaging/app-config/src/main/resources/META-INF/cattle/system-services/defaults.properties
alena1108: @cjellick not really. We already have plenty of processes triggering meta update...hard to make it worse. Plus activate/deactivate doesn't happen too often.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
```diff
@@ -114,10 +121,23 @@ private boolean isNetworkUp(long accountId) {
...
+            if (!(service.getState().equals(CommonStatesConstants.ACTIVE) && service.getHealthState().equals(HealthcheckConstants.HEALTH_STATE_HEALTHY))) {
                 return false;
             }
         }
         return true;
     }
+
+    private boolean isNetworkStack(long accountId, long instanceId) {
+        Instance instance = objectManager.findAny(Instance.class, INSTANCE.ID, instanceId);
+        List<? extends Service> services = instanceDao.findServicesFor(instance);
```
alena1108: see if we can change this method in instanceDao to accept instanceid (I bet instanceid is what get used in db query anyway). This way we don't have to load instance on the line 132.

code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
```diff
@@ -114,10 +121,23 @@ private boolean isNetworkUp(long accountId) {
...
+        if(services.size() > 0) {
+            List<Service> network_services = objectManager.find(Service.class, SERVICE.ACCOUNT_ID, accountId, SERVICE.REMOVED, null, SERVICE.STACK_ID,
```
alena1108: You already have the list of services for the instance on line 133, no need to look for them again. What you need to do is - locate the services having kind == network for the instance. You can introduce a new method for that in instanceDao that would accept instanceid and servicekind

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/InstanceDaoImpl.java
```diff
@@ -176,6 +176,15 @@ public Instance getInstanceByUuidOrExternalId(Long accountId, String uuid, Strin
                 .where(SERVICE_EXPOSE_MAP.INSTANCE_ID.eq(instance.getId()))
                 .fetchInto(ServiceRecord.class);
     }
+    
+    public List<? extends Service> findServicesForInstanceId(Long instanceId){
+    		return create().select(SERVICE.fields())
+    				.from(SERVICE)
+    				.join(SERVICE_EXPOSE_MAP)
+    				.on(SERVICE_EXPOSE_MAP.SERVICE_ID.eq(SERVICE.ID))
```
alena1108: should check for removed being null on both service and serviceexposemap

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/InstanceDaoImpl.java
```diff
@@ -176,6 +176,15 @@ public Instance getInstanceByUuidOrExternalId(Long accountId, String uuid, Strin
                 .where(SERVICE_EXPOSE_MAP.INSTANCE_ID.eq(instance.getId()))
                 .fetchInto(ServiceRecord.class);
     }
+    
+    public List<? extends Service> findServicesForInstanceId(Long instanceId){
```
alena1108: should be long primitive

alena1108:
@kinarashah can we squash 2 commits of yours into 1?

code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/InstanceDaoImpl.java
kinarashah: https://github.com/leodotcloud/cattle/blob/e27b2641a783d877d6480644606ff442cd5cc5d8/code/iaas/model/src/main/java/io/cattle/platform/core/dao/impl/InstanceDaoImpl.java#L171

should it be added for `findServicesFor` as well then?

alena1108:
LGTM

PR #3121, opened 2017-12-12
consider instance for gloabl svc healthcheck only if host is active

kinarashah:


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -750,6 +760,16 @@ protected String calculateServiceHealthState(Service service) {
             int startedOnce = 0;
             List<String> runningStates = Arrays.asList(InstanceConstants.STATE_RUNNING);
             for (Instance instance : serviceInstances) {
+                if (isStartOnce(instance)) {
+                    startedOnce++;
+                }
+                if(isGlobal) {
+                    String hostId = DataAccessor.fieldString(instance, InstanceConstants.FIELD_HOST_ID);
```
alena1108: We should use fieldInteger, or fieldLong

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -555,7 +559,13 @@ public void updateHealthState(final Stack stack) {
         List<Service> services = objectManager.find(Service.class, SERVICE.STACK_ID,
                 stack.getId(), SERVICE.REMOVED, null);
 
-        setServiceHealthState(services);
+        List<? extends Host> ActiveHosts = allocatorDao.getActiveHosts(stack.getAccountId());
+        HashSet<String> activeHosts = new HashSet<String>();
```
alena1108: Any particular reason why we convert to String here? I'd rather have it as long

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java
```diff
@@ -750,6 +760,16 @@ protected String calculateServiceHealthState(Service service) {
             int startedOnce = 0;
             List<String> runningStates = Arrays.asList(InstanceConstants.STATE_RUNNING);
             for (Instance instance : serviceInstances) {
+                if (isStartOnce(instance)) {
+                    startedOnce++;
+                }
+                if(isGlobal) {
+                    String hostId = DataAccessor.fieldString(instance, InstanceConstants.FIELD_HOST_ID);
+                    if(!activeHosts.contains(hostId)) {
```
alena1108: why are we doing it this way? It should be the other way around, if the activeHosts.contains, then increment healthy count. Or if we use !activeHosts construct, we should just continue, w/o incrementing anything. And I'd do it at the very beginning of the loop

kinarashah: https://github.com/kinarashah/cattle/blob/2672e4f9a64f4d4f856fad6904c2f2d781302278/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ServiceDiscoveryServiceImpl.java#L791

alena1108:
LGTM

PR #3127, opened 2017-12-28
Supports Chinese PL value

loganhz:
https://github.com/rancher/rancher/issues/7923

### DIFF
--- a/code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
+++ b/code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
@@ -11,6 +11,7 @@ import io.github.ibuildthecloud.gdapi.util.RequestUtils;
 import io.github.ibuildthecloud.gdapi.version.Versions;
 
 import java.io.IOException;
+import java.net.URLEncoder;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


code/framework/api/src/main/java/io/cattle/platform/api/servlet/ApiRequestFilter.java
```diff
@@ -200,7 +204,13 @@ private void addPLCookie(HttpServletRequest httpRequest, HttpServletResponse res
         }
 
         if (plCookie == null || !PL_SETTING.getValue().equalsIgnoreCase(plCookie.getValue())) {
-            plCookie = new Cookie(PL, PL_SETTING.getValue());
+            String plValue = PL_SETTING.getValue();
+            try {
+              plValue = URLEncoder.encode(PL_SETTING.getValue(), "UTF-8");
+            } catch (IOException e) {
+              log.error("Failed to encode PL value with UTF-8.", e);
```
alena1108, line 211: should we return when exception happens? the cookie value would be nil if exception happens

loganhz: added. Thanks

loganhz: added

PR #3143, opened 2018-01-25
region monitor 

kinarashah:
[rancher/rancher#10648](https://github.com/rancher/rancher/issues/10648) 
[rancher/rancher#10336](https://github.com/rancher/rancher/issues/10336)
[rancher/rancher#10299](https://github.com/rancher/rancher/issues/10299)
[rancher/rancher#10334](https://github.com/rancher/rancher/issues/10334) 
[rancher/rancher#10364](https://github.com/rancher/rancher/issues/10364)

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
@@ -1,0 +1,124 @@
+package io.cattle.platform.servicediscovery.process;
+
+import io.cattle.platform.core.constants.AccountConstants;
+import io.cattle.platform.core.constants.CommonStatesConstants;
+import io.cattle.platform.core.model.Account;
+import io.cattle.platform.core.model.AccountLink;
+import io.cattle.platform.core.model.Region;
+
+import static io.cattle.platform.core.model.tables.AccountTable.ACCOUNT;
+import static io.cattle.platform.core.model.tables.AccountLinkTable.ACCOUNT_LINK;
+import static io.cattle.platform.core.model.tables.RegionTable.REGION;
+import io.cattle.platform.db.jooq.dao.impl.AbstractJooqDao;
+import io.cattle.platform.json.JsonMapper;
+import io.cattle.platform.object.ObjectManager;
+import io.cattle.platform.object.process.ObjectProcessManager;
+import io.cattle.platform.object.util.DataAccessor;
+import io.cattle.platform.servicediscovery.service.RegionService;
+import io.cattle.platform.servicediscovery.service.impl.RegionServiceImpl;
+import io.cattle.platform.servicediscovery.service.impl.RegionUtil;
+import io.cattle.platform.servicediscovery.service.impl.RegionUtil.ExternalProject;
+import io.cattle.platform.servicediscovery.service.impl.RegionUtil.ExternalRegion;
+import io.cattle.platform.task.Task;
+import io.github.ibuildthecloud.gdapi.condition.Condition;
+import io.github.ibuildthecloud.gdapi.condition.ConditionType;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RegionMonitor extends AbstractJooqDao implements Task {
+    private static final Logger log = LoggerFactory.getLogger(RegionServiceImpl.class);
+    private static List<String> invalidStates = Arrays.asList(CommonStatesConstants.REGISTERING, CommonStatesConstants.REMOVING, 
+        CommonStatesConstants.REMOVED);
+
+    @Inject
+    ObjectProcessManager objectProcessManager;
+    @Inject
+    ObjectManager objectManager;
+    @Inject
+    JsonMapper jsonMapper;
+    @Inject
+    RegionService regionService;
+
+    @Override
+    public void run() {
+        List<Region> regions = objectManager.find(Region.class, REGION.REMOVED, new Condition(ConditionType.NULL));
+        if (regions.size() == 0) {
+            return;
+        }
+        Region localRegion = null; 
+        Map<Long, Region> regionMap = new HashMap<Long, Region>();
+        for(Region region : regions) {
+            regionMap.put(region.getId(), region);
+            if(region.getLocal()) {
+                localRegion = region;
+            }
+        }
+        List<AccountLink> accountLinks = objectManager.find(AccountLink.class, ACCOUNT_LINK.REMOVED, null, ACCOUNT_LINK.LINKED_REGION_ID, 
+            new Condition(ConditionType.NOTNULL));
+        Map<String, ExternalRegion> externalRegionMap = new HashMap<String, ExternalRegion>();
+        Map<String, ExternalProject> externalProjectMap = new HashMap<String, ExternalProject>();
+        for(AccountLink link : accountLinks) {
+            if (invalidStates.contains(link.getState())) {
+                continue;
+            }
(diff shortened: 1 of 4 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public void run() {
+        List<Region> regions = objectManager.find(Region.class, REGION.REMOVED, new Condition(ConditionType.NULL));
+        if (regions.size() == 0) {
+            return;
+        }
+        List<AccountLink> accountLinks = objectManager.find(AccountLink.class, ACCOUNT_LINK.REMOVED, null, ACCOUNT_LINK.LINKED_REGION_ID, 
+            new Condition(ConditionType.NOTNULL));
+        List<String> invalidStates = Arrays.asList(CommonStatesConstants.REGISTERING, CommonStatesConstants.REMOVING, CommonStatesConstants.REMOVED);
```
alena1108: nit: you can register this as a static variable. This way we don't have to redefine it on every run

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+        for(AccountLink link : accountLinks) {
+            if (invalidStates.contains(link.getState())) {
+                continue;
+            }
+            // targetRegion not present
+            Region targetRegion = objectManager.loadResource(Region.class, link.getLinkedRegionId());
```
alena1108: I'd suggest you create some sort of map for the regions. This way you don't have to pull it from the DB if the link with the same region has been processed already. Also think about pulling all the regions in a single call, and storing them in the map, instead of loading individual objects.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+            if(targetRegion == null) {
+                delete(link);
+                continue;
+            }
+            try {
+                // localRegion in targetRegion not present 
+                Region localRegion = objectManager.findAny(Region.class, REGION.LOCAL, true, REGION.REMOVED, null);
```
alena1108: Same here. Do the list region outside of the loop, save all regions in map, and local region to the special variable. This way we can save on a db call.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+                ExternalRegion externalRegion = RegionUtil.getExternalRegion(targetRegion, localRegion.getName(), jsonMapper);
```
alena1108: almost same here :) save external regions in a map, and try to locate it there first.

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+                if(externalRegion == null) {
+                    delete(link);
+                    continue;
+                }
+                // environment not present or its uuid changed 
+                ExternalProject externalProject = RegionUtil.getTargetProjectByName(targetRegion, link.getLinkedAccount(), jsonMapper);
```
alena1108: And same here :) use map :)

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,85 @@
...
+                String storedUUID = DataAccessor.fieldString(link, "uuid");
+                if(externalProject == null || (storedUUID!=null && !externalProject.getUuid().equals(storedUUID))) {
+                    delete(link);
+                }
+            } catch (Exception ex) {
+                throw new RuntimeException(String.format("Failed to monitor account link for [%d]", link.getId()), ex);
```
alena1108: we should never throw a runtime in a periodic task. Just log it, proceed with the loop. The next loop will retry

kinarashah:
https://github.com/rancher/rancher/issues/10648

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+        Map<String, ExternalRegion> externalRegionMap = new HashMap<String, ExternalRegion>();
+        Map<String, ExternalProject> externalProjectMap = new HashMap<String, ExternalProject>();
+        for(AccountLink link : accountLinks) {
+            if (invalidStates.contains(link.getState())) {
+                continue;
+            }
+            // targetRegion not present
+            Region targetRegion = regionMap.get(link.getLinkedRegionId());
+            if(targetRegion == null) {
+                delete(link);
```
alena1108: nit: lets just rename this method to deleteAccountLink. Or simply call object manager.remove as delete is really just a wrapper around it

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+                } else {
+                    externalProject = RegionUtil.getTargetProjectByName(targetRegion, link.getLinkedAccount(), jsonMapper);
+                    externalProjectMap.put(externalProjectKey, externalProject);
+                }
+                String storedUUID = DataAccessor.fieldString(link, "uuid");
+                if(externalProject == null || (storedUUID!=null && !externalProject.getUuid().equals(storedUUID))) {
+                    delete(link);
+                }
+            } catch (Exception ex) {
+                log.info("Failed to monitor account link for %s - %s", link.getId(), ex);
```
alena1108: lets log in warn

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+        List<Account> accounts = objectManager.find(Account.class, ACCOUNT.REMOVED, new Condition(ConditionType.NULL));
+        for(Account account : accounts) {
+            regionService.reconcileExternalLinks(account.getId());
```
alena1108: lets wrap it in try/catch too. There can be process cancel exception thrown underneath

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+                String storedUUID = DataAccessor.fieldString(link, "uuid");
+                if(externalProject == null || (storedUUID!=null && !externalProject.getUuid().equals(storedUUID))) {
+                    delete(link);
+                }
+            } catch (Exception ex) {
+                log.info("Failed to monitor account link for %s - %s", link.getId(), ex);
+            }
+        }
+        List<Account> accounts = objectManager.find(Account.class, ACCOUNT.REMOVED, new Condition(ConditionType.NULL));
```
alena1108: we should filter out accounts by kind I believe. There are diff account kinds like agent that we don't have to reconcile. We should only do it for the user account. Check what corresponding account kind is.

PR #3155, opened 2018-01-31
region fixes

kinarashah:
- avoid recreating link if invalid 
- links should have targetResourceAccount's linked_account_uuid 
- distinguish between null and 404 - delete only if 200 && null 
- agent should be deleted once link gets deleted

### DIFF
--- a/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
+++ b/code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
@@ -18,6 +18,7 @@ import io.cattle.platform.servicediscovery.service.RegionService;
 import io.cattle.platform.servicediscovery.service.impl.RegionServiceImpl;
 import io.cattle.platform.servicediscovery.service.impl.RegionUtil;
 import io.cattle.platform.servicediscovery.service.impl.RegionUtil.ExternalProject;
+import io.cattle.platform.servicediscovery.service.impl.RegionUtil.ExternalProjectResponse;
 import io.cattle.platform.servicediscovery.service.impl.RegionUtil.ExternalRegion;
 import io.cattle.platform.task.Task;
 import io.cattle.platform.object.process.StandardProcess;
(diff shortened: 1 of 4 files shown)


code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/service/impl/ExternalProjectResponse.java
```diff
@@ -0,0 +1,16 @@
+package io.cattle.platform.servicediscovery.service.impl;
```
alena1108: it's better to not put it to the separate file; either put it to regionutil, or regionserviceimpl

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -90,16 +91,19 @@ public void run() {
...
                 } else {
-                    externalProject = RegionUtil.getTargetProjectByName(targetRegion, link.getLinkedAccount(), jsonMapper);
-                    externalProjectMap.put(externalProjectKey, externalProject);
+                    externalProjectResponse = RegionUtil.getTargetProjectByName(targetRegion, link.getLinkedAccount(), jsonMapper);
+                    externalProjectMap.put(externalProjectKey, externalProjectResponse);
                 }
-                String storedUUID = DataAccessor.fieldString(link, "uuid");
-                if(externalProject == null || (storedUUID!=null && !externalProject.getUuid().equals(storedUUID))) {
+
+                String storedUUID = DataAccessor.fieldString(link, "linked_account_uuid");
```
alena1108: it really should be linkedAccountUuid

resources/content/schema/base/accountLink.json
```diff
@@ -0,0 +1,7 @@
+{
+	"linked_account_uuid" : {
```
alena1108: should be camelcased `linkedAccountUuid`

code/iaas/service-discovery/server/src/main/java/io/cattle/platform/servicediscovery/process/RegionMonitor.java
```diff
@@ -90,16 +91,19 @@ public void run() {
...
+                ExternalProject externalProject = externalProjectResponse.getExternalProject();
+                if((externalProject == null && externalProjectResponse.getStatusCode()==200)
```
alena1108: Nit: can we assign 2 conditions to vars, and then use the vars in if clause? So the condition one would be `boolean notfound` and condition 2 `uuidsDoNotMatch` or something

alena1108:
LGTM

PR #3159, opened 2018-02-06
Recreate container level volume on down host for local volumes only

alena1108:
https://github.com/rancher/rancher/issues/11310


cjellick:
LGTM

PR #3176, opened 2018-03-13
Explicitly specify user member attributes for openldap

mrajashree:
https://github.com/rancher/rancher/issues/10292


cjellick:
LGTM

PR #3188, opened 2018-04-02
add backoff time to service.trigger

StrongMonkey:


cjellick:
This needs much more investigation and consideration before merging

alena1108:

PR #3200, opened 2018-05-01
Check for duplicate existing k8s service before creating it

kinarashah:
https://github.com/rancher/rancher/issues/13123

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
@@ -1,9 +1,11 @@
 package io.cattle.platform.process.externalevent;
 
 import static io.cattle.platform.core.constants.ExternalEventConstants.*;
+import static io.cattle.platform.core.model.tables.ServiceTable.SERVICE;
 
 import io.cattle.platform.core.constants.CommonStatesConstants;
 import io.cattle.platform.core.constants.ExternalEventConstants;
+import io.cattle.platform.core.constants.ServiceConstants;
 import io.cattle.platform.core.dao.GenericResourceDao;
 import io.cattle.platform.core.dao.ServiceDao;
 import io.cattle.platform.core.dao.StackDao;
@@ -30,6 +32,7 @@ import io.github.ibuildthecloud.gdapi.factory.SchemaFactory;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.List;
 import java.util.concurrent.Callable;
 
 import javax.inject.Inject;
@@ -41,7 +44,6 @@ import org.slf4j.LoggerFactory;
 
 @Named
 public class ExternalServiceEventCreate extends AbstractDefaultProcessHandler {
-
     private static final Logger log = LoggerFactory.getLogger(ExternalServiceEventCreate.class);
 
     @Inject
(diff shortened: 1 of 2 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
```diff
@@ -106,6 +108,21 @@ void createService(ExternalEvent event, Map<String, Object> serviceData) {
             log.info("Can't process service event. Could not get or create stack. Event: [{}]", event);
             return;
         }
+        
+        List<Service> existingSvcs = objectManager.find(Service.class,
```
alena1108, line 112: you can use findAny

kinarashah: Then this won't remove the existing multiple services?

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
```diff
@@ -106,6 +108,21 @@ void createService(ExternalEvent event, Map<String, Object> serviceData) {
...
+        
+        List<Service> existingSvcs = objectManager.find(Service.class, 
+            SERVICE.NAME, serviceData.get(FIELD_NAME),
+            SERVICE.REMOVED, null, 
+            SERVICE.STACK_ID, stack.getId(), 
+            SERVICE.ACCOUNT_ID, event.getAccountId(),
+            SERVICE.KIND, ServiceConstants.KIND_KUBERNETES_SERVICE);
+
+        for(Service existingSvc : existingSvcs) {
+            if(existingSvc!=null && !existingSvc.getExternalId().equals(event.getExternalId())) {
```
alena1108: just to be on the safe side, lets check for nil value on externalId

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
```diff
@@ -106,6 +108,21 @@ void createService(ExternalEvent event, Map<String, Object> serviceData) {
...
+                log.info("Deleting existing kubernetes service {} with externalId {} and accountId {}",
```
alena1108: lets not log accountId

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
alena1108: actually...lets leave it the way it is; it can take care of multiple services cleanup - the ones that were created prior to this fix

code/iaas/logic/src/main/java/io/cattle/platform/process/externalevent/ExternalServiceEventCreate.java
```diff
@@ -106,6 +108,21 @@ void createService(ExternalEvent event, Map<String, Object> serviceData) {
...
+                            existingSvc.getName(), existingSvc.getExternalId(), existingSvc.getAccountId());
+                objectProcessManager.executeStandardProcess(StandardProcess.REMOVE, existingSvc, null);
```
alena1108: is remove allowed from active? for services. Or should we call deactivate then remove? 
Also it might be a good idea to check for Removing state, and skip those services

PR #3202, opened 2018-05-14
Health check fixes

kinarashah:
- drop healthcheck only when service event starts with down and network stack is upgrading
- ignore instances having healthcheck noop for cleanup
- always include the host where instance is deployed in the healthcheckers 

https://github.com/rancher/rancher/issues/13398

### DIFF
--- a/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
+++ b/code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
@@ -5,7 +5,6 @@ import io.cattle.platform.api.auth.Policy;
 import io.cattle.platform.api.utils.ApiUtils;
 import io.cattle.platform.core.constants.AgentConstants;
 import io.cattle.platform.core.constants.CommonStatesConstants;
-import io.cattle.platform.core.constants.HealthcheckConstants;
 import io.cattle.platform.core.constants.ServiceConstants;
 import io.cattle.platform.core.dao.AgentDao;
 import io.cattle.platform.core.dao.ServiceDao;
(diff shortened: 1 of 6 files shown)


code/iaas/api-logic/src/main/java/io/cattle/platform/iaas/api/filter/serviceevent/ServiceEventFilter.java
```diff
@@ -33,6 +32,10 @@
 public class ServiceEventFilter extends AbstractDefaultResourceManagerFilter {
 
     private static List<String> invalidStates = Arrays.asList(CommonStatesConstants.REMOVED, CommonStatesConstants.REMOVING);
+    private static List<String> upgradingStates = Arrays.asList(
+            ServiceConstants.STATE_UPGRADING,
```
alena1108, line 36: we should put all upgrade related states to the list, like `canceled-upgrade`, `upgraded`, `finishing-upgrade`

code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/service/impl/HealthcheckServiceImpl.java
```diff
@@ -297,21 +297,23 @@ protected void createHealthCheckHostMaps(HealthcheckInstanceType instanceType, l
...
+   
         // place inferiorHostId to the end of the list
         if (inferiorHostId != null) {
-            if (availableActiveHostIds.contains(inferiorHostId)) {
-                availableActiveHostIds.remove(inferiorHostId);
-                if (availableActiveHostIds.isEmpty() && allocatedActiveHostIds.isEmpty()) {
+                if (availableActiveHostIds.contains(inferiorHostId)) {
+                    availableActiveHostIds.remove(inferiorHostId);
+                }
+                if(!allocatedActiveHostIds.contains(inferiorHostId)) {
```
alena1108, line 306: I don't think this condition can be true at this point; so we can just remove, and put it back

kinarashah: we don't want id to be considered available if it's already allocated?

alena1108: sorry, read `allocatedActiveHostIds` as `availableActiveHostIds`

alena1108:
LGTM

PR #3205, opened 2018-05-25
Add more logging around websocket subscriptions

cjellick:
This introduces two logging changes:
1. Add the host id to the error that we log when  an agent misses the maximum number of pings and we log that we are scheduling a reconnect
2. When a webscocket connection is made for subscribing to events, If that subscription is **for an agent** (as opposed to a user in the UI), log more details about connects, disconnects, and errors. Sepcifically:
    - log when the connection is made and the agent for which it was made
    - log when the connection is closde and the agent for which it was made
    - if an abnormal exception is encountered, log it and the corresponding agent. I say "abnormal" exception because the websocket library we groups exceptions into two types. Some are "normal" and handle by the library. These include the connection being closed. Others are abnormal and the library provides a hook for doing something when they are encountered.


ibuildthecloud:
LGTM

deniseschannon:
https://github.com/rancher/rancher/issues/13674

PR #3210, opened 2018-07-02
Agent auth: handle the case when account does not exist

alena1108:
https://github.com/rancher/rancher/issues/11621


cjellick:
LGTM

PR #3212, opened 2018-07-06
Add setting to control restricted user exec

cjellick:
We currently prevent restricted users from execing into privileged containers
or ones that have capAdd set to any non-empty value. This change adds a setting
that can be used to disable that functionality.


cjellick:
Addresses https://github.com/rancher/rancher/issues/13886

alena1108:
LGTM

PR #3216, opened 2018-07-09
Make Azure admin credential fields optional

prachidamle:
Making Azure Admin username/password optional, since these fields are really not required for any functionality.

https://github.com/rancher/rancher/issues/11614


cjellick:
LGTM

PR #3218, opened 2018-07-11
add lock even when volume is active and detached

StrongMonkey:
we should also add lock in scheduling instances when volume is active
but detached. This will prevent a race condition when we try to
schedule instances using an existing volume or when it failed to pass
healthcheck and reschedule containers using existing volume.


StrongMonkey:
https://github.com/rancher/rancher/issues/14484

cjellick:
LGTM
test failure appears unrelated

PR #3219, opened 2018-07-12
always add lock when volume is singleHostRW

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/14484

cjellick:
LGTM

PR #3220, opened 2018-07-12
Adding few log statements to debug Azure issue where user cannot login

prachidamle:
Adding some log statements at TRACE to capture error from Azure.


cjellick:
lgtm

PR #3222, opened 2018-07-16
Validate TenantId while configuring Azure auth

prachidamle:
While configuring Azure auth, we take in parameters like 'tenantId', 'clientId, 'domain', along with azure credentials to test.

For testing the creds, 'tenantId' parameter is not required, so we miss out on finding if that parameter is wrong and when creds are authenticated we turn on the auth. Later while pulling the identities of that user, this tenantId is needed and this API then starts failing and blocking the user's login to Rancher.

Solution: Make an extra search request (for the username) to Azure during configuring auth which will need us to use the tenantID in the request - that will validate if the parameter is correct and error if not , avoiding this situation

@cjellick @deniseschannon need review
https://github.com/rancher/rancher/issues/14541


alena1108:
LGTM

PR #3225, opened 2018-07-17
store metadata info only for ipsec service

kinarashah:
https://github.com/rancher/rancher/issues/14587

### DIFF
--- a/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ServiceMetaData.java
+++ b/code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ServiceMetaData.java
@@ -142,9 +142,7 @@ public class ServiceMetaData {
             this.health_check = new HealthCheck(healthCheck);
         }
         this.system = service.getSystem();
-        Map<String, Object> service_metadata = DataAccessor.fieldMap(service, ServiceConstants.FIELD_METADATA);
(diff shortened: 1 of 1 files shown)


code/iaas/config-item/server/src/main/java/io/cattle/platform/configitem/context/data/metadata/common/ServiceMetaData.java
```diff
@@ -183,6 +181,16 @@ void populatePortsInfo(Service service, String serviceName) {
             this.expose.addAll((List<String>) exposeObj);
         }
     }
+    
+    void setServiceMetadata(Service service, String serviceName) {
+        if (serviceName.equals("ipsec")) {
```
alena1108, line 186: are we positive connectivity check svc will fetch it from ipsec metadata? from what I see in this PR, it's fetching it from self service (which is connectivity check svc)

kinarashah: with this change - all 3 containers of `ipsec` get the information and it works, need to understand why

kinarashah: It's fetching from self service, which gets printed as `ipsec`. 

Tested using `ipsec`, all 3 services corresponding to `ipsec`, `connectivity-check` and `router` will get the metadata info, and it works.

PR #3226, opened 2018-07-20
default service_event for ipsec to up if global flag is true

kinarashah:
https://github.com/rancher/rancher/issues/14668


alena1108:
LGTM. TF is unrelated

PR #3229, opened 2018-08-10
Fix "Form too large" exception

mrajashree:
When a SAML user belonging to multiple (~1k) groups tries to login,
a POST to v1/token containing all the group identities is done, so cattle
can create cattle identities for these groups and generates a token for the
user. This POST was causing "Form too large" exception and not letting such
users login

https://github.com/rancher/rancher/issues/8754
https://github.com/rancher/rancher/issues/14756


cjellick:
LGTM except for the checkstyle failure that is causing the build to fail

mrajashree:
@cjellick fixed checkstyle error. The current drone TF seems unrelated

cjellick:
LGTM

PR #3231, opened 2018-08-30
handle duplicate healthHostMaps

kinarashah:
delete duplicate healthcheck_instance_host_maps for same host and instance

https://github.com/rancher/rancher/issues/15284


alena1108:
LGTM

PR #3233, opened 2018-10-03
Bumped up jetty version and stop sending server version in http headers

prachidamle:
This PR bumps up the jetty version to 9.2.26.v20180806 and also stops sending Jetty server version in the HTTP headers.

https://github.com/rancher/rancher/issues/14554


alena1108:
LGTM

PR #3234, opened 2018-10-18
Support for timestamps and since as container logs args

alena1108:
https://github.com/rancher/rancher/issues/12924


StrongMonkey:
LGTM

alena1108:
TF is unrelated

PR #3235, opened 2018-10-22
Pull instance data from cache on selector based join

alena1108:
https://github.com/rancher/rancher/issues/15935


kinarashah (commented):
LGTM

PR #3250, opened 2019-01-21
Wait for consumed services to enter a finite state on service.activate

alena1108:
https://github.com/rancher/rancher/issues/15799


prachidamle:
Fix LGTM

PR #3257, opened 2019-01-23
engine install url to 18.06

alena1108:


code/packaging/app-config/src/main/resources/META-INF/cattle/api-server/defaults.properties
```diff
@@ -100,7 +100,7 @@ vm.enabled=false
 ui.pl=rancher
 ui.show.system=default_hide
 ui.show.custom.host=true
-engine.install.url=https://releases.rancher.com/install-docker/1.12.sh
+engine.install.url=https://releases.rancher.com/install-docker/18.06.1.sh
```
deniseschannon: it should be `18.06.sh` - that picks up whatever the latest 18.06.x is available.

deniseschannon:
LGTM

PR #3265, opened 2019-03-20
process health check if timestamp is not null

kinarashah:


code/iaas/healthcheck/src/main/java/io/cattle/iaas/healthcheck/process/ServiceEventCreate.java
```diff
@@ -66,7 +66,7 @@ private void processHealthcheck(ServiceEvent event) {
         } else {
             uuid = splitted[0];
         }
-        if (uuid != null) {
+        if (uuid != null && event.getExternalTimestamp() != null) {
```
alena1108: @kinarashah can we log a warning when an external timestamp is nil

alena1108:
LGTM

PR #3272, opened 2019-06-07
handle null/empty secrets

### LINKED ISSUE #17907: Can't mount empty secrets
**What kind of request is this (question/bug/enhancement/feature request):**
Bug

**Steps to reproduce (least amount of steps as possible):**
curl -u "${CATTLE_ACCESS_KEY}:${CATTLE_SECRET_KEY}" \
-X POST \
-H 'Accept: application/json' \
-H 'Content-Type: application/json' \
-d '{"description":"test", "name":"test", "value":""}' \
'http://"${CATTLE_URL}":8080/v2-beta/projects/1a5/secrets/'

**Result:**
Activating (Container should have been running but is in error state. Check logs for more information.: Failed to attach volume ########. Driver: rancher-secrets. Status code: 500. Status: 500 internal server error. Error Message: Unsuccessful request: 500 Internal Server Error)
NOTE: Any container that is using that secret will get stuck in a starting state.

**Other details that may be helpful:**
The UI will return a "Internal Server Error" if you try to create an empty secret. Also, you can do delete the secret from the UI as it will get stuck in a removing state. So you will have to go into the database and remove the entry.

**Environment information**
- Rancher version (`rancher/rancher`/`rancher/server` image tag or shown bottom left in the UI): v1.6.26
- Installation option (single install/HA): HA

**Cluster information**
- Cluster type (Hosted/Infrastructure Provider/Custom/Imported):
- Machine type (cloud/VM/metal) and specifications (CPU/memory):
- Kubernetes version (use `kubectl version`):

```
AWS EC2 using Cattle
rancher-server: c5.4xlarge; rancher-agent: m4.2xlarge
N/A
```

- Docker version (use `docker version`):

```
Client: 
Version: 18.06.1-ce 
API version: 1.38 
Go version: go1.10.6 
Git commit: e68fc7a 
Built: Tue Aug 21 17:16:31 2018 
OS/Arch: linux/amd64 
Experimental: false
```

leodotcloud:
Fixes: https://github.com/rancher/rancher/issues/17907

1) Don't allow the creation of secrets with null/empty values. (UI already has a check, API doesn't)
2) Allow deletion of already created secrets with null/empty values.

### DIFF
--- a/code/iaas/logic/src/main/java/io/cattle/platform/process/secret/SecretRemove.java
+++ b/code/iaas/logic/src/main/java/io/cattle/platform/process/secret/SecretRemove.java
@@ -12,6 +12,7 @@ import java.io.IOException;
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -25,13 +26,15 @@ public class SecretRemove extends AbstractDefaultProcessHandler {
 
     @Override
     public HandlerResult handle(ProcessState state, ProcessInstance process) {
-        Secret secret = (Secret)state.getResource();
-        try {
-            secretsService.delete(secret.getAccountId(), secret.getValue());
-        } catch (IOException e) {
-            log.error("Failed to delete secret from storage [{}]",
-                    secret.getId(), e);
-            throw new IllegalStateException(e);
+        Secret secret = (Secret) state.getResource();
+        String secretValue = secret.getValue();
+        if (StringUtils.isNotBlank(secretValue)) {
+            try {
+                secretsService.delete(secret.getAccountId(), secret.getValue());
+            } catch (IOException e) {
+                log.error("Failed to delete secret from storage [{}]", secret.getId(), e);
+                throw new IllegalStateException(e);
+            }
         }
         return null;
     }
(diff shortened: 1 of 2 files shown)


code/iaas/logic/src/main/java/io/cattle/platform/process/secret/SecretRemove.java
```diff
@@ -25,13 +26,15 @@
...
-        Secret secret = (Secret)state.getResource();
-        try {
-            secretsService.delete(secret.getAccountId(), secret.getValue());
-        } catch (IOException e) {
-            log.error("Failed to delete secret from storage [{}]",
-                    secret.getId(), e);
-            throw new IllegalStateException(e);
+        Secret secret = (Secret) state.getResource();
+        String secretValue = secret.getValue();
+        if (StringUtils.isNotBlank(secretValue)) {
```
alena1108, line 31: @leodotcloud we have to let qa know to test the scenario when empty secret was created prior to the fix, and then after performing the upgrade the secret should be removable.

alena1108:
LGTM

code/iaas/logic/src/main/java/io/cattle/platform/process/secret/SecretRemove.java
leodotcloud: yup, will work with QA for the validation.

PR #3274, opened 2019-09-06
Update default docker version to 19.03

galal-hussein:
https://github.com/rancher/rancher/issues/21808


code/packaging/app-config/src/main/resources/META-INF/cattle/api-server/defaults.properties
```diff
@@ -100,7 +100,7 @@ vm.enabled=false
 ui.pl=rancher
 ui.show.system=default_hide
 ui.show.custom.host=true
-engine.install.url=https://releases.rancher.com/install-docker/18.09.sh
+engine.install.url=https://releases.rancher.com/install-docker/19.03.sh
 account.version=2
 supported.docker.range=~v1.12.3 || ~v1.13.0 || ~v17.03.0 || ~v17.06.0 || ~v17.09.0 || ~v17.12.0 || ~v18.03.0 || ~v18.06.0 || ~v18.09.0
```
alena1108: don't we have to add 19.03 to the list of supported.docker.range?

galal-hussein: I see, will fix

alena1108 (approved):

PR #3275, opened 2019-09-17
For project member if member is deleted externally, set name as externalId if name is missing

prachidamle:
When the github team is removed on github, the projects where the team is added as a member, that environment can no longer be edited on the UI.

The reason is actually UI hitting a NPE since for github team, it tries to use the member.name property which may be null.

We are adding a fix on the backend to ensure that if the name of the member is null, we set the name to be externalID in such case where the backend entity is removed.


mrajashree (approved):
LGTM

alena1108:
https://github.com/rancher/rancher/issues/16850

